PR #10, opened 2020-01-22
Helm3 CRDs

yspotts:
This provides support for Helm 3 CRDs. In Helm 3, CRDs are defined in the new `crds` folder: https://helm.sh/docs/topics/chart_best_practices/custom_resource_definitions/. 

This solves an issue of using Kong as a dependency (subchart) of an umbrella chart and using the CRD (for example KongPlugin). This is currently not possible (or at the very least very challenging) in Helm 2 charts; see here: (helm/helm#2994).

I have followed the strategy of this project for example https://github.com/helm/charts/pull/18721 to ensure backward compatibility for helm2 by copying over the resource definitions to the templates directory as well. (See this as well: https://github.com/helm/charts/issues/19008)

Note that the new `crds` directory does not evaluate templates, and so values must be hard coded. Again, I followed the path outlined in the links above.


charts/kong/crds/custom-resource-definitions.yaml
```diff
@@ -0,0 +1,286 @@
...
+          type: string
+        type:
+          type: string
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: kongplugins.configuration.konghq.com
+  labels:
+    app.kubernetes.io/name: name
```
hbagdi: Please remove this label from all the resources.

hbagdi (changes requested):
Can we please update the docs in the README to document that installCRDs will not work with Helm 3?
This is similar to how https://github.com/helm/charts/pull/18721/files updates the documentation.

charts/kong/crds/custom-resource-definitions.yaml
yspotts: uggh, I meant to have the value be `kong`. Should I just remove the entire label or replace `name` with `kong`?

yspotts:
@hbagdi good call - will do

charts/kong/crds/custom-resource-definitions.yaml
hbagdi: Let's just remove it because it doesn't provide any value as far as I can see.

hbagdi (approved):

hbagdi:
Thank you for your contribution!

PR #20, opened 2020-01-24
[kong] Rework listener handling

rainest:

This change reworks and consolidate listener configuration:
* Replace separate per-service listener templates with a single context-based template.
* Update admin API listener handling to support the new unified template, and update its values.yaml block accordingly. The legacy single-stack admin API listener configuration is still supported, but generates a deprecation warning in NOTES.txt if found.
* Move Enterprise listener-related variables under Enterprise block in kong.env template.
* Remove checks for user-provided listener configuration around auto-generated listener configuration. This is now handled automatically for all environment variables. Remove related documentation from README, as user-provided listener overrides should be rare.
* Remove some obsolete documentation regarding the old special-case listener overrides. These are now handled by general-purpose variable overrides, and shouldn't be common enough to need documentation.

#### Checklist
- [x] Chart Version bumped
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[stable/mychartname]`)

### DIFF
--- a/charts/kong/Chart.yaml
+++ b/charts/kong/Chart.yaml
@@ -10,5 +10,5 @@ maintainers:
   email: harry@konghq.com
 name: kong
 sources:
-version: 1.1.0
+version: 1.2.0
 appVersion: 1.4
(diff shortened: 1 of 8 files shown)


rainest:
Closing as-is. CI initially failed because I missed the legacy behavior in the controller template. However, thinking about this more, I want to also add CI to ensure both variants work, and see if I can make the generic listener template more configurable (namely able to support configurations other than just TLS and not-TLS HTTP, since we offer a number of other options).

PR #26, opened 2020-01-30
Create CI configuration for "next" branch

rainest:
Creates a workflow for checking PRs to the `next` branch that skips the version increment check in `ct lint`. Remove the existing configuration we added to `master`, which should be kept as-is.

Tested locally to confirm desired effect: https://gist.github.com/rainest/95149d16e4f683d282a4777e68278ec7

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -18,7 +18,7 @@ jobs:
         env:
           CT_CHECK_VERSION_INCREMENT: "false"
         with:
-          command: lint
+          command: lint --config=ct-master.yaml
 
       - name: Create kind cluster
         uses: helm/kind-action@v1.0.0-alpha.3
@@ -29,7 +29,7 @@ jobs:
       - name: Run chart-testing (install)
         uses: helm/chart-testing-action@v1.0.0-alpha.3
         with:
-          command: install
+          command: install --config=ct-master.yaml
   release:
     needs: lint-test
     runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


rainest:
In conjunction with this, we should use https://help.github.com/en/github/administering-a-repository/configuring-protected-branches to require PRs for merger to next (and master, if we do not already).

hbagdi:
I've setup `master` and `next` to be protected branches that need at least one Github review approval. This includes admins who are authors. I've not enforced CI since our CI has issues around version bumps.
We need to setup https://github.blog/2017-07-06-introducing-code-owners/ and then I'll update the configuration to ensure that only approval from maintainers count.

rainest:
Updated to handle this the other way around: the default ct.yaml now disables version checks, and master's CI jobs explicitly specify a different config, ct-master.yaml, which enables version checks.

Did not use flags for master's jobs since (a) it appears that [install also does some increment checking](https://github.com/helm/chart-testing/blob/master/doc/ct_install.md) but doesn't have a flag to control that alone and (b) there are several master jobs, and it's simpler to just point them at a single config file rather than add flags to each individually if need be in the future.

.github/workflows/release.yaml
```diff
@@ -18,7 +18,7 @@ jobs:
         env:
           CT_CHECK_VERSION_INCREMENT: "false"
```
hbagdi: Can we remove this?

hbagdi (approved):
The duplication of workflow is very concerning and I'm sure there is a way to simplify this in Actions, we just don't know about it yet.

PR #32, opened 2020-01-31
fix ci

hbagdi:


hbagdi:
cc @yasn77

hbagdi:
This solves only one part of the problem.
The other side is the release job runs on push to master and it fails currently because it expects the version to be bump up.
This could be solved once we adopt the `next` branch convention.

hbagdi:
@rainest included these changes in #26.

PR #34, opened 2020-02-04
Helm3 warning

yspotts:
add a section in the readme about helm2 vs helm3

I also set the default value of false. Not sure if we want to keep this... Nice for helm 3 but could mess up those using helm 2


charts/kong/values.yaml
```diff
@@ -205,7 +205,7 @@ ingressController:
     # If not set and create is true, a name is generated using the fullname template
     name:
 
-  installCRDs: true
+  installCRDs: false
```
hbagdi: Please do not change the default. A significant, maybe even the majority, of users are using Helm v2.

hbagdi: Please upgrade the above description accordingly as well.

yspotts: will change back. 

sorry  - when you say "the above description" do you mean in the readme regarding the default value?

hbagdi: Yes and the text in Helm 2 vs 3, where applicable.

yspotts: got it! I will remove this commit from the PR

hbagdi (approved):
Looks good. Please resolve the conflict.

lindhe:
Which conflict?

yspotts:
Conflict resolved

hbagdi:
@yspotts Thank you for your contribution. Please fill in the contributor form to claim your Kong swag: https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#contributor-t-shirt.

Issue #30, opened 2020-01-31
Helm upgrade causes service account to be recreated (recreating service token)


yasn77:
Helm upgrade recreates service account, which in turn recreates the secret token
for the service account. This breaks current deployments. Below are symptoms:

Before Upgrade:
--------------

```
kubectl -n kong get serviceaccount -l app.kubernetes.io/name=kong-app -o jsonpath='{.items[].secrets}'
[map[name:kong-tst-kong-app-token-k7qv2]]
```

```
kubectl -n kong get pod kong-tst-kong-app-cc9787b65-fwzrd -o json | jq '.spec.containers[] | select(.name == "ingress-controller").volumeMounts'
[
  {
    "mountPath": "/var/run/secrets/kubernetes.io/serviceaccount",
    "name": "kong-tst-kong-app-token-k7qv2",
    "readOnly": true
  }
]
```

(Note: Token is the same)

After Upgrade:
-------------

```
kubectl -n kong get serviceaccount -l app.kubernetes.io/name=kong-app -o jsonpath='{.items[].secrets}'
[map[name:kong-tst-kong-app-token-rrzfh]]
```

```
kubectl -n kong get pod kong-tst-kong-app-cc9787b65-fwzrd -o json | jq '.spec.containers[] | select(.name == "ingress-controller").volumeMounts'
[
  {
    "mountPath": "/var/run/secrets/kubernetes.io/serviceaccount",
    "name": "kong-tst-kong-app-token-k7qv2",
    "readOnly": true
  }
]
```

(Note: Tokens are different)

From the logs (after upgrade):

Tiller:
------
```
[tiller] 2020/01/31 17:01:46 deleting pre-upgrade hook kong-tst-kong-app for release kong-tst due to "before-hook-creation" policy
[kube] 2020/01/31 17:01:46 Starting delete for "kong-tst-kong-app" ServiceAccount
[kube] 2020/01/31 17:01:46 Waiting for 60 seconds for delete to be completed
[kube] 2020/01/31 17:01:48 building resources from manifest
[kube] 2020/01/31 17:01:48 creating 1 resource(s)
[kube] 2020/01/31 17:01:48 Watching for changes to ServiceAccount kong-tst-kong-app with timeout of 5m0s
[kube] 2020/01/31 17:01:48 Add/Modify event for kong-tst-kong-app: ADDED
```

From IngresController:
----------------------
```
[...]

PR #36, opened 2020-02-05
[kong] add hook to init migrations job

rainest:
**What this PR does / why we need it:**

Control the lifecycle of init-migrations jobs with a hook. The job is created after resources are created (namely, after Postgres is created if its subchart is enabled), runs, and is deleted if it succeeds. Install flow should proceed normally after. The job is no longer created during upgrades.

**Which issue this PR fixes**

I thought there was an issue filed for this somewhere, but could not find it. It may have been lost in the depths of the helm/charts repo. See previous discussion in https://github.com/Kong/charts/pull/29#pullrequestreview-351858753

Not deleting the job automatically and creating it for upgrades also allows it to conflict with previous versions of itself, resulting in a very unintuitive error and requiring that users know they need to delete it manually.

### DIFF
--- a/charts/kong/FAQs.md
+++ b/charts/kong/FAQs.md
@@ -30,3 +30,42 @@ If you do, then please delete the release, delete the volume, and then
 do a fresh installation. PersistentVolumes can remain in the cluster even if
 you delete the namespace itself (the namespace in which they were present).
 
+#### Upgrading a release fails due to missing ServiceAccount
+
+When upgrading a release, some configuration changes result in this error:
+
+```
+Error creating: pods "releasename-kong-pre-upgrade-migrations-" is forbidden: error looking up service account releasename-kong: serviceaccount "releasename-kong" not found
+```
+
+Enabling the ingress controller or PodSecurityPolicy requires that the Kong
+chart also create a ServiceAccount. When upgrading from a configuration that
+previously had neither of these features enabled, the pre-upgrade-migrations
+Job attempts to use this ServiceAccount before it is created. It is [not
+possible to easily handle this case automatically](https://github.com/Kong/charts/pull/31).
+
+Users encountering this issue should temporarily modify their
+[pre-upgrade-migrations template](https://github.com/Kong/charts/blob/master/charts/kong/templates/migrations-pre-upgrade.yaml),
+adding the following at the bottom:
+
+```
+{{ if or .Values.podSecurityPolicy.enabled (and .Values.ingressController.enabled .Values.ingressController.serviceAccount.create) -}}
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ template "kong.serviceAccountName" . }}
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/hook": pre-upgrade
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+  labels:
+    {{- include "kong.metaLabels" . | nindent 4 }}
+{{- end -}}
+```
+
+Upgrading with this in place will create a temporary service account before
+creating the actual service account. After this initial upgrade, users must
+revert to the original pre-upgrade migrations template, as leaving the
+temporary ServiceAccount template in place will [cause permissions issues on
+subsequent upgrades](https://github.com/Kong/charts/issues/30).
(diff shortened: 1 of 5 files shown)


rainest:
Helm 3 creates this job fine, but either Helm 2 does not or something weird is going on with CI. `post-install` should be available there also: https://v2.helm.sh/docs/charts_hooks/#hooks

hbagdi:
`post-install` will not work because Kong deployment doesn't reach healthy status.
We have been bit by it before.

I'd recommend just deleting the migration job on success. We should do it for pre and post upgrade migration jobs as well.

rainest:
It shouldn't unless instructed to. Per that doc:
> 8. Tiller loads the resulting resources into Kubernetes. Note that if the --wait flag is set, Tiller will wait until all resources are in a ready state and will not run the post-install hook until they are ready.

And, in reality, it doesn't: installing with either Helm3 or Helm2 using this change works fine. However, CI does not live in reality, it lives in the world of chart-testing 🙂

chart-testing hard-codes `--wait` into its Helm invocations, e.g. https://github.com/helm/chart-testing/blame/v2.4.0/pkg/tool/helm.go#L62-L63

Not clear why that's the case. It exists in the oldest version in the chart-testing repo, but doesn't in the version that was present in update.sh when it was moved out of the charts repo: https://github.com/helm/charts/pull/5231/files

May be related to timeout handling, but not sure. Sadly, I don't think there's any non-code change way to override it.

hbagdi:
Any reason we really want the `post-install` hook?
(It is so counter-intuitive that it is post-install, it should be pre-install. I know Helm details don't allow for that and that's not the point here either.)

rainest:
`init-migrations` depends on the release spawning Postgres when the subchart is enabled, and Helm waits for `pre-install` resources to become ready (successfully completed for Job resources) before spawning anything else. This will cause `init-migrations` to either get stuck on waiting for Postgres or prevent it from spawning at all (it also depends on the ServiceAccount if using features that need it).

Running in `pre-install` would require that we do the same spawn+delete for the ServiceAccount and Postgres that led to #30. Although that shouldn't be an issue with install, it's still kinda hacky and I'm not actually sure we can do it without forking the Postgres subchart.

`post-install` sounds counter-intuitive, but it does make sense here, in that it's effectively "install Kong and its various dependencies, then run `kong migrations bootstrap` before running `kong start`" given our various initContainers.

rainest:
After some research, I am less keen on K8S' native https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/#ttl-mechanism-for-finished-jobs as an alternative. It looks like GKE supports it out of the box without further configuration (or I enabled alpha features earlier on my test cluster and forgot), but EKS and AKS likely do not support it at all:
* At least as of July 2019, EKS [apparently rejects this configuration outright](https://github.com/jaegertracing/jaeger-operator/issues/494). A [request to add it](https://github.com/aws/containers-roadmap/issues/255) remains open, but discussion there indicates it will not be available until it enters beta.
* AKS apparently [ignores it silently and does allow users to enable it via a feature flag](https://github.com/Azure/AKS/issues/833). However, documentation suggests that it's not granular and enables all experimental functionality.

hbagdi:
Two options AFAICS:
1. Contribute a PR to chart-testing.
2. Try using only `helm.sh/hook-delete-policy: "hook-succeeded"` and remove the other annotation and see the behavior of Helm.

rainest:
Not viable as proposed, and would presumably need new Helm functionality (a new hook) to support (it's not possible to fix it in chart-testing only). Closing as such, but keeping the branch around for now.

hbagdi:
Travis,

I thought this was almost there and only CI issues were the blocker. Am I misunderstanding something?

rainest:
The CI issues are quite hard to work around. More detail in https://github.com/helm/chart-testing/issues/202 but in short:
* chart-testing relies on `--wait` to know whether the chart installed successfully. Confirming the same without `--wait` would require some other command that provided the same information.
* Helm has more knowledge of individual release components' status inside `install` execution than outside. It previously had functionality to get that level of detail outside, but that functionality was removed because it was very prone to breaking when other components were updated.

Absent some command-level alternative to `--wait` I can't think of any way to handle this in chart-testing, and the next best option would be a new hook phase we could use in the template. Even assuming that's accepted, it'd break Helm 2 compatibility.

PR #45, opened 2020-02-12
Fixing small typo in hpa

chethanuk:
@yasn77


yasn77:
Ah... my bad. Good catch @ChethanUK. That's difficult to pick in tests since all the k8 clusters I tested with support `v2beta2`, I so i never hit that code branch :/

chethanuk:
> pick in tests since all the k8 clusters

Ya.

hbagdi (approved):

PR #47, opened 2020-02-13
kong: support for grpc proxying

yndai:
This PR adds options for the proxy to expose ports for gRPC traffic. I have done some basic manual testing on my end. Let me know if there are issues.

### DIFF
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -57,18 +57,17 @@ Create the KONG_PROXY_LISTEN value string
 */}}
 {{- define "kong.kongProxyListenValue" -}}
 
-{{- if and .Values.proxy.http.enabled .Values.proxy.tls.enabled -}}
-   0.0.0.0:{{ .Values.proxy.http.containerPort }},0.0.0.0:{{ .Values.proxy.tls.containerPort }} ssl
-{{- else -}}
 {{- if .Values.proxy.http.enabled -}}
-   0.0.0.0:{{ .Values.proxy.http.containerPort }}
-{{- end -}}
-{{- if .Values.proxy.tls.enabled -}}
-   0.0.0.0:{{ .Values.proxy.tls.containerPort }} ssl
-{{- end -}}
+    0.0.0.0:{{ .Values.proxy.http.containerPort }}{{- if .Values.proxy.tls.enabled -}},0.0.0.0:{{ .Values.proxy.tls.containerPort }} ssl{{- end -}}{{- if .Values.proxy.grpc.enabled -}},0.0.0.0:{{ .Values.proxy.grpc.containerPort }} http2{{- end -}}{{- if .Values.proxy.grpcs.enabled -}},0.0.0.0:{{ .Values.proxy.grpcs.containerPort }} http2 ssl{{- end -}}
+{{- else if .Values.proxy.tls.enabled -}}
+    0.0.0.0:{{ .Values.proxy.tls.containerPort }} ssl{{- if .Values.proxy.grpc.enabled -}},0.0.0.0:{{ .Values.proxy.grpc.containerPort }} http2{{- end -}}{{- if .Values.proxy.grpcs.enabled -}},0.0.0.0:{{ .Values.proxy.grpcs.containerPort }} http2 ssl{{- end -}}
+{{- else if .Values.proxy.grpc.enabled -}}
+    0.0.0.0:{{ .Values.proxy.grpc.containerPort }} http2{{- if .Values.proxy.grpcs.enabled -}},0.0.0.0:{{ .Values.proxy.grpcs.containerPort }} http2 ssl{{- end -}}
+{{- else if .Values.proxy.grpcs.enabled -}}
+    0.0.0.0:{{ .Values.proxy.grpcs.containerPort }} http2 ssl
 {{- end -}}
 
-{{- end }}
+{{- end -}}
 
 {{/*
 Create the KONG_ADMIN_GUI_LISTEN value string
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -57,14 +57,15 @@ Create the KONG_PROXY_LISTEN value string
 */}}
 {{- define "kong.kongProxyListenValue" -}}
 
+{{- if and .Values.proxy.http.enabled .Values.proxy.tls.enabled -}}
+   0.0.0.0:{{ .Values.proxy.http.containerPort }} reuseport,0.0.0.0:{{ .Values.proxy.tls.containerPort }} ssl http2 reuseport
+{{- else -}}
 {{- if .Values.proxy.http.enabled -}}
-    0.0.0.0:{{ .Values.proxy.http.containerPort }}{{- if .Values.proxy.tls.enabled -}},0.0.0.0:{{ .Values.proxy.tls.containerPort }} ssl{{- end -}}{{- if .Values.proxy.grpc.enabled -}},0.0.0.0:{{ .Values.proxy.grpc.containerPort }} http2{{- end -}}{{- if .Values.proxy.grpcs.enabled -}},0.0.0.0:{{ .Values.proxy.grpcs.containerPort }} http2 ssl{{- end -}}
-{{- else if .Values.proxy.tls.enabled -}}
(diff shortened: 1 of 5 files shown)


hbagdi:
Before we get into the code, can you elaborate on how you are using gRPC with Kong in your use-case? Specifically,
- Are you using grpc or grpcs?
- Who are the clients that are calling the grpc services?
- How are you hosting these and how are you configuring Kong to route traffic to different grpc services?

yndai:
Hi @hbagdi ,
My organization is in the middle of transitioning to services in K8s and we would like to use Kong as the ingress controller. We already have it deployed and proxying some REST services, but we will soon begin deploying gRPC services as well.

- **Are you using grpc or grpcs?**
Initially it would likely just be internal grpc traffic, but public grpcs traffic may be down the line
- **Who are the clients that are calling the grpc services?**
First it will be serving internal service traffic from outside the K8s cluster
- **How are you hosting these and how are you configuring Kong to route traffic to different grpc services?**
We host our K8s cluster in cloud instances and have a loadbalancer in front of Kong. For gRPC services we will probably use path based routing in our Ingress resources.

hbagdi:
Thanks for your response, @yndai.

Thoughts here:

HTTP(S) has standardized ports so we don't run into problems. TCP/TLS proxy is another feature that is coming in 0.8 version and we will need to add support for it. The way I see `grpc`, I think that overtime, more and more people will use `grpcs`, and then we port 443 (or 8443 ssl in Kong) can be re-used for grpc traffic.

I think it is worth while to make it easy for grpc support to be enabled/disabled easily, but the way it works current is in this PR is limited. We should make it possible to expose any number of ports of Kong (in addition to the current standard ports), and then associate different protocols with those ports.

We should also try to re-use port 443/8443 with http2 for grpcs rather than introducing a new port.

charts/kong/values.yaml
```diff
@@ -103,6 +103,22 @@ proxy:
...
+    # Set a nodePort which is available if service type is NodePort
+    # nodePort: 32081
+
+  # gRPC/HTTP2 TLS traffic (disabled by default)
+  grpcs:
+    enabled: false
+    servicePort: 9443
+    containerPort: 9443
+    # Set a nodePort which is available if service type is NodePort
+    # nodePort: 32444
```
hbagdi: I think rather than introducing a section for `grpc` and `grpcs` here, we should go for something along the lines:

```yaml
additionalPorts:
- servicePort: 9443
  containerPort: 9443
  protocols: # bad name, these are suffixes (https://github.com/Kong/kong/blob/master/kong.conf.default#L195)
  - ssl
  - http2 
  - backlog=N
```

This works for TCP/TLS proxying cases as well.

@rainest has thought about this problem as well and I'm sure he will have more thoughts here.

yndai: I agree with your thoughts and I like this solution! More elegant than what I have. Will make this change.
Regarding the naming of `protocols` I think `options` would be an appropriate name.

rainest: I haven't thought about the values.yaml format much yet (I'd done work on the template end and then realized that we didn't have anything for handling the additional options). I wouldn't want to distinguish between "standard" and "additional" listens, however, since there's not really any distinction between them in the resulting configuration, and it's easier to template if they're both instances of the same type of config.

That, however, would be break with our existing values.yaml format, which does designate specific listens. IMO we should tack parameters (not really a clearer name than the other options, but it's [what NGINX calls them](http://nginx.org/en/docs/http/ngx_http_core_module.html#listen)) onto those existing blocks, e.g.

```
proxy:
  tls:
    enabled: true
    servicePort: 443
    containerPort: 8443
    parameters:
    - http2
    - ssl
    - reuseport
    - backlog=16384
```
We can then defer a more flexible format to some future major release where we do introduce significant breaking changes to the values.yaml format.

That does not allow adding more listens, but I think that's fine for now: for most of our listeners, there's not much reason to add additional ports, since Kong doesn't distinguish between them (you cannot, for example, create a route that is available on one proxy listener but not another). The exception to this is stream listens, which absolutely do function that way, and for which the chart does not have any support currently.

A simpler temporary option (since it may take a while to complete the non-breaking listener rework) would be to just add the [2.0 default parameters](https://github.com/Kong/kong/blob/2.0.0/kong.conf.default#L186) to the [existing proxy_listen helper](https://github.com/helm/charts/blob/master/stable/kong/templates/_helpers.tpl#L60-L69). Although they weren't in the default listener until 2.0, they should be compatible with anything we'd realistically find in the wild.

yndai:
@rainest Hi, I have made changes based on your temporary option. I would have preferred the more flexible approach to proxy plaintext gRPC, but I understand that it could be more awkward to transition from down the line. 

Regarding the `backlog=N` option, I left it out because enabling it in the current chart caused the following error: 
```
Error: proxy_listen must be of form: [off] | <ip>:<port> [ssl] [http2] [proxy_protocol] [transparent] [deferred] [bind] [reuseport], [... next entry ...]
```
I am guessing because the packaged nginx version in `kong:1.4.1` is older? `nginx version: openresty/1.15.8.2`

rainest (approved):

charts/kong/templates/_helpers.tpl
```diff
@@ -58,17 +58,17 @@ Create the KONG_PROXY_LISTEN value string
...
-   0.0.0.0:{{ .Values.proxy.http.containerPort }},0.0.0.0:{{ .Values.proxy.tls.containerPort }} ssl
+   0.0.0.0:{{ .Values.proxy.http.containerPort }} reuseport,0.0.0.0:{{ .Values.proxy.tls.containerPort }} ssl http2 reuseport
 {{- else -}}
 {{- if .Values.proxy.http.enabled -}}
-   0.0.0.0:{{ .Values.proxy.http.containerPort }}
+   0.0.0.0:{{ .Values.proxy.http.containerPort }} reuseport
 {{- end -}}
 {{- if .Values.proxy.tls.enabled -}}
-   0.0.0.0:{{ .Values.proxy.tls.containerPort }} ssl
+   0.0.0.0:{{ .Values.proxy.tls.containerPort }} ssl http2 reuseport
```
hbagdi, line 67: The repetition here needs to be fixed but is not in the scope of the PR.

hbagdi (approved):
I don't think these will cause any problems but in case they do, we will have to revert the `reuseport` option. In practice, `reuseport` actually improves Kong's performance quite a bit, but there might be corner cases that we are not currently aware of.
Good improvement. THanks @yndai

rainest:
> I would have preferred the more flexible approach to proxy plaintext gRPC, but I understand that it could be more awkward to transition from down the line.

@yndai Can you elaborate on where/how you need this in practice? Plaintext gRPC does still require HTTP/2 transport as far as I know, and plaintext HTTP/2 is quite uncommon. NGINX's support for it is limited as there isn't [support for HTTP/1.1 and HTTP/2 on the same plaintext listen](https://trac.nginx.org/nginx/ticket/816).

> Regarding the `backlog=N` option, I left it out because enabling it in the current chart caused the following error:

Bummer, looks like we have [a subset of allowed parameters](https://github.com/Kong/kong/blob/1.3.0/kong/conf_loader.lua#L957-L960) and validate against that. Shouldn't be any NGINX-level restriction: they don't mention any in their doc and that parameter is for a syscall argument rather than something in NGINX's own internals.

adambom:
@rainest I'm curious if there's some background context or knowledge I'm missing around this statement:

> Plaintext gRPC does still require HTTP/2 transport as far as I know, and plaintext HTTP/2 is quite uncommon

Couldn't you still take advantage of the durable connections and bi-directional streaming capabilities of HTTP/2 without needing SSL?

hbagdi:
> Bummer, looks like we have a subset of allowed parameters and validate against that. Shouldn't be any NGINX-level restriction: they don't mention any in their doc and that parameter is for a syscall argument rather than something in NGINX's own internals.

Yeah, that was a limitation Datong went ahead with because it required a lot more work in conf_loader. PR welcome to fix it. Most people can leave backlog alone.

yndai:
> @yndai Can you elaborate on where/how you need this in practice? Plaintext gRPC does still require HTTP/2 transport as far as I know, and plaintext HTTP/2 is quite uncommon. NGINX's support for it is limited as there isn't support for HTTP/1.1 and HTTP/2 on the same plaintext listen.

@rainest We would like to use Kong for ingress of both public and some internal traffic (outside of k8s cluster, but still internal to our systems) and plaintext gRPC would be a bit simpler for the internal traffic case.

hbagdi:
Let's open a Github issue to track this conversation and feature request.

hbagdi:
@yndai  Thank you for your contribution. Please fill in the contributor form to claim your Kong swag: https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#contributor-t-shirt.

PR #48, opened 2020-02-14
fix(nodePort) chart respects LoadBalancer NodePort

artburkart:
# What

This PR introduces ability to specify a custom `nodePort` in values
passed to `helm`. Before, the chart would only respect the `nodePort`
value if `proxy.type` was set to `NodePort`. Now it also respects
`nodePort` for default `proxy.type` of `LoadBalancer`.

# Testing

Deployed and verified working with https://bit.ly/echo-service


hbagdi (approved):
1. It might be better to just remove the checks and populate `nodePort` if it is set by the user. If there is one more service type that needs it, then we will probably do that.
2. There is value in fixing this for other services too (admin, manager, portal). I'm going for iterative improvements here and going ahead and merging this in for now.

PR #49, opened 2020-02-14
crd install hook

yspotts:
Make the crds `crd-install` hook. 
This will hopefully go the next step in providing helm 2 and helm 3 support seamlessly. The crd-install hook is ignored (AFAIK) in helm 3 so the crds wont be installed a second time in helm 3. 

(If this works as designed, not sure we need the same warning as presented in: https://github.com/Kong/charts/pull/34)

### DIFF
--- a/charts/kong/crds/custom-resource-definitions.yaml
+++ b/charts/kong/crds/custom-resource-definitions.yaml
@@ -2,6 +2,8 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   name: kongconsumers.configuration.konghq.com
+  annotations:
+    "helm.sh/hook": crd-install
 spec:
   group: configuration.konghq.com
   version: v1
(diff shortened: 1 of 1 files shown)


hbagdi:
The root cause of CI failures is https://github.com/helm/helm/issues/5806.

Any suggestions on how to move forward?

yspotts:
did this issue just present itself with this PR? seems like it would be an issue no matter what.

hbagdi:
I'm not sure why this popped up with this change. It was never an issue before.

rainest (commented):
It looks like https://github.com/helm/helm/issues/5806 only crops up if the CRDs are installed via hooks, which explains why they only fail with this PR. With the hook, Helm creates the CRD and then checks that it's established before proceeding with the next step of the install, and in doing so tries to convert it into the appropriate schema, resulting in the `unable to convert to CRD type: unable to convert unstructured object to apiextensions.k8s.io/v1beta1, Kind=CustomResourceDefinition: cannot convert int64 to float64` error.

Without the hooks, Helm just yeets the CRDs into Kubernetes and doesn't care what happens with them after, so the problem conversion never occurs. I think we're just blocked on this until/if there's a 2.x fix for that bug.

rainest:
The next release (2.0.0) will be Helm 3-only, and use its CRDs directory. We've taken a different approach to handling the old templated CRDs, since they can't easily be separated from a release: https://github.com/Kong/charts/pull/305

PR #54, opened 2020-02-21
Remove resuseport from proxy listen

rainest:

I should have been more diligent when checking legacy support in #47. Turns out there's a lot of stuff still common in the wild (everything <1.3.0) that still doesn't support `resuseport` in listens.

The good news is that `next` is working as intended to catch this sort of thing before release!

#### Checklist


hbagdi (approved):

hbagdi:
Another reason to make this configurable rather than hard-coding.

PR #38, opened 2020-02-10
Fix NOTES indentation

lindhe:
Previously, the installation notes (NOTES.txt) had inconsistent indentation making it hard to read sometimes. This commit fixes that.

I also chose to remove the proceeding whitespaces so that the commands are easier to select with the mouse in order to copy-paste them.

**Before:**
Before
**After**
After

PR #55, opened 2020-02-22
Remove dedicated Portal auth configuration from values.yaml

### LINKED ISSUE #52: Cannot configure portal_auth_conf when using openid-connect for portal_auth
Looking at the values.yaml for enterprise portal, there isn't a way to configure `portal_auth_conf` when `portal _auth: openid-connect`.   The only option, which is also required, is to provide a `session_conf_secret`.  According to the docs here:  https://docs.konghq.com/enterprise/1.3-x/developer-portal/configuration/authentication/sessions/, `session_conf` cannot be applied to openid-connect.

rainest:
Remove the dedicated Portal auth/session configuration. These are not necessary for most versions, as Kong Enterprise will start fine without them and allows users to set them via a workspace's Portal configuration in Manager.

This change continues support for the existing configuration to maintain compatibility with existing user values.yamls. It places a warning in NOTES.txt to inform users that the configuration is deprecated and should be moved under `env` if they wish to keep it in place.

Fix #52

Unlike other versions, 0.35 does *not* support configuring sessions via Manager: the kong.conf `portal_session_conf` is the only way to set Portal session configuration, and is used by all workspaces. Setting this in `env` or continuing to use the deprecated configuration works for 0.35.

#### Checklist

### DIFF
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -471,9 +471,9 @@ secret/kong-session-config created
 The exact plugin settings may vary in your environment. The `secret` should
 always be changed for both configurations.
 
-After creating your secret, set its name in values.yaml, in the
-`.enterprise.rbac.session_conf_secret` and optionally in
-`env.portal_session_conf` using a secretKeyRef.
+After creating your secret, set its name in values.yaml in
+`.enterprise.rbac.session_conf_secret`. If you create a Portal configuration,
+add it at `env.portal_session_conf` using a secretKeyRef.
 
 ### Email/SMTP
 
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -452,9 +452,13 @@ accessible outside the Pod.
 
 Login sessions for Kong Manager and the Developer Portal make use of
 [the Kong Sessions plugin](https://docs.konghq.com/enterprise/latest/kong-manager/authentication/sessions).
-Their configuration must be stored in Secrets, as it contains an HMAC key.
-If using either RBAC or the Portal, create a Secret with `admin_gui_session_conf`
-and `portal_session_conf` keys.
+When configured via values.yaml, their configuration must be stored in Secrets,
+as it contains an HMAC key.
+
+Kong Manager's session configuration must be configured via values.yaml,
+whereas this is optional for the Developer Portal on versions 0.36+. Providing
+Portal session configuration in values.yaml provides the default session
+configuration, which can be overriden on a per-workspace basis.
 
 ```
 $ cat admin_gui_session_conf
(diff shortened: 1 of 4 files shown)


charts/kong/README.md
```diff
@@ -466,8 +470,8 @@ The exact plugin settings may vary in your environment. The `secret` should
 always be changed for both configurations.
 
 After creating your secret, set its name in values.yaml, in the
-`.enterprise.rbac.session_conf_secret` and
-`.enterprise.portal.session_conf_secret` keys.
+`.enterprise.rbac.session_conf_secret` and optionally in
+`env.portal_session_conf` using a secretKeyRef.
```
hbagdi: To promote good security practices, let's encourage secretKeyRef rather than mentioning it as an option.

charts/kong/templates/NOTES.txt
```diff
@@ -1,16 +1,23 @@
...
+PORT=$(kubectl get svc --namespace {{ .Release.Namespace }} {{ template "kong.fullname" . }}-proxy -o jsonpath='{.spec.ports[0].nodePort}')
 {{- end -}}
 export PROXY_IP=${HOST}:${PORT}
 curl $PROXY_IP
 
 Once installed, please follow along the getting started guide to start using Kong:
 https://bit.ly/k4k8s-get-started
 
+{{ if and (.Values.enterprise.portal.enabled) (or (.Values.enterprise.portal.portal_auth) (.Values.enterprise.portal.session_conf_secret)) -}} {{/* Legacy Portal auth handling */}}
+/!\ WARNING: You are currently using legacy Portal authentication configuration in values.yaml (https://github.com/Kong/charts/blob/kong-1.2.0/charts/kong/values.yaml#L384-L392). Support for this will be removed in a future release.
```
hbagdi, line 17: 1. I recommend to respect 80 char column limit in this file as well. It renders better for end users as well.
2. I'm not sure if the configuration value is declared as deprecated. If it is then cool, but please double check to make sure we are confusing the end user.

hbagdi (approved):
Looks good.

Please merge once the comments are resolved, thanks!

Please also see: https://github.com/Kong/charts/pull/38#issuecomment-590677258

charts/kong/templates/NOTES.txt
rainest: To clarify, I'm deprecating their current dedicated values.yaml field. The settings still exist, but should now just be placed under `env` if needed. The updated NOTES message just instructs all users to do this to avoid reverting to the standard default values accidentally. New users generally shouldn't create them at all, and should instead rely on workspace-level configuration.

PR #65, opened 2020-02-27
[kong] update registry secret documentation

rainest:
* Update the registry secret documentation to add secrets for both kong-enterprise-k8s and kong-enterprise-edition-docker registries. We do include both in values.yaml comments, but it's easy to miss that they're in separate registries (and need separate secrets) rather than different images in the same registry.
* Update examples to clarify the credential type and format.
* Indicate more clearly that users should use their Bintray API key, which may be different from their Bintray password.

#### Checklist

### DIFF
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -398,22 +398,30 @@ Kong is going to be deployed.
 #### Kong Enterprise Docker registry access
 
 Next, we need to setup Docker credentials in order to allow Kubernetes
-nodes to pull down Kong Enterprise Docker image, which is hosted as a private
-repository.
+nodes to pull down Kong Enterprise Docker images, which are hosted in a private
+registry.
 
-As part of your sign up for Kong Enterprise, you should have received
-credentials for these as well.
+You should received credentials to log into https://bintray.com/kong after
+purchasing Kong Enterprise. After logging in, you can retrieve your API key
+from \<your username\> \> Edit Profile \> API Key. Use this to create registry
+secrets:
 
 ```bash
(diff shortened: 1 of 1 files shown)


charts/kong/README.md
```diff
@@ -398,22 +398,30 @@ Kong is going to be deployed.
...
-secret/kong-enterprise-docker created
+    --docker-username=<your-username@kong> \
+    --docker-password=<your-bintray-api-key>
+secret/kong-enterprise-k8s-docker created
+
+$ kubectl create secret docker-registry kong-enterprise-edition-docker \
+    --docker-server=kong-docker-kong-enterprise-edition-docker.bintray.io \
+    --docker-username=<your-username@kong> \
+    --docker-password=<your-bintray-api-key>
+secret/kong-enterprise-edition-docker created
```
hbagdi, line 420: This is a small and a positive change but it breaks instructions that we have in other places. We assume `kong-enterprise-docker` as the secret name in other places.
We will have to co-ordinate this change across multiple places.

rainest:
Anywhere other than the docs site and controller docs/examples? I've created PRs for those.

We would want to release the docs/controller changes simultaneously, but can probably release the chart change independently, as the chart doesn't have the same tight coupling between the documentation and actual configuration (there is no default pull secret, and you must configure whatever name you use in values.yaml).

hbagdi:
https://gist.github.com/hbagdi/3ca587f2306282ec57eb8d480db5768c

This is referenced as well. I've updated it as soon as we merge all the PRs.

hbagdi (approved):

PR #71, opened 2020-03-10
feat(kong): adding pod annotations for Kuma and Istio support

subnetmarco:
Adding annotations to support upstream Kuma and Istio meshes in Kong.

### DIFF
--- a/charts/kong/templates/deployment.yaml
+++ b/charts/kong/templates/deployment.yaml
@@ -6,6 +6,11 @@ metadata:
     {{- include "kong.metaLabels" . | nindent 4 }}
     app.kubernetes.io/component: app
 spec:
+  template:
+    metadata:
+      annotations:
+        kuma.io/gateway: enabled
+        traffic.sidecar.istio.io/includeInboundPorts: ""
   replicas: {{ .Values.replicaCount }}
   selector:
     matchLabels:
@@ -29,8 +34,6 @@ spec:
         {{- if .Values.podAnnotations }}
 {{ toYaml .Values.podAnnotations | indent 8 }}
         {{- end }}
-        kuma.io/gateway: enabled
-        traffic.sidecar.istio.io/includeInboundPorts: ""
       labels:
         {{- include "kong.metaLabels" . | nindent 8 }}
         app.kubernetes.io/component: app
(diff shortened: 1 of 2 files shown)


hbagdi:
@subnetmarco I might have pointed you in the wrong direction.
Do you want to add the annotations at the Deployment level or pod level?

Deployment: https://github.com/Kong/charts/blob/master/charts/kong/templates/deployment.yaml#L4
Pod: https://github.com/Kong/charts/blob/master/charts/kong/templates/deployment.yaml#L20

This PR is adding it at the Pod-level.

subnetmarco:
@hbagdi good catch - updated.

charts/kong/templates/deployment.yaml
```diff
@@ -29,6 +29,8 @@ spec:
         {{- if .Values.podAnnotations }}
 {{ toYaml .Values.podAnnotations | indent 8 }}
         {{- end }}
+        kuma.io/gateway: enabled
+        traffic.sidecar.istio.io/includeInboundPorts: ""
```
hbagdi: This is same as the previous commit. The annotations are still at pod level instead of Deployment level. Deployment annotations will go under Line 7 (in a new annotations block).

subnetmarco: I would like to replicate the same behavior as https://github.com/Kong/kubernetes-ingress-controller/pull/564

subnetmarco:
@hbagdi updated again, this time it should be in the right place. I noticed that the `Deployment` here doesn't specify any `label` like we do in our other installation method by having:

```yaml
labels:
  app: ingress-kong
```

charts/kong/templates/deployment.yaml
```diff
@@ -6,6 +6,11 @@ metadata:
     {{- include "kong.metaLabels" . | nindent 4 }}
     app.kubernetes.io/component: app
 spec:
+  template:
+    metadata:
+      annotations:
+        kuma.io/gateway: enabled
+        traffic.sidecar.istio.io/includeInboundPorts: ""
```
hbagdi, line 13: I don't this is what we want. These annotations will be injected into the pods that belong the Deployment but not to the deployment itself.

To add the annotations to the deployment itself, the annotations block should go into the metadata block at line 3 for this file.

hbagdi:
>  I noticed that the Deployment here doesn't specify any label like we do in our other installation method by having

Yes, this is intentional because we inject Helm managed label: `app.kubernetes.io/name`.

hbagdi:
Implemented in #81

PR #72, opened 2020-03-12
[kong] Refactor listener templates

rainest:
Consolidates listeners templates into a single generic template. This DRYs helper template code, makes the listener/service configuration consistent, and paves the way towards consolidating Service/Ingress templates in the future.

* Use the same listener/service configuration format in values.yaml for all Kong services. This mainly affects the admin API, which previously had a unique format. That format is still supported, but is deprecated. Users are warned to update it in NOTES.txt output if the old format is detected.
* Use the same template for generating all `KONG_<SERVICE>_LISTEN` variables.
* Add support for additional listener parameters (see previous discussion in #47).
* Combine documentation for various service values.yaml parameters.
* Add CI values.yaml for legacy admin API configuration.
* Move generation of the ingress controller `--kong-url` argument into a new `kong.adminLocalURL` helper.

PR contains un-squashed commits to show dev history, but should be squashed before merger, as most of the details are irrelevant after review. If merging, please use the commit message from https://github.com/Kong/charts/commit/6451e7f8ab28112f648d42ce7f7b49c581946aaf with one additional bullet point:

```
 * Add a "kong.adminLocalURL" helper to generate the local URL for the
   admin API, for use with the ingress controller.
```

#### Checklist

### DIFF
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -203,6 +203,12 @@ Kong can be configured via two methods:
 
 ### Kong parameters
 
+The various `SVC.*` parameters below are common to the various Kong services
+(the admin API, proxy, Kong Manger, the Developer Portal, and the Developer
+Portal API) and define their listener configuration, K8S Service properties,
+and K8S Ingress properties. Defaults are listed only if consistent across the
+individual services: see values.yaml for their individual default values.
+
 | Parameter                          | Description                                                                           | Default             |
 | ---------------------------------- | ------------------------------------------------------------------------------------- | ------------------- |
 | image.repository                   | Kong image                                                                            | `kong`              |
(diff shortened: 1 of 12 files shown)


charts/kong/values.yaml
```diff
@@ -45,22 +45,42 @@ image:
...
-  # if set to false also set readinessProbe and livenessProbe httpGet scheme's to 'HTTP'
-  useTLS: true
-  servicePort: 8444
-  containerPort: 8444
-  # Kong admin service type
+  http:
+    # Enable plaintext HTTP listen for the admin API
+    # Disabling this and using a TLS listen only is recommended for most configuration
+    enabled: false
+    servicePort: 8001
```
hbagdi, line 64: If not breaking, then these ports for all services should be `80` and `443`.
If breaking, let's take a note of this in #73.

charts/kong/values.yaml
```diff
@@ -45,22 +45,42 @@ image:
...
+    servicePort: 8444
+    containerPort: 8444
+    # Set a target port for the TLS port in the admin API service, useful when using TLS
+    # termination on an ELB.
+    # overrideServiceTargetPort: 8000
+    # Set a nodePort which is available if service type is NodePort
+    # nodePort: 32443
+    # Additional listen parameters, e.g. "reuseport", "backlog=16384"
+    parameters:
+    - http2
```
hbagdi, line 83: Is there a reason for adding `http2` in all blocks?
If copy-pasta only, then please remove `http2` from parameters at all places except proxy.

charts/kong/README.md
```diff
@@ -203,50 +203,45 @@ Kong can be configured via two methods:
 
 ### Kong parameters
 
+The various `SVC.*` parameters below are common to the various Kong services
+(the admin API, proxy, Kong Manger, the Developer Portal, and the Developer
+Portal API) and define their listener configuration, K8S Service properties,
+and K8S Ingress properties. Defaults are listed only if consistent across the
+individual services: see values.yaml for their individual default values.
+
```
hbagdi, line 231: 1. Let's split this out and create a new section in the readme with a table dedicated to these `svc.*` configuration.
2. Let's be very clear about what `svc` can be substituted by else this will trigger a lot of support request. I'd also recommend replacein `SVC` with `{SVC}` or some other notation to indicate that it needs to be substituted. Something more polished like:

`{SVC}` in the below table can be substite by:
- `proxy`
- `admin`
- `manager`
- and so on

charts/kong/README.md
```diff
@@ -341,22 +336,6 @@ For complete list of Kong configurations please check the
...
-This allows you to be more prescriptive when defining listen directives.
-
-**Note:** Overriding `env.proxy_listen` and `env.admin_listen` will
-potentially cause `admin.containerPort`, `proxy.http.containerPort` and
-`proxy.tls.containerPort` to become out of sync,
-and therefore must be updated accordingly.
-
-For example, updating to `env.proxy_listen: 0.0.0.0:4444, 0.0.0.0:4443 ssl`
-will need `proxy.http.containerPort: 4444` and `proxy.tls.containerPort: 4443`
-to be set in order for the service definition to work properly.
```
hbagdi, line 358: Is override not possible at all? From the current read, it seems it should be possible.
There will be use-cases that will demand override, and for those, we should mention this.

Thoughts?

charts/kong/values.yaml
```diff
@@ -102,6 +127,9 @@ proxy:
     # overrideServiceTargetPort: 8000
```
hbagdi, line 126: Any reason this port is not carried over to other services?
For folks who terminate TLS at the LB layer, they need this, else things usually break.

We recommend the Ingress Controller setup but there will be always a set of users who will need to terminate TLS at ELB.

charts/kong/templates/service-kong-admin.yaml
```diff
@@ -33,3 +34,55 @@ spec:
   selector:
     {{- include "kong.selectorLabels" . | nindent 4 }}
 {{- end -}}
+{{- else -}} {{/* Modern admin handler */}}
+{{- if and .Values.admin.enabled (or .Values.admin.http.enabled .Values.admin.tls.enabled) -}}
```
hbagdi, line 38: Let's move `if .Values.admin.enabled` out to the top.
This value is being checked in both the if else blocks.

charts/kong/templates/ingress-admin.yaml
```diff
@@ -29,4 +30,36 @@ spec:
     - {{ $hostname }}
     secretName: {{ $tls }}
   {{- end -}}
+{{- else -}} {{/* Modern admin handler */}}
+{{- $serviceName := include "kong.fullname" . -}}
+{{- $servicePort := include "kong.ingress.servicePort" .Values.admin -}}
+{{- $path := .Values.admin.ingress.path -}}
+{{- $tls := .Values.admin.ingress.tls -}}
+{{- $hostname := .Values.admin.ingress.hostname -}}
```
hbagdi, line 38: Let's not repeat these common properties here. Please move them under the main `if` block on line 1 instead, and re-use the variables for both the cases.

charts/kong/templates/_helpers.tpl
```diff
@@ -53,76 +53,59 @@ Create the name of the service account to use
 {{- end -}}
 
 {{/*
-Create the KONG_PROXY_LISTEN value string
+Create KONG_SERVICE_LISTEN strings
```
hbagdi, line 56: This will throw us in the future, let's add one liner below this to explain `SERVICE` here.

charts/kong/templates/_helpers.tpl
```diff
@@ -53,76 +53,59 @@ Create the name of the service account to use
...
+    {{- else -}}
+http://localhost:{{ .Values.admin.containerPort }}
+    {{- end -}}
+  {{- else -}}
+    {{- if .Values.admin.tls.enabled -}}
+https://localhost:{{ .Values.admin.tls.containerPort }}
+    {{- else if .Values.admin.http.enabled -}}
+http://localhost:{{ .Values.admin.http.containerPort }}
+    {{- else -}}
+http://localhost:9999 # You have an ingress controller enabled, but no admin listens!
```
hbagdi: Can we rephrase this as: "Ingress Controller will not work because there are no admin listeners enabled in Kong."?

charts/kong/templates/_helpers.tpl
```diff
@@ -360,56 +343,54 @@ the template that it itself is using form the above sections.
...
-{{- end }}
-
-{{- if and (.Values.portalapi.ingress.enabled) (.Values.enterprise.enabled) (.Values.enterprise.portal.enabled) }}
-  {{- $_ := set $autoEnv "KONG_PORTAL_API_URL" (include "kong.ingress.serviceUrl" .Values.portalapi.ingress) -}}
-{{- end }}
+{{- $_ := set $autoEnv "KONG_PROXY_LISTEN" (include "kong.listen" .Values.proxy) -}}
 
 {{- if .Values.enterprise.enabled }}
+  {{- $_ := set $autoEnv "KONG_ADMIN_GUI_LISTEN" (include "kong.listen" .Values.manager) -}}
+  {{- if and (.Values.manager.ingress.enabled) (.Values.enterprise.enabled) }}
```
hbagdi: Please remove `(.Values.enterprise.enabled) ` check from this and other if conditions in this block. The parent if block takes care of that already.

charts/kong/templates/_helpers.tpl
```diff
@@ -360,56 +343,54 @@ the template that it itself is using form the above sections.
...
+    {{- $_ := set $autoEnv "KONG_ADMIN_GUI_URL" (include "kong.ingress.serviceUrl" .Values.manager.ingress) -}}
+  {{- end -}}
+
   {{- if not .Values.enterprise.vitals.enabled }}
     {{- $_ := set $autoEnv "KONG_VITALS" "off" -}}
   {{- end }}
 
   {{- if .Values.enterprise.portal.enabled }}
     {{- $_ := set $autoEnv "KONG_PORTAL" "on" -}}
+        {{- $_ := set $autoEnv "KONG_PORTAL_GUI_LISTEN" (include "kong.listen" .Values.portal) -}}
```
hbagdi: Please fix the indentation here.

charts/kong/templates/_helpers.tpl
```diff
@@ -360,56 +343,54 @@ the template that it itself is using form the above sections.
...
+    {{- $_ := set $autoEnv "KONG_PORTAL_API_LISTEN" (include "kong.listen" .Values.portalapi) -}}
+
+    {{- if and (.Values.portal.ingress.enabled) (.Values.enterprise.enabled) (.Values.enterprise.portal.enabled) }}
```
hbagdi: Please remove `(.Values.enterprise.enabled) (.Values.enterprise.portal.enabled)` from this if block. This applies to another if statement below as well.

hbagdi (changes requested):
This is looking pretty good. Most of these are smaller comments.

charts/kong/values.yaml
rainest: Adding it generally shouldn't break anything, and should be beneficial for loading multiple assets from the Portal and Manager services.

The admin API does have it enabled by default in the latest kong.confs. The Enterprise services do not but I suspect that's an unintended omission; they lack the other new parameters also.

charts/kong/values.yaml
rainest: It is, since we can't rely on only chart components using them (e.g. the draft collector chart must specify the service hostname/port manually).

charts/kong/README.md
rainest: What cases are you thinking of in particular? The simplest I can think of (restricting the listen IP) doesn't seem like it'd be of use in containerized deployments.

While it's still possible, I couldn't think of obvious cases for overriding and figured it would be best left unmentioned.

charts/kong/values.yaml
rainest: It's only implemented for the proxy Service template. Consolidating those is a long-term goal, but out of scope for this.

charts/kong/templates/service-kong-admin.yaml
rainest: This is intentional--the idea is that these are otherwise complete admin Service templates within the `{{- if .Values.admin.containerPort -}}` cases, and that removing the legacy behavior will remove the first case (and legacy if block) entirely, leaving the contents of the second case alone.

charts/kong/templates/ingress-admin.yaml
rainest: Ditto comment from the service. The duplication is by design to make cutting out the legacy block easier later.

charts/kong/values.yaml
hbagdi: Concerns:
1. This path hasn't been tested enough so I'm not sure.
2. Make sure that the property is included in Kong Enterprise releases in the past year. Otherwise, this will be same as `reuseport` problem we ran into recently.

If the above are not an issue, I'm fine with merging this in and reverting if needed due to any other issue that comes up.

charts/kong/README.md
hbagdi: OSS users can get very creative sometimes.
We don't need a dedicated section but let's mention this in the env section itself maybe?

charts/kong/values.yaml
hbagdi: Okay, tracking this internally.

charts/kong/README.md
rainest: They can be, but it's easier to support them if they aren't. I don't want to document this even though it does work, since IMO it shouldn't appear that it's a standard way of configuring things.

If there are cases where the standard listen config doesn't work, I'd want those to come in via issues for us to try and see if there's a way we can support it without overrides.

hbagdi (approved):

PR #75, opened 2020-03-16
[kong] feat: allow to use replica set and allow to disable proxy service

aldor007:

This PR allows deploying kong using the spinnaker 

#### Checklist

### DIFF
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -224,6 +224,7 @@ Kong can be configured via two methods:
 | admin.ingress.hosts                | List of ingress hosts.                                                                | `[]`                |
 | admin.ingress.path                 | Ingress path.                                                                         | `/`                 |
 | admin.ingress.annotations          | Ingress annotations. See documentation for your ingress controller for details        | `{}`                |
+| proxy.enabled                      | Enables k8s service                                                                   | true                |
 | proxy.http.enabled                 | Enables http on the proxy                                                             | true                |
 | proxy.http.servicePort             | Service port to use for http                                                          | 80                  |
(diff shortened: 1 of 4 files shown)


hbagdi:
Can you explain why do you want to deploy Kong using a ReplicaSet instead of Deployment?

aldor007:
@hbagdi 
Because spinnaker can manage traffic only when using ReplicaSet.  In this mode, spinnaker can deploy new ReplicasaSet using Red/Black strategy  etc. 

Error message from spinnaker: 
> Spinnaker can manage traffic for one ReplicaSet only. Please deploy one ReplicaSet manifest or disable rollout strategies.

hbagdi:
@rainest Thoughts on the ReplicaSet part?

rainest:
Is this the latest version of Spinnaker? The PRs linked from https://github.com/spinnaker/spinnaker/issues/5025 indicate they've at least merged a change that allows multiple.

Does this work if you [add a history limit](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#revision-history-limit) to the Deployment template and set it to 0? At least on my own instance the old revisions are the only other RSes associated with the Deployment. If that works, I'd prefer it.

https://www.spinnaker.io/reference/providers/kubernetes/#deploy suggests that Spinnaker should be able to manage Deployments, so I think this is more that it thinks it needs to manage the old versions that aren't doing anything meaningful.

#72 has changes that handle disabling service creation already, so I'd request you remove that part.

aldor007:
Ok, thx.
I will test the new version of spinnaker, so now I'm closing this PR.

PR #79, opened 2020-03-19
[kong] add support for watch-namespace

wopol:
Adding support for watch-namespace in chart

### DIFF
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -290,19 +290,22 @@ The name of the service used for the ingress controller's validation webhook
   # Set the ingress class
   - --ingress-class={{ .Values.ingressController.ingressClass }}
   - --election-id=kong-ingress-controller-leader-{{ .Values.ingressController.ingressClass }}
-  # Set watch-namespace - limit ingress-controller to only one namespace 
-  {{- if .Values.ingressController.watchNamespace }}
-  - --watch-namespace={{ .Values.ingressController.watchNamespace }}
-  {{- end }}
   # the kong URL points to the kong admin api server
   {{- if .Values.admin.useTLS }}
   - --kong-url=https://localhost:{{ .Values.admin.containerPort }}
-  - --admin-tls-skip-verify # TODO make this configurable
+  - --admin-tls-skip-verify
   {{- else }}
   - --kong-url=http://localhost:{{ .Values.admin.containerPort }}
   {{- end }}
   {{- if .Values.ingressController.admissionWebhook.enabled }}
   - --admission-webhook-listen=0.0.0.0:{{ .Values.ingressController.admissionWebhook.port }}
+  {{- end }}
+
+
+  {{ if .Values.ingressController.args}}
+  {{- range $val := .Values.ingressController.args }}
+  - {{ $val }}
+  {{- end }}
   {{- end }}
   env:
   - name: POD_NAME
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -300,8 +300,6 @@ The name of the service used for the ingress controller's validation webhook
   {{- if .Values.ingressController.admissionWebhook.enabled }}
   - --admission-webhook-listen=0.0.0.0:{{ .Values.ingressController.admissionWebhook.port }}
   {{- end }}
-
-
   {{ if .Values.ingressController.args}}
   {{- range $val := .Values.ingressController.args }}
   - {{ $val }}
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -290,6 +290,10 @@ The name of the service used for the ingress controller's validation webhook
   # Set the ingress class
   - --ingress-class={{ .Values.ingressController.ingressClass }}
   - --election-id=kong-ingress-controller-leader-{{ .Values.ingressController.ingressClass }}
+  # Set watch-namespace - limit ingress-controller to only one namespace 
+  {{- if .Values.ingressController.watchNamespace }}
+  - --watch-namespace={{ .Values.ingressController.watchNamespace }}
+  {{- end }}
   # the kong URL points to the kong admin api server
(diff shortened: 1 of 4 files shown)


hbagdi:
Can you change this to instead inject arbitrary CLI arguments to the controller?
Something along the lines of taking in an array of strings and passing it verbatim to the controller.

We need to do this to make it easy enough for folks to override other properties via CLI flags.

wopol:
Yes, i can do this. Did you mean something like that?
```
ingressController:
  installCRDs: false
  ingressClass: myClass
  rbac:
    create: false
  serviceAccount:
    create: false
    name: sa-name
  args:
    - --watch-namespace=namespace-name
    - --other-arg=value
```
And one more question. What  with hardcoded arguments in `kong.controller-container` function  like `admin-tls-skip-verify`, `--admission-webhook-listen` etc.  Should these arguments stay there?

hbagdi:
> Should these arguments stay there?

No. Let's move them over to the values.yaml file as well so that a user can change them, if needed.

wopol:
> > Should these arguments stay there?
> 
> No. Let's move them over to the values.yaml file as well so that a user can change them, if needed.

Now this is a little bit more complicated, but we have all default arguments and we can replace them when we define the same argument in args Map.

hbagdi:
> Now this is a little bit more complicated, but we have all default arguments and we can replace them when we define the same argument in args Map.

In values.yaml, we can simply add and keep this simple:
```yaml
ingressController
  args:
  - --flag=value1
  - --bolean-flag
  - --flag2=value2
```

Users  can inject any flag and edit the default as well.

wopol:
> > Now this is a little bit more complicated, but we have all default arguments and we can replace them when we define the same argument in args Map.
> 
> In values.yaml, we can simply add and keep this simple:
> 
> ```yaml
> ingressController
>   args:
>   - --flag=value1
>   - --bolean-flag
>   - --flag2=value2
> ```
> 
> Users can inject any flag and edit the default as well.

I prefer simple solution but what if someone use arguments like that:
```
ingressController:
  ingressClass: other-class
  args:
    - --ingress-class=class-name
    - --publish-service=name
```

Which values should have higher priority for `--ingress-class` ingressController.ingressClass or ingressController.args[0] ?
Which values should have higher priority for `--publish-service` ingressController.args[1] or `{{ .Release.Namespace }}/{{ template "kong.fullname" . }}-proxy`

hbagdi:
> Which values should have higher priority for --ingress-class ingressController.ingressClass or ingressController.args[0] ?

Specifying via args should have a higher priority.

Some background:
We do want to add arg based approach, but we will recommend using environment variables for all configuration changes, especially to these ones that are already supported by environment variables.

Only logging based variables are not yet supported by environment variables, which is why we want to merge this PR in.
Flag/env should take precedence over specifying the value directly using a chart variable, like `ingressClass`.

wopol:
> Only logging based variables are not yet supported by environment variables, which is why we want to merge this PR in.
> Flag/env should take precedence over specifying the value directly using a chart variable, like `ingressClass`.

This will be more difficulty and less readable to check if in args array exist some key when not exist generate default e.g `election-id=kong-ingress-controller-leader-{{ .Values.ingressController.ingressClass }}`
I will try to do it.

charts/kong/templates/_helpers.tpl
```diff
@@ -293,13 +293,20 @@ The name of the service used for the ingress controller's validation webhook
...
-  - --admin-tls-skip-verify # TODO make this configurable
+  - --admin-tls-skip-verify
   {{- else }}
   - --kong-url=http://localhost:{{ .Values.admin.containerPort }}
   {{- end }}
   {{- if .Values.ingressController.admissionWebhook.enabled }}
   - --admission-webhook-listen=0.0.0.0:{{ .Values.ingressController.admissionWebhook.port }}
   {{- end }}
+
+
```
hbagdi: Please remove these line jumps.

hbagdi:
This looks good to me since going forward, users should add CLI args only for logging related settings.
The existing CLI flags will be moved over to environment variables after controller 0.8 is released, and we already have good amount of code on how to set precedence there.
@rainest thoughts?

charts/kong/templates/_helpers.tpl
wopol: Done

rainest:
@hbagdi you mean migrate `kong.ingressController.env` to something like `kong.env`'s dict-based deduplication system and handle all defaults via environment variables? That seems reasonable.

Post-0.8 is there any reason you'd still use CLI arguments over their equivalent variable? It almost seems like it'd make more sense to wait and have users use environment variables only, since we wouldn't have any means of avoiding conflicts between envvar/CLI settings.

hbagdi:
> dict-based deduplication system and handle all defaults via environment variables?

Yes.

> Post-0.8 is there any reason you'd still use CLI arguments over their equivalent variable? It almost seems like it'd make more sense to wait and have users use environment variables only, since we wouldn't have any means of avoiding conflicts between envvar/CLI settings.

Yes, because https://github.com/Kong/kubernetes-ingress-controller/issues/506.
Hence, the comment on logging CLI args only. Everything else will move to env vars.
We should move logging CLI args to env vars as well but that's not a priority (for me at least).

rainest (dismissed):
Ah, okay, I thought 0.8 was going to add support also.

hbagdi (changes requested):
This is looking good and we would like to merge it in.
Please also add a test case to an existing test (don't add another test file as it puts pressure on the CI).
Add a harmless argument like `--ingress-class=test-kong` or something like that. A logging arg would be even better. Full list here: https://github.com/Kong/kubernetes-ingress-controller/blob/master/docs/references/cli-arguments.md.


Thank you for pushing through this even though the purpose of the PR changed significantly.

hbagdi (approved):

hbagdi:
@wopol  Thank you for your contribution. Please fill in the contributor form to claim your Kong swag: https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#contributor-t-shirt.

PR #91, opened 2020-03-26
[kong] add deprecation notices to CHANGELOG.md

rainest:
Adds deprecation warnings to UPGRADE.md and adds an upgrade guide link to an old deprecation.

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -10,11 +10,12 @@
 
 ### Improvements
 
-* Service and listen configuration now use a unified configuration format.
-  Listen configuration now supports specifying parameters. Kubernetes service
-  creation can now be enabled or disabled for all Kong services. Users should
-  review the [1.4.0 upgrade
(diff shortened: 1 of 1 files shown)


hbagdi (commented):
Historically, I've followed https://keepachangelog.com/en and it does help a little bit.

Can we please add a `Deprecated` section at the top of each release where there is a deprecation to point this out clearly? Thanks.

rainest:
My rationale for this format (:warning: and deprecated setting alongside its accompanying update) was:
- It's the format we use for the main Kong project, and I like keeping consistency with the rest of the organization where possible.
- Most deprecations should be accompanied by new functionality that requires its own explanation. IMO it's an easier read to see both of those alongside one another rather than having a Deprecations section that just says `admin.containerPort is now deprecated` without context. Adding the same context again would be redundant.
- We do already call out deprecations by themselves with the NOTES.txt warnings. I prefer these since they cross versions: someone upgrading directly from 1.2 to 1.4 gets both at once, without having to check through each version's changelog.

With that in mind, would you still prefer a separate section?

hbagdi (approved):
I don't have strong objections here as long as we communicate the goal to our users.

PR #95, opened 2020-03-27
[kong] improve RBAC documentation

rainest:
Improves existing RBAC docs to better show how to set up the initial super-admin and allow the controller to talk to RBAC-secured instances.

* This does not discuss readiness/liveness checks, which also currently require tokens. Intent is to add a custom server block listen to approximate status_listen (and eventually replace it with the standard status_listen), removing the need for admin API status calls. This will be a separate PR.
* https://github.com/Kong/charts/commit/ed7e1b609e59a6784a23bd2f75d51fbb32df59cd further adds an example for the controller within values.yaml. None for Kong itself, since that variable should only be in place temporarily.

#### Checklist

### DIFF
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -474,14 +474,8 @@ env:
         key: CHANGEME-admin-token-key
 ```
 
-This value is not used after the initial install. It (and the associated
-secret) should be removed after.
-
-If using the ingress controller, the controller needs an RBAC token that grants
-at least workspace admin privileges in its workspace (`default` by default,
-settable by setting a `workspace` variable under `ingressController.env`). You
-should specify its token using `kong_admin_token` in your environment
-variables. For example:
+If using the ingress controller, it needs access to the token as well, by
+specifying `kong_admin_token` in its environment variables:
 
 ```yaml
 ingressController:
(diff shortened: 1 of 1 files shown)


hbagdi:
> This does not discuss readiness/liveness checks, which also currently require tokens. Intent is to add a custom server block listen to approximate status_listen (and eventually replace it with the standard status_listen), removing the need for admin API status calls. This will be a separate PR.

I might be missing something. Doesn't the existing server block configmap and metrics take care of the health checks? We don't rely on Admin API status checks.

charts/kong/README.md
```diff
@@ -453,21 +459,39 @@ for more details on these settings.
...
+env:
+ password:
+   valueFrom:
+     secretKeyRef:
+        name: CHANGEME-admin-token-secret
+        key: CHANGEME-admin-token-key
+```
+
+This value is not used after the initial install. It (and the associated
+secret) should be removed after.
```
hbagdi: Most deployments will start off with controller also using this same token.
This is not secure but that's how users can start with it.
Can we change the wordings in this and the next paragraph to communicate to the end user that you can start of with using a single token/password for migration and controller, but then you should take care of creating a dedicated credential for controller and update the secret accordingly.

hbagdi (changes requested):

hbagdi (approved):

PR #101, opened 2020-03-30
[kong] add configurable migration job annotations

### LINKED ISSUE #19: kong: Ability to add custom annotations to jobs
Reported in https://github.com/helm/charts/issues/19563 for Istio compatibility

rainest:
Add a migrationAnnotations setting to values.yaml with default annotations disabling the Istio and Kuma sidecars. Annotations are added to migration jobs.

Fixes #19 

All jobs receive the same set of annotations. For the original use case this seems desirable, and I couldn't immediately think of a reason to split them. It's easy enough to do so, but adds more config clutter and duplication, so absent a clear need I made them identical.

#### Checklist

### DIFF
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -234,6 +234,7 @@ Kong can be configured via two methods:
 | plugins                            | Install custom plugins into Kong via ConfigMaps or Secrets                            | `{}`                |
 | env                                | Additional [Kong configurations](https://getkong.org/docs/latest/configuration/)      |                     |
 | runMigrations                      | Run Kong migrations job                                                               | `true`              |
+| migrationAnnotations               | Annotations for migration jobs                                                        | `{"sidecar.istio.io/inject": "false", "kuma.io/sidecar-injection": "disabled"}` |
 | waitImage.repository               | Image used to wait for database to become ready                                       | `busybox`           |
 | waitImage.tag                      | Tag for image used to wait for database to become ready                               | `latest`            |
 | waitImage.pullPolicy               | Wait image pull policy                                                                | `IfNotPresent`      |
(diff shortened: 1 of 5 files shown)


rainest:
[Add support for Job-level annotations](https://trello.com/c/6ABeTXiU/352-add-support-for-job-level-annotations)

charts/kong/README.md
```diff
@@ -234,6 +234,7 @@ Kong can be configured via two methods:
 | plugins                            | Install custom plugins into Kong via ConfigMaps or Secrets                            | `{}`                |
 | env                                | Additional [Kong configurations](https://getkong.org/docs/latest/configuration/)      |                     |
 | runMigrations                      | Run Kong migrations job                                                               | `true`              |
+| migrationAnnotations               | Annotations for migration jobs                                                        | `{"sidecar.istio.io/inject": "false", "kuma.io/sidecar-injection": "disabled"}` |
```
hbagdi, line 237: Question:

Did you consider doing instead?
```yaml
migration:
  annotations: {}
```

We already have `runMigrations` as another top-level field.

rainest: If the migration configuration gets more complex I think it'd make sense to do so. My rationale was that it'd be weird to have
```
runMigrations: true
migrations:
  annotations: {}
```
with a split between some config under `migrations` and some under the root. We could move both under, but it'd be a breaking change or require that we support both `runMigrations` or `migrations.enabled`.

IMO two settings is small enough that I don't really mind both being at the root. Do you think we should restructure or now, or fine to let it wait until we have more in the future?

One other thing to consider is that the plans for split deployments touch this also, e.g. incorporating those we'd maybe have:
```
migrations:
  init: true
  pre-upgrade: true
  post-upgrade: true
  annotations: {}
```
That's more breaking, but it's needed anyway. Should we try for that, add a deprecation warning for runMigrations, and release both as part of 1.5?

hbagdi: This sounds good to me. Let's do this.

rainest:
Continued in https://github.com/Kong/charts/pull/102

PR #102, opened 2020-04-01
[kong] rework migration job configuration

### LINKED ISSUE #19: kong: Ability to add custom annotations to jobs

### LINKED ISSUE #43: "helm upgrade" fails due to init-migrations job conflict
Helm Version: `version.BuildInfo{Version:"v3.0.3", GitCommit:"ac925eb7279f4a6955df663a0128044a8a6b7593", GitTreeState:"clean", GoVersion:"go1.13.7"}`

Kong Version: Tested against 1.4 and 1.3 Enterprise with PostgresDB enabled

During a helm upgrade --install, with runMigrations set to true, the migration fails due to the init-migrations job. Helm 3 appears to be failing as the job specs are marked as immutable. This failure will still occur with no changes to the values.yaml between upgrades. 
Example command:
`helm upgrade --install kong/kong kong-test  -f ./values.yaml --debug`

Example error:
```bash 
Error: UPGRADE FAILED: cannot patch "kong-test-kong-init-migrations" with kind Job: Job.batch "kong-test-kong-init-migrations" is invalid: spec.template: Invalid value: core.PodTemplateSpec{ObjectMeta:v1.ObjectMeta... }: field is immutable
```

I was able to bypass this error by locally adding `runInitMigrations: false` to the values.yaml file and adding it to the and conditions used to create the job in the template.

rainest:
* Create a new `migrations` section in values.yaml.
* Add configurable annotations to migration jobs, with default annotations to disable Istio and Kuma sidecars.
* Splits `runMigrations` into per-migration toggles under `migrations`.
* Points NOTES.txt deprecation warnings to the relevant UPGRADE.md section rather than explaining the issue in place.
* Add documentation for split deployments and init conflict issue.

This addresses several different issues:
* Migration annotations address an issue with sidecar containers blocking job completion (#19).
* Avoiding the init job conflict (#43)
* Allowing users to disable pre/post-upgrade migrations is needed for deployments that split different role nodes across multiple releases.

Fixes #19
Fixes #43 

This supersedes https://github.com/Kong/charts/pull/101

#### Checklist

### DIFF
--- a/charts/kong/FAQs.md
+++ b/charts/kong/FAQs.md
@@ -84,7 +84,8 @@ is immutable
 This occurs if a `RELEASE-NAME-kong-init-migrations` Job is left over from a
 previous `helm install` or `helm upgrade`. Deleting it with
 `kubectl delete job RELEASE-NAME-kong-init-migrations` will allow the upgrade
-to proceed.
+to proceed. Setting `migrations.init: false` will prevent the job from spawning
+and creating this issue for future upgrades.
 
 This job is not removed automatically [due to limitations in
 Helm](https://github.com/Kong/charts/pull/36).
--- a/charts/kong/FAQs.md
+++ b/charts/kong/FAQs.md
@@ -84,8 +84,4 @@ is immutable
 This occurs if a `RELEASE-NAME-kong-init-migrations` Job is left over from a
 previous `helm install` or `helm upgrade`. Deleting it with
 `kubectl delete job RELEASE-NAME-kong-init-migrations` will allow the upgrade
-to proceed. Setting `migrations.init: false` will prevent the job from spawning
-and creating this issue for future upgrades.
-
-This job is not removed automatically [due to limitations in
-Helm](https://github.com/Kong/charts/pull/36).
+to proceed. Chart versions greater than 1.5.0 delete the job automatically.
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -33,6 +33,7 @@ $ helm install kong/kong --generate-name --set ingressController.installCRDs=fal
   - [Database](#database)
   - [Runtime package](#runtime-package)
   - [Configuration method](#configuration-method)
+  - [Separate admin and proxy nodes](#separate-admin-and-proxy-nodes)
 - [Configuration](#configuration)
   - [Kong Parameters](#kong-parameters)
     - [Kong Service Parameters](#kong-service-parameters)
@@ -95,6 +96,9 @@ $ helm install kong/kong
 # Helm 3
 $ helm install kong/kong --generate-name --set ingressController.installCRDs=false
 ```
+
+After installing, set `migrations.init: false` in values.yaml to prevent the
+initial migrations from spawning during upgrades.
 
 ## Uninstall
 
@@ -201,6 +205,45 @@ The package to run can be changed via `image.repository` and `image.tag`
 parameters. If you would like to run the Enterprise package, please read
 the [Kong Enterprise Parameters](#kong-enterprise-parameters) section.
 
+### Separate admin and proxy nodes
+
+Users may wish to split their Kong deployment into multiple instances that only
+run some of Kong's services, e.g. where some nodes only run the proxy and other
+only run the admin API, or where some nodes only run Developer Portal services.
+At present, these require separate Helm releases (i.e. you run `helm install`
+once for every instance type you wish to create).
+
+To disable Kong services on an instance, you should set `SVC.enabled`,
+`SVC.http.enabled`, `SVC.tls.enabled`, and `SVC.ingress.enabled` all to
+`false`, where `SVC` is `proxy`, `admin`, `manager`, `portal`, or `portalapi`.
+
+The standard chart upgrade automation process assumes that there is only a
+single Kong release in the Kong cluster, and runs both `migrations up` and
+`migrations finish` jobs. To handle clusters split across multipl releases, you
+should:
+1. Upgrade one of the releases with `helm upgrade RELEASENAME -f values.yaml
+   --set migrations.preUpgrade=true --set migrations.postUpgrade=false`.
+2. Upgrade all but one of the remaining releases with `helm upgrade RELEASENAME
+   -f values.yaml --set migrations.preUpgrade=false --set
+   migrations.postUpgrade=false`.
+3. Upgrade the final release with `helm upgrade RELEASENAME -f values.yaml
+   --set migrations.preUpgrade=false --set migrations.postUpgrade=true`.
+
+This ensures that all instances are using the new Kong package before running
+`kong migrations finish`.
+
+Users should note that Helm supports supplying multiple values.yaml files,
+allowing you to separate shared configuration from instance-specific
+configuration. For example, you may have a shared values.yaml that contains
+environment variables and other common settings, and then several
+instance-specific values.yamls that contain service configuration only. You can
+then create releases with:
+
+```
+helm install proxy-only -f shared-values.yaml -f only-proxy.yaml kong/kong
+helm install admin-only -f shared-values.yaml -f only-admin.yaml kong/kong
+```
+
 ### Configuration method
 
 Kong can be configured via two methods:
@@ -233,7 +276,10 @@ Kong can be configured via two methods:
 | replicaCount                       | Kong instance count                                                                   | `1`                 |
 | plugins                            | Install custom plugins into Kong via ConfigMaps or Secrets                            | `{}`                |
 | env                                | Additional [Kong configurations](https://getkong.org/docs/latest/configuration/)      |                     |
-| runMigrations                      | Run Kong migrations job                                                               | `true`              |
+| migrations.init                    | Run "kong migrations bootstrap" jobs                                                  | `true`              |
+| migrations.preUpgrade              | Run "kong migrations up" jobs                                                         | `true`              |
+| migrations.postUpgrade             | Run "kong migrations finish" jobs                                                     | `true`              |
+| migrations.annotations             | Annotations for migration jobs                                                        | `{"sidecar.istio.io/inject": "false", "kuma.io/sidecar-injection": "disabled"}` |
 | waitImage.repository               | Image used to wait for database to become ready                                       | `busybox`           |
 | waitImage.tag                      | Tag for image used to wait for database to become ready                               | `latest`            |
 | waitImage.pullPolicy               | Wait image pull policy                                                                | `IfNotPresent`      |
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -216,8 +216,8 @@ To disable Kong services on an instance, you should set `SVC.enabled`,
 
 The standard chart upgrade automation process assumes that there is only a
 single Kong release in the Kong cluster, and runs both `migrations up` and
-`migrations finish` jobs. To handle clusters split across multipl releases, you
-should:
+`migrations finish` jobs. To handle clusters split across multiple releases,
+you should:
 1. Upgrade one of the releases with `helm upgrade RELEASENAME -f values.yaml
    --set migrations.preUpgrade=true --set migrations.postUpgrade=false`.
 2. Upgrade all but one of the remaining releases with `helm upgrade RELEASENAME
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -97,9 +97,6 @@ $ helm install kong/kong
 $ helm install kong/kong --generate-name --set ingressController.installCRDs=false
 ```
 
-After installing, set `migrations.init: false` in values.yaml to prevent the
-initial migrations from spawning during upgrades.
-
 ## Uninstall
 
 To uninstall/delete a Helm release `my-release`:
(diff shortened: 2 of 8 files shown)


charts/kong/README.md
```diff
@@ -201,6 +205,45 @@ The package to run can be changed via `image.repository` and `image.tag`
...
+Users should note that Helm supports supplying multiple values.yaml files,
+allowing you to separate shared configuration from instance-specific
+configuration. For example, you may have a shared values.yaml that contains
+environment variables and other common settings, and then several
+instance-specific values.yamls that contain service configuration only. You can
+then create releases with:
+
+```
+helm install proxy-only -f shared-values.yaml -f only-proxy.yaml kong/kong
+helm install admin-only -f shared-values.yaml -f only-admin.yaml kong/kong
```
hbagdi, line 244: Why is this included in this PR?

charts/kong/values.yaml
```diff
@@ -183,8 +183,21 @@ plugins: {}
 # - kong-admin-tls
 secretVolumes: []
 
-# Set runMigrations to run Kong migrations
-runMigrations: true
+# Enable/disable migration jobs, and set annotations for them
+migrations:
+  # Enable initial migrations/"kong migrations bootstrap"
+  # This should be set to false after the initial install
+  init: true
```
hbagdi: Is it possible to detect if a rendering process is an install or a delete inside the template?
If so, we can further automate this. If not, NBD.

charts/kong/values.yaml
```diff
@@ -183,8 +183,21 @@ plugins: {}
...
+  # Enable pre-upgrade migrations/"kong migrations up"
+  preUpgrade: true
+  # Enable post-upgrade migrations/"kong migrations finish"
```
hbagdi: The `/` is weird here. Can we use words instead so as to not throw off users?
Something along the lines:
Enable post-upgrade migrations (equivalent to "kong migrations finish")

This applies to other two instances as well.

charts/kong/templates/migrations.yaml
```diff
@@ -1,11 +1,15 @@
...
 kind: Job
 metadata:
   name: {{ template "kong.fullname" . }}-init-migrations
   labels:
     {{- include "kong.metaLabels" . | nindent 4 }}
     app.kubernetes.io/component: init-migrations
+  annotations:
+    {{- range $key, $value := .Values.migrationAnnotations }}
+    {{ $key }}: {{ $value | quote }}
+    {{- end }}
```
hbagdi: This always gets interesting. I think the original motivation is to have these annotations on the pods that are created part of the Job (spec.template.metadata.annotations). I'm not sure about this. Can you look into Istio's injection mechanism for jobs and correct this if needed?

charts/kong/UPGRADE.md
```diff
@@ -41,7 +49,40 @@ text ending with `field is immutable`. This is typically due to a bug with the
...
+individual migrations:
+
+```
+migrations:
+  init: true
+  preUpgrade: true
+  postUpgrade: true
+  annotations:
+    sidecar.istio.io/inject: false
+    kuma.io/sidecar-injection: "disabled"
```
hbagdi: Is it necessary to specify annotations here?
Most users don't need to worry about these annotations, and hence I'd recommend removing that part here.
Minimizing the amount of information that a user needs to digest during an upgrade is worthwhile.

charts/kong/UPGRADE.md
```diff
@@ -41,7 +49,40 @@ text ending with `field is immutable`. This is typically due to a bug with the
...
+The new format addresses several needs:
+* The initial migrations job can be disabled after the initial install,
+  preventing [conflicts on upgrades](https://github.com/Kong/charts/blob/master/charts/kong/FAQs.md#running-helm-upgrade-fails-because-of-old-init-migrations-job).
+* The upgrade migrations jobs can be disabled as need for managing
+  [multi-release clusters](https://github.com/Kong/charts/blob/master/charts/kong/README.md#separate-admin-and-proxy-nodes).
+  This enables management of clusters that have nodes with different roles,
+  e.g. nodes that only run the proxy and nodes that only run the admin API.
+* Migration jobs now allow specifying annotations, and provide a default set
+  of annotations that disable some service mesh sidecars. Because sidecar
+  containers do not terminate, they prevent the jobs from terminating.
```
hbagdi: sidecar container do not terminate what?

hbagdi: It might be better to rephrase this.

charts/kong/README.md
```diff
@@ -201,6 +205,45 @@ The package to run can be changed via `image.repository` and `image.tag`
 parameters. If you would like to run the Enterprise package, please read
 the [Kong Enterprise Parameters](#kong-enterprise-parameters) section.
 
+### Separate admin and proxy nodes
+
+Users may wish to split their Kong deployment into multiple instances that only
+run some of Kong's services, e.g. where some nodes only run the proxy and other
+only run the admin API, or where some nodes only run Developer Portal services.
+At present, these require separate Helm releases (i.e. you run `helm install`
```
hbagdi: Let's frame this as being a positive rather than a limitation.

charts/kong/README.md
```diff
@@ -201,6 +205,45 @@ The package to run can be changed via `image.repository` and `image.tag`
 parameters. If you would like to run the Enterprise package, please read
 the [Kong Enterprise Parameters](#kong-enterprise-parameters) section.
 
+### Separate admin and proxy nodes
+
+Users may wish to split their Kong deployment into multiple instances that only
+run some of Kong's services, e.g. where some nodes only run the proxy and other
+only run the admin API, or where some nodes only run Developer Portal services.
```
hbagdi, line 209: or, in case of Kong Enterprise, some nodes only run Developer Portal services.

charts/kong/FAQs.md
```diff
@@ -84,7 +84,8 @@ is immutable
 This occurs if a `RELEASE-NAME-kong-init-migrations` Job is left over from a
 previous `helm install` or `helm upgrade`. Deleting it with
 `kubectl delete job RELEASE-NAME-kong-init-migrations` will allow the upgrade
-to proceed.
+to proceed. Setting `migrations.init: false` will prevent the job from spawning
+and creating this issue for future upgrades.
```
hbagdi: Question: If `migration.init` is set to `false` during an upgrade, will helm delete the migration job that was created during installation because it no longer sees that resource as part of the release?

hbagdi (changes requested):

charts/kong/templates/migrations.yaml
rainest: Nope, I just wasn't looking where I was adding it. I also forgot to change the variable name from the original `migrationAnnotations` :woman_facepalming:

charts/kong/README.md
rainest: "At present" was meant more in the sense that it may change in the future, but isn't really necessary. I removed it, though that just leaves a neutral phrase. Not sure of any particular way to cast it in a positive light (mainly because we don't really have concrete alternatives to compare it to).

charts/kong/values.yaml
rainest: It is! And it handles deleting it automatically too!

charts/kong/README.md
rainest: IMO this is already self-evident: a node that can run Portal services is necessarily Enterprise. Adding the clarification feels like it makes the sentence more awkward, so I've left this as-is.

rainest:
20f853f527cd639476ba39aa8a577ed4ff59ea77 is mostly doc changes requested above. It has one major functionality change: the init job handling is now fully automated and avoids the conflict issue. It should be squashed into 857215eb5a72c9f0bdbc0d2eabe3c3b1b93a6829 with an update to the original message.

Edit: `git commit --fixup` does not work how I think.  6913d3d should be squashed also.

I tried to make `migrations.init` effectively a hidden setting, i.e. it's not listed in values.yaml but does exist, and if set to `false` in values.yaml would indeed disable that job. For some reason that's not clear to me, the default in the template appears to always apply even when that is set, so the job cannot be disabled.

I've left the default in there since it'd otherwise rely on `runMigrations` only, with the net effect that the job is _never_ disabled, even if you have `runMigrations: false`. However, with the addition of the `.Release.IsInstall` condition, I don't think this should be a concern: there really shouldn't be any reason to ever disable the init migrations job on install, and worst case (you have an already-initialized database), it simply exits without doing anything.

charts/kong/README.md
```diff
@@ -201,6 +202,45 @@ The package to run can be changed via `image.repository` and `image.tag`
...
+These require separate Helm releases (i.e. you run `helm install` once for
+every instance type you wish to create).
+
+To disable Kong services on an instance, you should set `SVC.enabled`,
+`SVC.http.enabled`, `SVC.tls.enabled`, and `SVC.ingress.enabled` all to
+`false`, where `SVC` is `proxy`, `admin`, `manager`, `portal`, or `portalapi`.
+
+The standard chart upgrade automation process assumes that there is only a
+single Kong release in the Kong cluster, and runs both `migrations up` and
+`migrations finish` jobs. To handle clusters split across multipl releases, you
```
hbagdi: Can we change to: "To handle Kong cluster split across multiple releases"?

charts/kong/values.yaml
```diff
@@ -183,8 +183,18 @@ plugins: {}
 # - kong-admin-tls
 secretVolumes: []
 
-# Set runMigrations to run Kong migrations
-runMigrations: true
+# Enable/disable migration jobs, and set annotations for them
+migrations:
```
hbagdi, line 187: Question: Do we need to clarify with a doc line saying this is only valid for Kong installations backed by DB?
The reason I ask is because it is not totally obvious from just reading this.

rainest: Not really--the job templates already disable themselves if DB-less is in use, and no changes are required to this config for it.

hbagdi (approved):

chenguang-Steve:
Hi @hbagdi @rainest 
I see the below changes are closed and not included in latest release.  I want to trigger migrations-post-upgrade.yaml and migrations-pre-upgrade.yaml when using helm install. But currently these are only triggered by helm upgrade. Is it possible add another hook annotations like  helm.sh/hook: "pre-install" in migrations-pre-upgrade.yaml and helm.sh/hook: "post-install"  in migrations-pre-upgrade.yaml

annotations:
    helm.sh/hook: "post-upgrade"
    helm.sh/hook-delete-policy: "before-hook-creation"
    {{- range $key, $value := .Values.migrationAnnotations }}
    {{ $key }}: {{ $value | quote }}
    {{- end }}

rainest:
Could you elaborate? That's by design, as there's no reason for `migrations up` and `migrations finish` to roun outside the context of an upgrade.

Installs are handled using a different command (`migrations bootstrap`) with its own Job: https://github.com/Kong/charts/blob/master/charts/kong/templates/migrations.yaml. That's essentially always enabled (it does in fact have its own enable/disable toggle, but we don't list it in values.yaml, as it should never be necessary to disable that Job).

chenguang-Steve:
Hi, thanks for your quick response.
The kong is integrated into my project. When project execute "stop" commands, then stop kong(helm delete), when project start then start kong(helm install)
The steps:
1. project start(helm install kong),  migrations command execute successfully. (migrations bootstrap)
2. project stop(helm delete kong)
3. Add additional plugin migration.
4. project start(helm install kong)
   The kong failed to start because of below errors.

waiting for db
Error: /usr/local/share/lua/5.1/kong/cmd/start.lua:64: nginx: [error] init_by_lua error: /usr/local/share/lua/5.1/kong/cmd/utils/migrations.lua:20: New migrations available; run 'kong migrations up' to proceed
stack traceback:
	[C]: in function 'error'
	/usr/local/share/lua/5.1/kong/cmd/utils/migrations.lua:20: in function 'check_state'
	/usr/local/share/lua/5.1/kong/init.lua:411: in function 'init'
	init_by_lua:3: in main chunk
lua:3: in main chunk


I guess when add new plugin migrations based on already bootstrapped  , we have to execute "migrations up"? 
That is why I want execute to pre-update job when install.  But this may does not work.
So for my case, possibly always execute helm install and helm upgrade when start my project?

rainest:
@chenguang-Steve yeah--that's an atypical strategy we wouldn't normally expect to support: we'd normally expect users to have any custom plugins present when running `helm install` (which would then run their migrations during the `kong bootstrap` job) or to run `helm upgrade` when adding new plugins that require migrations.

Is there any reason you perform a full delete and re-install for this? I'd normally expect a `helm upgrade` in the scenario you describe, and don't see any reason it wouldn't achieve the same end.

You could probably perform a full delete and re-install when adding the additional plugin if you also reset your database or delete its PVC after deletion (note: **don't** do this if you aren't using the ingress controller or decK to manage configuration--it will destroy all configuration added manually through the admin API). However, it normally shouldn't be necessary to do so--an upgrade should suffice for most use cases. If there's a case where it doesn't, I'd be curious to know--it may be something we can handle better with improvements elsewhere.

PR #103, opened 2020-04-03
[kong] add support for stream listens

rainest:
Adds support for stream listens as an additional listen type on the proxy.

I'm still running into issues with the balancer when attempting to configure these:

```
2020/04/03 20:45:38 [error] 37#0: *37954 lua entry thread aborted: runtime error: /usr/local/share/lua/5.1/kong/runloop/balancer.lua:569: bad argument #1 to 'ipairs' (table expected, got nil)
stack traceback:
coroutine 0:
	[C]: in function 'ipairs'
	/usr/local/share/lua/5.1/kong/runloop/balancer.lua:569: in function 'get_all_upstreams'
	/usr/local/share/lua/5.1/kong/runloop/balancer.lua:612: in function 'get_upstream_by_name'
	/usr/local/share/lua/5.1/kong/runloop/balancer.lua:638: in function 'get_balancer'
	/usr/local/share/lua/5.1/kong/runloop/balancer.lua:919: in function 'execute'
	/usr/local/share/lua/5.1/kong/runloop/handler.lua:930: in function 'balancer_execute'
	/usr/local/share/lua/5.1/kong/runloop/handler.lua:1097: in function 'after'
	/usr/local/share/lua/5.1/kong/init.lua:622: in function 'preread'
	preread_by_lua(nginx-kong-stream.conf:66):2: in main chunk while prereading client data, client: 10.138.15.225, server: 0.0.0.0:9001
```
I'd run into them previously with TCPIngress testing and got them again when configuring manual test config for this. That definitely warrants investigation, but I don't think it's anything to do with the PR: the PR code is only setting up the listens and making them accessible, and that does work.

#### Checklist

### DIFF
--- a/charts/kong/ci/test3-values.yaml
+++ b/charts/kong/ci/test3-values.yaml
@@ -26,12 +26,3 @@ dblessConfig:
             message: "dbless-config"
 proxy:
   type: NodePort
-# - add stream listens
-  stream:
-  - containerPort: 9000
-    servicePort: 9000
-    parameters: []
-  - containerPort: 9001
-    servicePort: 9001
-    parameters:
-    - ssl
--- a/charts/kong/ci/test3-values.yaml
+++ b/charts/kong/ci/test3-values.yaml
@@ -1,4 +1,4 @@
-# CI test for testing dbless deployment without ingress controllers and stream listens
+# CI test for testing dbless deployment without ingress controllers
 # - disable ingress controller
 ingressController:
   enabled: false
--- a/charts/kong/ci/test3-values.yaml
+++ b/charts/kong/ci/test3-values.yaml
@@ -1,4 +1,4 @@
-# CI test for testing dbless deployment without ingress controllers
+# CI test for testing dbless deployment without ingress controllers and stream listens
 # - disable ingress controller
 ingressController:
   enabled: false
@@ -26,3 +26,12 @@ dblessConfig:
             message: "dbless-config"
 proxy:
   type: NodePort
+# - add stream listens
+  stream:
+  - containerPort: 9000
+    servicePort: 9000
+    parameters: []
+  - containerPort: 9001
+    servicePort: 9001
+    parameters:
+    - ssl
--- a/charts/kong/ci/test4-values.yaml
+++ b/charts/kong/ci/test4-values.yaml
@@ -1,4 +1,4 @@
-# CI test for testing dbless deployment without ingress controllers using legacy admin listen
+# CI test for testing dbless deployment without ingress controllers using legacy admin listen and stream listens
 # TODO: remove legacy admin listen behavior at a future date
 # - disable ingress controller
 ingressController:
@@ -34,3 +34,12 @@ dblessConfig:
             message: "dbless-config"
 proxy:
   type: NodePort
+# - add stream listens
+  stream:
+  - containerPort: 9000
+    servicePort: 9000
+    parameters: []
+  - containerPort: 9001
+    servicePort: 9001
+    parameters:
+    - ssl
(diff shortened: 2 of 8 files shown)


charts/kong/templates/_helpers.tpl
```diff
@@ -57,26 +57,43 @@ Create KONG_SERVICE_LISTEN strings
...
-    {{- end -}}
+    {{/*
+    This is a bit of a hack to support always including "ssl" in the parameter
+    list for TLS listens. It's not possible to set a variable to an object from
+    .Values and then modify one of the objects values locally, although
+    https://github.com/helm/helm/issues/4987 indicates it should be. Instead,
+    this creates a new object and new parameters list built from the original.
+    */}}
+    {{- $tls := dict -}}
+    {{- $parameters := append .tls.parameters "ssl" -}}
```
hbagdi, line 76: And what happens when `ssl` is already present in the list?

hbagdi (changes requested):

charts/kong/values.yaml
```diff
@@ -130,6 +130,20 @@ proxy:
...
+  # ports and parameters.
+  stream: {}
+    # - name: example-stream
+    #   containerPort: 9000
+    #   servicePort: 9000
+    #   # Optionally set a static nodePort if the service type is NodePort
+    #   # nodePort: 32080
+    #   # Additional listen parameters, e.g. "ssl", "reuseport", "backlog=16384"
+    #   # "ssl" is not supported on versions <2.0, but is not required for
+    #   # SNI-based listens. It is used to reject any non-TLS connections.
```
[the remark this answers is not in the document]
rainest: I misread http://nginx.org/en/docs/stream/ngx_stream_core_module.html#listen re requiring TLS. Not sure about our support for it. The changelog suggests that it was only added as of 2.0, whereas https://docs.konghq.com/1.3.x/admin-api/#add-route indicates it is supported. Changelog is probably more trustworthy, so assuming 2.0+ is the only option unless I hear otherwise from the core team.

charts/kong/values.yaml
```diff
@@ -130,6 +130,20 @@ proxy:
     parameters:
     - http2
 
+  # Define stream (TCP) listen
+  # To enable, remove "{}", uncomment the section below, and select your desired
+  # ports and parameters.
+  stream: {}
+    # - name: example-stream
```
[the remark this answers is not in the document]
rainest: It makes the example look awkward since everything else has explanatory comments before :)

In practice maybe, as it's used in the SRV record Kubernetes constructs. Probably not too crucial though. It's easy enough to make it default to an auto-generated name with an optional user-specified name if needed later, so defaulting to autogenerated for now.

hbagdi:
Please update an existing test to include a stream listen in DB and DB-less mode. Otherwise, looks good.

charts/kong/ci/test4-values.yaml
```diff
@@ -34,3 +34,12 @@ dblessConfig:
             message: "dbless-config"
 proxy:
   type: NodePort
+# - add stream listens
```
hbagdi, line 37: Why are these added to two DB-less deployments? Can we move one over to a DB deployment?
It would also be best if we can keep one with IC and one without.

Total 2 combinations.

rainest: Because I cannot read comments good :(

charts/kong/ci/test3-values.yaml
```diff
@@ -1,4 +1,4 @@
-# CI test for testing dbless deployment without ingress controllers
+# CI test for testing dbless deployment without ingress controllers and stream listens
```
hbagdi: Please remove this diff before merging in.

hbagdi (approved):
One comment, please correct that and then merge.
Please squash and merge.

hbagdi (approved):

PR #107, opened 2020-04-08
[kong] add status listen

rainest:
Add the status listen, which provides a dedicated status endpoint on versions 1.4+.

kong.conf does not provide a TLS listen in the example. The chart adds one with a somewhat arbitrary port choice.

#### Checklist


rainest:
`ssl` support is lacking certificate directives in nginx_kong.lua. We should consider whether to fix that first.

rainest:
Force-pushed to reconcile conflicts with next after merging stream listen.

Having thought over this a bit more, I think we should default HTTPS to false if we want to release this for 2.0.x and adhere to what we claim in the appVersion. Even a future minor release adds HTTPS support, I'm wary of introducing changes that break startup on releases that do fall under what the chart indicates support for. If we want to default to true, we should move towards supporting that in 2.1 and defer this until then.

Added a core backlog item to discuss that.

charts/kong-collectorapi/glum-lizard.yaml
```diff
@@ -0,0 +1,51 @@
+image:
+  repository: gcr.io/formidable-app-197923/kong-brain-immunity
```
hbagdi: Please remove these collector related files.

charts/kong/README.md
```diff
@@ -300,6 +300,11 @@ individual services: see values.yaml for their individual default values.
 * `manager`
 * `portal`
 * `portalapi`
+* `status`
+
+Unlike most services, `status` cannot be exposed outside the pod, and cannot
+create a Kubernetes service or ingress. It supports the settings under
+`SVC.http` and `SVC.tls` only.
```
hbagdi: > Unlike most services

Unlike all other services

> cannot be exposed outside the pod, and cannot
create a Kubernetes service or ingress

Can we instead frame this as "The chart doesn't allow exposing the endpoint as outside the Kubernetes cluster because the endpoint is meant for intra-cluster concerns only.".

charts/questions.txt
```diff
@@ -0,0 +1,7 @@
+* Would there be any reason not to generate a service for each listen? Bit cluttered, but whatever.
+* Generic listens present a challenge: there is a static list of possible X_listen parameters, and shouldn't really be user-configurable. We don't have any way to prevent users from just removing or changing them. Probably better to keep dedicated blocks and maybe reuse templates underneath.
```
hbagdi: Please remove this file.

charts/kong-collectorapi/collector-notes.txt
```diff
@@ -0,0 +1,24 @@
+# Non-agnostic README/NOTES items
+* `open` is OS X-only.
+* Should not assume Minikube.
+* Should not assume open NodePort. We're guilty of this too--we should probably add a port-forward.
```
hbagdi: Please remove this file.

charts/kong/templates/_helpers.tpl
```diff
@@ -109,7 +109,7 @@ Create a single listen (IP+port+parameter combo)
 {{- define "kong.singleListen" -}}
   {{- $listen := list -}}
   {{- $listen = append $listen (printf "0.0.0.0:%d" (int64 .containerPort)) -}}
-  {{- range $param := .parameters | uniq }}
+  {{- range $param := .parameters | default (list) | uniq }}
```
hbagdi, line 112: What is the reason for this change?
Does this error out if parameters is undefined?

charts/kong/values.yaml
```diff
@@ -101,6 +101,20 @@ admin:
...
+  http:
+    # Enable plaintext HTTP listen for the status listen
+    enabled: true
+    containerPort: 8100
+
+  tls:
+    # Enable HTTPS listen for the status listen
+    # Kong does not currently support HTTPS status listens, so this should remain false
+    enabled: false
+    containerPort: 8143
```
hbagdi: nit: Since Kong adds 443 to a port for TLS, let's change this to 8543?

hbagdi (changes requested):

charts/kong/templates/_helpers.tpl
rainest: Yeah, see the last commit/CI log for the commit prior. Range over null generates an error.

charts/kong/README.md
rainest: There's no service creation either--it's not accessible outside the pod, barring things like readiness/liveness checks. Added the why in front, but I think we should keep the specifics of what it does--K8S-saavy users should recognize what no service/no ingress means in practice.

charts/kong/README.md
```diff
@@ -300,6 +300,12 @@ individual services: see values.yaml for their individual default values.
 * `manager`
 * `portal`
 * `portalapi`
+* `status`
+
+`status` is intended for internal use within the Kong pod only. Unlike other
```
hbagdi: > internal use within the Kong pod only

I think you meant within the internal Kubernetes cluster only.

hbagdi: status exposes health and metrics endpoints that are for in-cluster services only.

hbagdi (approved):

PR #110, opened 2020-04-10
[kong] correct serviceMonitor.interval docs

artburkart:
#### What this PR does / why we need it:

The scraping interval is not set in the `values.yaml` file of the Helm
chart. As such, the [default scraping][1] interval is used. This is
[actually 30s][2] and not 10s.

[1]:
https://github.com/coreos/prometheus-operator/blob/7ebf899f9768ead08ac7af6d59688d642cb8ef9c/pkg/prometheus/promcfg.go#L627-L629
[2]:
https://github.com/coreos/prometheus-operator/blob/7ebf899f9768ead08ac7af6d59688d642cb8ef9c/pkg/prometheus/promcfg.go#L175-L189

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [x] PR is based off the current tip of the `next` branch and targets `next`, not `master`
- [x] New or modified sections of values.yaml are documented in the README.md
- [x] Title of the PR and commit headers start with chart name (e.g. `[kong]`)

PR #112, opened 2020-04-10
[kong] target proxy with ServiceMonitor selector

artburkart:
This commit limits the ServiceMonitor, so it only targets the proxy
Service resource instead of both the proxy and the admin Service.

When we set `serviceMonitor.enabled` to `true`, a `ServiceMonitor` resource is created, which targets both the admin `Service` and the proxy `Service`. The metrics are identical, which means they're doubled in in Prometheus. This change makes it so the `ServiceMonitor` only targets the proxy `Service`.

I didn't know which label we'd think would be appropriate, and since naming is hard, I opted to throw a strawman label into this PR, so we could discuss what we want here.

#### Checklist

### DIFF
--- a/charts/kong/templates/service-kong-proxy.yaml
+++ b/charts/kong/templates/service-kong-proxy.yaml
@@ -8,6 +8,7 @@ metadata:
       {{ $key }}: {{ $value | quote }}
     {{- end }}
   labels:
+    enable-metrics: "true"
     {{- include "kong.metaLabels" . | nindent 4 }}
 spec:
   type: {{ .Values.proxy.type }}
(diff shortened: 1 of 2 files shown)


rainest (changes requested):
Same merge conflict mentioned in https://github.com/Kong/charts/pull/110#issuecomment-612236855

Could you also use a descriptive lablel? e.g.

```
enable-metrics: true
```
versus
```
foobar-label: enables-metrics
```

artburkart:
Reopening per comment here: https://github.com/Kong/charts/pull/113#issuecomment-617980654

rainest (approved):

PR #113, opened 2020-04-11
[kong] add additionalMatchLabels to ServiceMonitor

artburkart:

When we set `serviceMonitor.enabled` to true, a `ServiceMonitor`
resource is created, which targets both the admin `Service` and the
proxy `Service`. The metrics are identical, which means they're doubled
in in Prometheus.

This change makes it so it's possible to configure a label on the proxy
`Service`, which is used in the `matchLabels` selector on the
`ServiceMonitor`.

The reason this is optional is because setting it by default would break
reverse-compatibility for folks who may already be alerting on metrics
coming from the admin `Service`.

#### Checklist

### DIFF
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -404,8 +404,9 @@ For a complete list of all configuration values you can set in the
 | serviceMonitor.enabled             | Create ServiceMonitor for Prometheus Operator                                         | false               |
 | serviceMonitor.interval            | Scrapping interval                                                                    | 10s                 |
 | serviceMonitor.namespace           | Where to create ServiceMonitor                                                        |                     |
-| secretVolumes                      | Mount given secrets as a volume in Kong container to override default certs and keys. | `[]`                |
 | serviceMonitor.labels              | ServiceMonito Labels                                                                  | {}                  |
+| serviceMonitor.additionalMatchLabels | Add matchLabels selector to ServiceMonitor and corresponding label to proxy Service | `{}`                |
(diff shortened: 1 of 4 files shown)


rainest (changes requested):
Should this be configurable? Can you provide an example scenario where a configurable label provides functionality beyond a static label?

IMO the original static labeling from https://github.com/Kong/charts/pull/112/commits/c66a121efd2ada50c91cf7b5e9779f471d9a42cf was better: functionally there shouldn't be any difference between matching labels in the template itself and matching labels rendered from the same values.yaml. The static label is simpler in the template and doesn't add a values.yaml setting that's unlikely to change.

Same request regarding the label naming from #112: that name/value should be descriptive (e.g. `enable-metrics: true`), such that someone taking a glance at the proxy service alone can determine why the label is present.

artburkart:
My apologies for letting this hang out for so long. My concern is that if someone set up alerts based on the Prometheus metrics labelled with "admin", this change would break their alert setup. I wanted to avoid that. Given this concern, do we still want to make this reverse-incompatible change?

rainest:
I think so, yes. Impact shouldn't be too severe (possible loss of metrics, but no interruption of traffic or change in proxy behavior) and remediation should be simple (change the label over). We can reasonably cover those in UPGRADE.md, and I think the benefit of not requiring manual configuration in the future outweighs the cost of the one-time change for any configurations broken by this.

PR #116, opened 2020-04-14
[kong] Allow the proxy container's lifecycle hooks to be overridden

rifelpet:
Allows the proxy container's lifecycle hooks to be overridden. In order to prevent dropped requests during the shutdown process, it can help to sleep first before signaling the shutdown to the container process. This gives `kube-proxy` and other ingress controllers time to remove the IP of the terminating pod from their iptables rules or load balancing configurations. There was a KubeCon NA talk that demonstrated this: [video](https://youtu.be/0o5C12kzEDI?t=329) + [slides](https://static.sched.com/hosted_files/kccncna19/f7/The%20Gotchas%20of%20Zero-Downtime%20Traffic.pdf)

  - fixes # N/A

#### Checklist

### DIFF
--- a/charts/kong/values.yaml
+++ b/charts/kong/values.yaml
@@ -397,6 +397,13 @@ livenessProbe:
   successThreshold: 1
   failureThreshold: 3
 
+# Proxy container lifecycle hooks
+# Ref: https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/
+lifecycle:
+  preStop:
+    exec:
+      command: ["/bin/sh", "-c", "/bin/sleep 15 && kong quit"]
+
 # Affinity for pod assignment
 # Ref: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity
 # affinity: {}
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -389,6 +389,7 @@ For a complete list of all configuration values you can set in the
 | updateStrategy                     | update strategy for deployment                                                        | `{}`                |
 | readinessProbe                     | Kong readiness probe                                                                  |                     |
 | livenessProbe                      | Kong liveness probe                                                                   |                     |
+| lifecycle                          | Proxy container lifecycle hooks                                                       | see `values.yaml`   |
 | affinity                           | Node/pod affinities                                                                   |                     |
 | nodeSelector                       | Node labels for pod assignment                                                        | `{}`                |
 | deploymentAnnotations              | Annotations to add to deployment                                                      |  see `values.yaml`  |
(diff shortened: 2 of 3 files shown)


charts/kong/values.yaml
```diff
@@ -397,6 +397,13 @@ livenessProbe:
   successThreshold: 1
   failureThreshold: 3
 
+# Proxy container lifecycle hooks
+# Ref: https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/
+lifecycle:
+  preStop:
+    exec:
+      command: ["/bin/sh", "-c", "sleep 30 && kong quit"]
```
hbagdi: Default value of `terminationGracePeriodSeconds` is 30 seconds and `kong quit` can itself take some time.
Can we change this to be 15 or 20 seconds?

rifelpet: oops, I accidentally committed the sleep, I meant to keep the value identical to the chart's current value. I just updated it to 15 seconds if you do want to keep that in there.

hbagdi: I think it is OK to sleep for 15 seconds so that other controllers/LBs have enough time to stop sending requests to Kong. 

@rainest @artburkart Thoughts?

artburkart: I think making it configurable is the biggest win here. If the default value proves to be too much or too little, the user of the chart is empowered to configure it according to their needs.

That said, 15 seconds seems like a reasonable amount to me and including it as part of the chart serves as a cue to users that it's something they may want to tweak. 

I'll need to test it to provide more substantial feedback. Odds are, I'll try it out this week. Thanks for surfacing this!

rainest (approved):
I'm on the fence as to whether this should indeed be configurable--it seems like something that'd have a reasonable universal setting that just requires some testing to confirm what value achieves the desired effect.

Providing more stripped-down example values.yamls is a future goal though, so the increasing size and complexity of the actual base values.yaml won't be as much of a concern.

hbagdi (approved):

artburkart:
I see that it's been merged, but to speak to the point raised by @rainest, it seems this also opens up the opportunity to put something other than a `sleep` into the value. 

My immediate thought when I saw this was that maybe one could `curl` another endpoint that was reporting info on their system and then only execute `kong quit` once that endpoint said it was safe to do. I've had cases, for example, where I've had to drain off 100s of thousands of web socket connections over a controlled period of time that was significantly longer than 15 seconds.

hbagdi:
@rifelpet Please fill in the contributor form to claim your Kong swag: https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#contributor-t-shirt

PR #119, opened 2020-04-15
[kong] enable kic by default

hbagdi:
#### Checklist


rainest (dismissed):

rainest (changes requested):
This should be within an if block for `ingressController.enabled`, shouldn't it? This will report that the controller is on for any release.

It'd need to be removed from the `env` section also, since that doesn't reflect whether the controller is enabled, and will take precedence over the setting in the template.

rainest (approved):

PR #121, opened 2020-04-21
[kong] add the ability to include adhoc pod labels

aspring:
This PR adds the ability to include adhoc pod labels.

#### Checklist
- [X] PR is based off the current tip of the `next` branch and targets `next`, not `master`

### DIFF
--- a/charts/kong/templates/deployment.yaml
+++ b/charts/kong/templates/deployment.yaml
@@ -36,6 +36,9 @@ spec:
       labels:
         {{- include "kong.metaLabels" . | nindent 8 }}
         app.kubernetes.io/component: app
+        {{- if .Values.podLabels }}
+        {{ toYaml .Values.podLabels | nindent 8 }}
+        {{- end }}
     spec:
       {{- if .Values.priorityClassName }}
       priorityClassName: "{{ .Values.priorityClassName }}"
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -393,6 +393,7 @@ For a complete list of all configuration values you can set in the
 | nodeSelector                       | Node labels for pod assignment                                                        | `{}`                |
 | deploymentAnnotations              | Annotations to add to deployment                                                      |  see `values.yaml`  |
 | podAnnotations                     | Annotations to add to each pod                                                        | `{}`                |
+| podLabels                          | Labels to add to each pod                                                             | `{}`                |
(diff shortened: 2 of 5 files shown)


hbagdi:
I'm not against it but @aspring, can you elaborate on your use-case for this?

aspring:
We run Kong with Istio and by adding labels to the pod spec they can be utilized by Istio for enhancing tracing as well as visualizations.

hbagdi:
Interesting. I figured it would always be via annotations but I think label makes sense for this.

charts/kong/templates/deployment.yaml
```diff
@@ -36,6 +36,9 @@ spec:
       labels:
         {{- include "kong.metaLabels" . | nindent 8 }}
         app.kubernetes.io/component: app
+        {{- if .Values.podLabels }}
+        {{ toYaml .Values.podLabels | nindent 8 }}
+        {{- end }}
```
hbagdi, line 41: Please update an existing test file to include a test case.

hbagdi (changes requested):

rainest (approved):
This should probably make use of nested items in values.yaml that touch the various places that don't have their own annotation/label sections. Roughly:

```
metadata:
  deployment:
    labels:
    - foo
    - bar
  pod:
    labels:
    - foo
    - bar
    annotations:
      example.com/example: whatever
```
However given the existing `podAnnotations` we'd need to make a breaking change to support it. I don't think that's pressing though, and this is consistent with the existing format, so fine as-is.

charts/kong/templates/deployment.yaml
aspring: This should be all set.  The following snippet from the test showing the app and environment labels added.

```bash
Node:         chart-testing-control-plane/172.17.0.2
Start Time:   Fri, 24 Apr 2020 03:22:06 +0000
Labels:       app=kong
              app.kubernetes.io/component=app
              app.kubernetes.io/instance=kong-5n1v8lzb7n
              app.kubernetes.io/managed-by=Tiller
              app.kubernetes.io/name=kong
              app.kubernetes.io/version=2
              environment=test
```

hbagdi (approved):

hbagdi:
@aspring Please claim your contributor swag: https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#contributor-t-shirt. Thank you for your contribution.

PR #123, opened 2020-04-24
[kong] persistent volumes and local plugins

psulowicz:
Add support for:
* mounting existing PVC as volumes in Kong Container
* loading plugins located on Kong container (e.g. on mounted PVC)

The motivation for this functionality was the need
to load plugins written in Golang, which are too large
to be stored in ConfigMap or Secret.
It can also be used for other purposes, e.g. injecting
complex nginx configuration snippets.

#### Checklist

### DIFF
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -274,7 +274,7 @@ Kong can be configured via two methods:
 | image.pullPolicy                   | Image pull policy                                                                     | `IfNotPresent`      |
 | image.pullSecrets                  | Image pull secrets                                                                    | `null`              |
 | replicaCount                       | Kong instance count                                                                   | `1`                 |
-| plugins                            | Install custom plugins into Kong via ConfigMaps or Secrets                            | `{}`                |
+| plugins                            | Install custom plugins into Kong via ConfigMaps, Secrets or from file system          | `{}`                |
 | env                                | Additional [Kong configurations](https://getkong.org/docs/latest/configuration/)      |                     |
 | migrations.preUpgrade              | Run "kong migrations up" jobs                                                         | `true`              |
 | migrations.postUpgrade             | Run "kong migrations finish" jobs                                                     | `true`              |
@@ -402,6 +402,7 @@ For a complete list of all configuration values you can set in the
 | podSecurityPolicy.spec             | Collection of [PodSecurityPolicy settings](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#what-is-a-pod-security-policy) | |
 | priorityClassName                  | Set pod scheduling priority class for Kong pods                                       | `""`                |
 | secretVolumes                      | Mount given secrets as a volume in Kong container to override default certs and keys. | `[]`                |
+| persistentVolumes                  | Mount existing PVC as a volume in Kong container to inject extra files e.g. custom plugins. | `[]`                |
 | serviceMonitor.enabled             | Create ServiceMonitor for Prometheus Operator                                         | `false`             |
 | serviceMonitor.interval            | Scraping interval                                                                     | `30s`               |
 | serviceMonitor.namespace           | Where to create ServiceMonitor                                                        |                     |
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -253,6 +253,11 @@ The name of the service used for the ingress controller's validation webhook
   secret:
     secretName: {{ . }}
 {{- end }}
+{{- range .Values.persistentVolumes }}
+- name: {{ .claimName }}
+  persistentVolumeClaim:
+    claimName: {{ .claimName }}
+{{- end }}
 {{- end -}}
 
 {{- define "kong.volumeMounts" -}}
@@ -292,6 +297,11 @@ The name of the service used for the ingress controller's validation webhook
   readOnly: true
 {{- end }}
 {{- end }}
+{{- range .Values.persistentVolumes }}
+- name: {{ .claimName }}
+  mountPath: {{ .mountPath }}
+  readOnly: true
+{{- end }}
 {{- end -}}
 
 {{- define "kong.plugins" -}}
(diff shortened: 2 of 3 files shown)


hbagdi:
I can see the overall motivation for the patch.

Questions:
- What is your workflow to set this up in a new k8s cluster?
- How do you provision the PV?
- How is the volumen populated with the necessary data? Who maintains it?
- Should the PV be part of the chart?
- What is the accessMode being used?

My concerns is users in future will wish for more things to live on PVs because of this, which further complicates the chart quite a bit.
Furthermore, this makes delivering Kong to multiple k8s clusters much more complicated than it should be. You need to setup new resources, missing dependencies will lead to undocumented behaviors, and lead to increase in pain points and a sharp drop in UX of Kong.

**Have you considered building Docker images to include the Go plugins in the image itself?** That is the recommended path for custom plugins, and is much safer, reproducible, and, maintainable.

The original motivation for introducing the feature of mounting of plugins via secrets and configmaps was to ease development and enable the use-case for simpler plugins. This has significantly grown over the past few months, something I was concerned about adding in the first place as well.

rainest:
Ultimately, I agree that this is better-handled by building a custom image. Using PVs opens a lot of questions that are difficult to answer in a generic way, and we don't really need them: we're not actually performing any writes that need to persist, at least not for the plugin use case. Custom images are simple to build (`FROM kong; do whatever`) and private container registries are usually readily available.

Easy-ish answers, at least from my POV:

> What is your workflow to set this up in a new k8s cluster?
> How do you provision the PV?

As-is we can kind of leave these open: as best I can tell, the expectation is that this is largely left up to the user, and it should be possible to handle without input from us. Storage management isn't in our domain of concern.

> What is the accessMode being used?
In keeping with how we handle ConfigMaps, these should be read-only if used for plugins. One alternate use I can think of is providing persistent volumes for log data (if you dislike stdout/stderr handling), however.

Less easy, and more important:
> How is the volumen populated with the necessary data? Who maintains it?
> Should the PV be part of the chart?

Here we run into issues. As written this allows users to specify a single claim used for the entire release. This works well if you have a single replica and never run upgrade jobs, but becomes rapidly less workable otherwise: Kubernetes does not very much like starting pods that use a PVC if it's already in use and doesn't specify a multi-access mode. Those exist, but aren't consistently supported (AWS notably lacks any): https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes

GCP ostensibly supports them, but I ran into issues with what I thought should work:

```
$ kubectl get pvc trash-test-volume -o yaml |grep -A2 status
status:
  accessModes:
  - ReadOnlyMany
$ kubectl describe po example-kong-kong-5d697ff4b4-fg2kh | grep FailedMount
  Warning  FailedMount         <invalid> (x2 over 108s)  kubelet, gke-kong-cse-cluster-default-pool-9ee5fab6-4xtv  Unable to mount volumes for pod "example-kong-kong-5d697ff4b4-fg2kh_default(377fcc89-5c14-4414-afc9-44f5deca33f3)": timeout expired waiting for volumes to attach or mount for pod "default"/"example-kong-kong-5d697ff4b4-fg2kh". list of unmounted volumes=[trash-test-volume]. list of unattached volumes=[example-kong-kong-prefix-dir example-kong-kong-tmp custom-nginx-template-volume example-kong-admin-example-tls trash-test-volume example-kong-kong-token-sthf6]
```
ReadWriteMany has even more sparse support, and if it's not available, updating the volume contents becomes difficult--you'd presumably need to initialize a new volume and replace all pods to use it.

Depending on your storage provider, you may or may not be able to share the PVC between instances. If you can, the answer to these is "you figure it out" and "probably no", but if you can't you'd need some way to create a PV per-pod and copy data into it. At that point, you may as well use an emptyDir, since nothing in the volume actually needs to persist. Populating it via LuaRocks is an option, but that doesn't handle private plugins or Go plugins.

Lastly, setting `KONG_LUA_PACKAGE_PATH` is necessary to use non-standard plugin file locations. We can handle this okay with the ConfigMap mounter, since it's fairly limited and always uses `/opt/kong`, but there's no way to handle it for an unknown mount point with an unknown structure. A custom build that dumps everything the standard `/usr/local/share/lua/5.1` path (and furthermore under `/usr/local/share/lua/5.1/kong/plugins` for the plugins themselves) is simple enough and sidesteps that issue. This isn't the case for Go plugins, but it's worth considering.

psulowicz:
> What is your workflow to set this up in a new k8s cluster?
> How do you provision the PV?

We configure new k8s cluster using ansible playbooks and in this initial setup we create required PVC etc.

> How is the volumen populated with the necessary data? Who maintains it?

Volume is populated by Jenkins Pipeline running k8s job and triggered after PR with plugin code is merged.
Dedicated pipeline is also responsible for invoke Helm to upgrade proper Kong instances.

> Should the PV be part of the chart?

I don't think so. Creating and populating volume should not be a chart responsibility.

> What is the accessMode being used?

Volume is mounted within POD in ReadOnly mode and pipelines are organised in such a way that they do not overwrite existing files.

---
We were thinking about dedicated Docker images, but idea with ConfigMap and Secrets seemed much more comfortable and elegant for us and eventually convinced us to do this in this way.

I understand yours concerns about risk of using PVs for everything and problems that may arise from this. So, what do you think about limit this feature only for plugins ?
Let's say, we change 'persistentVolumes' to something like 'pluginPVC' and mount it in dedicated, hardcoded path (similar to ConfigMap/Secrets approach).

After all PV is only a way to inject something to container (used in many various charts) and this injected thing have to be still attached to application in supported way.

psulowicz:
Hi Guys, i thought about it for a while, and solution i proposed in last comment (pluginPVC) is not flexible enough for our needs. So, if current PR does not fit in chart vision I will decline it and we will reorganize our flow to use custom docker images.

hbagdi:
@psulowicz Thank you for understanding the issues that might arise because of this feature.

We will hold off on this feature for now but we can always reconsider in future based on feedback that we receive from you and other users.

PR #125, opened 2020-04-24
Use internal address for admin API when appropriate

rainest:
If the admin service is disabled, but the admin listens are enabled (i.e. the configuration we expect for controller-managed instances), use `127.0.0.1` for the admin listen address to prevent access from outside the pod.

Additionally:
* If a service's service is disabled, but its listens are enabled, do not render port information for that service in the Kong container template.
* Listens now support specifying the address in general in the template, but currently only the admin API makes use of this, as I'm not aware of any situation where it's useful for the other services--although they support the same service disabled/listens enabled configuration, we shouldn't encounter that in the wild, and they don't have the same security concerns as the admin API.

#### Checklist

### DIFF
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -60,7 +60,10 @@ Generic tool for creating KONG_PROXY_LISTEN, KONG_ADMIN_LISTEN, etc.
   {{- $unifiedListen := list -}}
 
   {{- if .http.enabled -}}
-    {{- $httpListen := (include "kong.singleListen" .http) -}}
+    {{- $listenConfig := dict -}}
+    {{- $listenConfig := merge $listenConfig .http -}}
+    {{- $_ := set $listenConfig "address" (default "0.0.0.0" .address) -}}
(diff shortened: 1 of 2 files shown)


hbagdi (approved):
I'm a little concerned about disabling exposing ports when there are no k8s service for a component (manager, portal etc). Could there be a case where someone doesn't want a k8s service but wants to expose the ports? While technically possible, the likelihood is extremely low given how these components interact.

Nice work with the patch and also the PR description. The description very clear and helped so much in understanding the changes.

rainest:
Hard to say. Per https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#container-v1-core this is informational, and by that token they probably should be declared if the listen exists. The ports are always exposed if the listen exists, and removing this just kinda pretends they aren't.

That said, I have limited practical concerns about it:
* At worst, it results in inaccurate display information. The port still functions normally.
* For those services, there shouldn't be any actual reason to disable the service while keeping the listens, other than possibly in some embedded context where Kong lives within another app's pod.

PR #127, opened 2020-04-30
kong added smtp poperty auth type and ssl

satwant17:
What this PR does / why we need it:
User can add the auth type for smtp server in the smtp section of the helm chart. Certain providers like send grid will need a auth type of plain or login to function properly.  Also enable the ssl if needed.

Which issue this PR fixes
NA

Special notes for your reviewer:
Updated the ReadMe file section of SMTP.

#### Checklist
- [ ] PR is based off the current tip of the `next` branch and targets `next`, not `master`

### DIFF
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -599,10 +599,9 @@ Setting `.enterprise.smtp.disabled: true` will set `KONG_SMTP_MOCK=on` and
 allow Admin/Developer invites to proceed without sending email. Note, however,
 that these have limited functionality without sending email.
 
-If your SMTP server requires authentication, you should the `username` and
-`smtp_password_secret` keys under `.enterprise.smtp.auth`.
-`smtp_password_secret` must be a Secret containing an `smtp_password` key whose
-value is your SMTP password.
+If your SMTP server requires authentication, you must provide the `username` and `smtp_password_secret` keys under `.enterprise.smtp.auth`. `smtp_password_secret` must be a Secret containing an `smtp_password` key whose value is your SMTP password.
+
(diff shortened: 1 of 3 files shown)


hbagdi:
@satwant17 Please author your commit using your konghq.com email address and then please sign the CLA.

charts/kong/README.md
```diff
@@ -599,7 +599,7 @@ Setting `.enterprise.smtp.disabled: true` will set `KONG_SMTP_MOCK=on` and
 allow Admin/Developer invites to proceed without sending email. Note, however,
 that these have limited functionality without sending email.
 
-If your SMTP server requires authentication, you should the `username` and
+If your SMTP server requires authentication,you should provide `smtp_auth_type` default value is `nil` but for some smtp server you will need to provide auth type  `plain` or `login`.Also you should provide the `username` and
```
rainest: Missing spaces after `,` and `.`

Would also rework this a bit: the defaults is a bit misleading because it doesn't truly reflect the underlying library behavior. Per https://github.com/GUI/lua-resty-mail#new that truly defaults to `nil` only if no credentials are provided (because you're not actually authenticating); if you do provide a username and password it defaults to `plain`. I suggest:

> If your SMTP server requires authentication, you must provide the `username` and `smtp_password_secret` keys under `.enterprise.smtp.auth`. `smtp_password_secret` must be a Secret containing an `smtp_password` key whose
value is your SMTP password.
>
> By default, SMTP uses `AUTH PLAIN` when you provide credentials. If your provider requires `AUTH LOGIN`, set `smtp_auth_type: login`.

rainest (changes requested):
Template good, docs should be clarified a bit since the underlying library defaults aren't immediately obvious from what we pass it.

PR #128, opened 2020-05-04
[kong] add settings to specify images with a single string

rainest:
Adds support for specifying the Kong and Kong Ingress Controller images via a single string.

This is necessary to satisfy offline operator constraints. We need a way to override images anyway to specify the versions on RHCC, so may as well kill two birds with one PR.

See https://github.com/Kong/kong-operator/blob/fork/operatorhub/watches.yaml#L7-L9 and https://github.com/Kong/kong-operator/blob/fork/operatorhub/olm/0.2.6/kong.v0.2.6.clusterserviceversion.yaml#L142-L145 for draft usage.

These are intentionally omitted from documentation, as they are not intended for general usage, only by the operator.

#### Checklist

### DIFF
--- a/charts/kong/ci/single-image-default.yaml
+++ b/charts/kong/ci/single-image-default.yaml
@@ -1,0 +1,15 @@
+# install chart with default values
+# use single image strings instead of repository/tag
+
+image:
+  single: kong:2.0
+proxy:
+  type: NodePort
+
+env:
+  anonymous_reports: "off"
+ingressController:
+  env:
+    anonymous_reports: "false"
+  image:
+    single: kong-docker-kubernetes-ingress-controller.bintray.io/kong-ingress-controller:0.8.1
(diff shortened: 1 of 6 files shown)


hbagdi:
1. A quick `grep` reveals that there is `waitImage` as well. Can we include this change for that too?
2. `single` is a literal name for such a thing. Can we name the new thing along the lines of `unifiedRepoTag` or `repoTag`?

rainest:
> A quick grep reveals that there is waitImage as well. Can we include this change for that too?

It's there, although it was broken prior to 44e4a75.

> single is a literal name for such a thing. Can we name the new thing along the lines of unifiedRepoTag or repoTag?

Done. This is all pretty obfuscated from the user though: they control these via environment variables on the operator deployment.

Additional fix for the controller using the wrong condition.

hbagdi (approved):

PR #134, opened 2020-05-13
[kong] add example values.yaml configurations

rainest:
Add example values.yamls with basic settings for common configurations of Kong and Kong Enterprise. See README.md for info on each.

Tested by performing a fresh install of each. All do come online if all mandatory secrets are populated.

This is, however, far from a one-step install of everything: the full Kong Enterprise example requires 16 overrides to work. The rest aren't so bad, but if we want to reduce those steps we need changes in the product itself--Helm most likely isn't going to be able to easily automate any of this effectively.

#### Checklist


charts/kong/example-values/minimal-kong-controller.yaml
```diff
@@ -0,0 +1,12 @@
+# The minimum values.yaml configuration necessary for Kong with the ingress controller
```
hbagdi: This is the default, why are we including this?

charts/kong/example-values/minimal-kong-standalone.yaml
```diff
@@ -0,0 +1,19 @@
+# The minimum values.yaml configuration necessary for Kong without the ingress controller, using the Postgres subchart
+image:
+  repository: kong
+  tag: "2.0"
```
hbagdi, line 9: Why do you need to set these if this is "minimal"? These are defaults, right?

charts/kong/example-values/minimal-kong-standalone.yaml
```diff
@@ -0,0 +1,19 @@
...
+  prefix: /kong_prefix/
+  database: postgres
+
+postgresql:
+  enabled: true
+  postgresqlUsername: kong
+  postgresqlDatabase: kong
+  service:
+    port: 5432
+
```
hbagdi, line 28: Do you not need to deploy an "admin" service in this case? Else, how are you supposed to configure Kong?

charts/kong/example-values/minimal-kong-enterprise-k8s.yaml
```diff
@@ -0,0 +1,18 @@
+
+
+image:
+  repository: kong-docker-kong-enterprise-k8s.bintray.io/kong-enterprise-k8s
+  tag: 2.0.4.1-centos
+  pullSecrets:
+    - you-must-create-a-registry-secret
```
hbagdi: Let's rename this to the standard secret names that we use in our documentation.
Why?
1. Makes it ready to use; most enterprise users follow the guide.
2. We can include a comment if needed.
3. The ultimate goal here would be to then include a link to this file in the `helm` install section of [this](https://github.com/Kong/kubernetes-ingress-controller/blob/master/docs/deployment/k4k8s-enterprise.md#helm) doc. The current link in that doc is out of date.

This comment applies to all the files in this PR.

charts/kong/example-values/minimal-k4k8s-enterprise.yaml
```diff
@@ -0,0 +1,18 @@
...
+enterprise:
+  enabled: true
+  license_secret: you-must-create-a-license-secret
+
+env:
+  database: "off"
+
+ingressController:
+  enabled: true
+  installCRDs: false
```
hbagdi, line 22: This seems wrong. You shouldn't even need `ingressController` section in any of the setups which include the controller.

hbagdi: I see. Is this for Helm 3 compat? Have we considered Helm 2 vs Helm 3?

charts/kong/example-values/README.md
```diff
@@ -0,0 +1,20 @@
+# Example values.yaml configurations
+
+* minimal-kong-controller.yaml installs Kong open source with the ingress
```
hbagdi: There are six files but you only list 4 here, is that intentional?

charts/kong/example-values/README.md
```diff
@@ -0,0 +1,20 @@
...
+  controller in DB-less mode.
+
+* minimal-kong-standalone.yaml installs Kong open source and Postgres with no
```
hbagdi: Can we also link to the file from this readme?

charts/kong/example-values/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+* minimal-k4k8s-with-kong-enterprise.yaml installs Kong for Kubernetes with
+  Kong Enterprise with the ingress controller and PostgreSQL. It does not
+  enable Enterprise services (Kong Manager, the Dev Portal, etc.) and does not
+  expose the admin API outside the Pod.  full-k4k8s-with-kong-enterprise.yaml
+  installs Kong for Kubernetes with Kong Enterprise with the ingress controller
+  in PostgreSQL. It enables all Enterprise services.
+
+All Enterprise examples require some level of additional user configuration to
+install properly. Read the comments at the top of each file for instructions.
```
hbagdi, line 27: Any reason we don't include a link to this file from the main readme?

hbagdi (changes requested):

charts/kong/example-values/minimal-kong-standalone.yaml
rainest: Consistency, for the most part--different images aren't something we're able to represent in the main values.yaml, so that's a major facet of what these examples set.

These aren't really intended to be "the absolute minimum diff from the base values.yaml" but more just minimalist configurations that don't set up much. Can rename them to "basic" instead if that makes more sense.

charts/kong/example-values/minimal-k4k8s-enterprise.yaml
rainest: That and in line with the earlier comment re the not quite minimal meaning of minimal here--even though this is the default, it should call out the major things it's doing explicitly so that you don't have to cross-reference.

As far as Helm 2, I'm working from the assumption that most new installs should (hopefully) be using Helm 3. I'm not really targeting Helm 2 for first-class support across the board, only where it'd break the template entirely for Helm 2. Will note as much in the examples README.

charts/kong/example-values/minimal-kong-standalone.yaml
rainest: Yeah, good catch--this and minimal-k4k8s-with-kong-enterprise.yaml need it (the latter to make Manager useful). However, they're in kind of a weird place since the services aren't exactly accessible unless we turn them into LoadBalancer types. Debating whether to do that since it will usually be non-free, but simpler, or instruct people to use port-forwards. `KONG_ADMIN_API_URI` is a bit of an issue also, since we can't rely on ingress auto-populating it.

charts/kong/example-values/minimal-k4k8s-enterprise.yaml
hbagdi: In that case, all docs should have `helm install kong/kong --values <url>` for helm3 and then for helm2, `helm install kong/kong --values <url> --set <key>=value`

hbagdi:
Some comments have been addressed so I'm not reviewing this at this moment.

rainest:
Yeah--will re-request review once all are handled.

charts/kong/example-values/minimal-k4k8s-enterprise.yaml
rainest: Showing `-f <filepath>` instead: half of these won't work as-is, so I don't want to suggest otherwise with the URL.

charts/kong/example-values/minimal-kong-standalone.yaml
rainest: Went with port-forwards. This conveniently lets us ignore `KONG_ADMIN_API_URI` and `KONG_ADMIN_GUI_URL`, since they're all localhost!

charts/kong/example-values/minimal-kong-standalone.yaml
rainest: Re-worded comments, filenames left as-is: I forget if renaming them makes it harder to read the diff. Please review the remainder, and if that looks good, I'll rename them "basic" instead of "minimal".

charts/kong/example-values/minimal-k4k8s-enterprise.yaml
```diff
@@ -0,0 +1,21 @@
...
+image:
+  repository: kong-docker-kong-enterprise-k8s.bintray.io/kong-enterprise-k8s
+  tag: 2.0.4.1-centos
+  pullSecrets:
+    # you-must add this: https://github.com/Kong/charts/blob/master/charts/kong/README.md#kong-enterprise-docker-registry-access
+    - kong-enterprise-k8s-docker
+
+enterprise:
+  enabled: true
+  license_secret: you-must-create-a-license-secret
```
hbagdi: When I meant that replace the secret names by the names we use in our guides, I meant it for all secrets. Please change this to how you changed the image pull secret.

charts/kong/example-values/README.md
```diff
@@ -0,0 +1,35 @@
...
+and disable legacy CRD templates (`ingressController.installCRDs: false`;
+you must change this value to `true` if you use Helm 2).
+
+* minimal-kong-controller.yaml installs Kong open source with the ingress
+  controller in DB-less mode.
+
+* minimal-kong-standalone.yaml installs Kong open source and Postgres with no
+  controller.
+
+* minimal-k4k8s-enterprise.yaml installs Kong for Kubernetes Enterprise with
```
hbagdi: These still don't have hyperlinks.

charts/kong/example-values/minimal-k4k8s-with-kong-enterprise.yaml
```diff
@@ -0,0 +1,54 @@
...
+    enabled: false
+
+env:
+  prefix: /kong_prefix/
+  database: postgres
+  password:
+    valueFrom:
+      secretKeyRef:
+        name: you-must-create-an-rbac-bootstrap-secret
+        key: you-must-replace-with-your-key
```
hbagdi: ditto for `name` and `key` here.

hbagdi: Assume that most folks will not even look inside this values.yaml file.

charts/kong/example-values/minimal-k4k8s-with-kong-enterprise.yaml
```diff
@@ -0,0 +1,54 @@
...
+admin:
+  enabled: true
+  http:
+    enabled: true
+    servicePort: 8001
+    containerPort: 8001
+
+enterprise:
+  enabled: true
+  license_secret: you-must-create-a-license-secret
```
hbagdi: ditto

charts/kong/example-values/minimal-k4k8s-with-kong-enterprise.yaml
```diff
@@ -0,0 +1,54 @@
...
+  pullSecrets:
+    # you-must add this: https://github.com/Kong/charts/blob/master/charts/kong/README.md#kong-enterprise-docker-registry-access
+    - kong-enterprise-edition-docker
+
+admin:
+  enabled: true
+  http:
+    enabled: true
+    servicePort: 8001
+    containerPort: 8001
```
hbagdi, line 22: 1. Can we default to exposing Admin API on 8444 with TLS?
2. Does this configuration actually work? The controller will dial on 8444 by default and fail here, no?

charts/kong/example-values/full-k4k8s-with-kong-enterprise.yaml
```diff
@@ -0,0 +1,206 @@
...
+    portal_emails_from: none@example.com
+    portal_emails_reply_to: none@example.com
+    admin_emails_from: none@example.com
+    admin_emails_reply_to: none@example.com
+    smtp_admin_emails: none@example.com
+    smtp_host: smtp.example.com
+    smtp_port: 587
+    smtp_auth_type: nil
+    smtp_ssl: nil
+    smtp_starttls: true
```
hbagdi, line 102: We should instead resort to SMTP mocking here for a simpler setup?

hbagdi (changes requested):

hbagdi:
Meta note: The documentation around Helm is growing and will grow further. We will need to move this to docs site in future for UX reasons (hello @lena-larionova).

lena-larionova:
> Meta note: The documentation around Helm is growing and will grow further. We will need to move this to docs site in future for UX reasons (hello @lena-larionova).

With you 100%. We've had some pushback when suggesting this recently, so I'm happy to see you say this.

hbagdi:
> We've had some pushback when suggesting this recently, so I'm happy to see you say this.

I've been the one who has been pushing back on this up to this point because we didn't have the docs site organized properly to add this somewhere. My understanding/opinion is we still don't have that (but are making **a lot of progress**, hence the comment).

lena-larionova:
> I've been the one who has been pushing back on this up to this point because we didn't have the docs site organized properly to add this somewhere. My understanding/opinion is we still don't have that (but are making **a lot of progress**, hence the comment).

Hah, I hadn't realized. If you have site suggestions that will help get it to the right state though, let me know! Actively working on it.

charts/kong/example-values/minimal-k4k8s-with-kong-enterprise.yaml
rainest: It works. This has used a template to choose the correct URL for a while now https://github.com/Kong/charts/blob/1.5.0/charts/kong/templates/_helpers.tpl#L121-L137

This example, by design, does very little beyond requiring the necessary secrets and exposing the admin API for use with Manager to economize on what you need to change. It's not intending to create a secure configuration; the full version is (though it's still limited in that respect because we don't support passing the certificates Kong serves as environment variables).

Using the self-signed certs is an option, but is likely to run into issues where requests from Manager fail unless you visit the admin API in your browser and click past its warning first.

charts/kong/example-values/full-k4k8s-with-kong-enterprise.yaml
rainest: It does; `enabled: false` ignores these and sets mocking on. The other settings are present here just to demonstrate that they exist and should be set if you wish to enable it.

charts/kong/example-values/minimal-k4k8s-with-kong-enterprise.yaml
rainest: You have to for the Enterprise examples: all of them require at least one setting that has to change. I can align the ones that exist, but the whole purpose of these is to provide a values.yaml for users to look at that's targeted at one particular deployment pattern, since the standard values.yaml covers settings for all options at once.

hbagdi: > all of them require at least one setting that has to change.

DB-less ones don't. We want to minimize the friction of standing up Enterprise as much as possible. 
What do you think about using the default values as the one listed in this guide:
https://github.com/Kong/kubernetes-ingress-controller/blob/master/docs/deployment/kong-enterprise.md#kong-enterprise-bootstrap-password

charts/kong/example-values/full-k4k8s-with-kong-enterprise.yaml
hbagdi: Got it. Thanks for the explanation.

charts/kong/values.yaml
```diff
@@ -505,7 +505,7 @@ enterprise:
   # Kong Enterprise license secret name
   # This secret must contain a single 'license' key, containing your base64-encoded license data
   # The license secret is required for all Kong Enterprise deployments
-  license_secret: you-must-create-a-kong-license-secret
+  license_secret: kong-enterprise-license
```
hbagdi, line 508: Are changes in this file intentional?

charts/kong/example-values/README.md
```diff
@@ -0,0 +1,35 @@
...
+  Manager, and does not expose it or the Admin API via a TLS-secured ingress.
+
+* [full-k4k8s-with-kong-enterprise.yaml](full-k4k8s-with-kong-enterprise.yaml)
+  installs Kong for Kubernetes with Kong Enterprise with the ingress controller
+  in PostgreSQL. It enables all Enterprise services.
+
+All Enterprise examples require some level of additional user configuration to
+install properly. Read the comments at the top of each file for instructions.
+
+Examples are designed for use with Helm 3, and disable Helm 2 CRD installation.
```
hbagdi, line 29: Can we mention somewhere in this file to refer to comments in respective values.yaml for more guidance?

rainest: We do: https://github.com/Kong/charts/pull/134/files#diff-2180e214f28f23d7e012fcbf288767e6R26-R27

charts/kong/values.yaml
rainest: Yes. If we're using that name in the guide and examples, I don't see any reason the stock values.yaml should be any different.

hbagdi (approved):

Issue #135, opened 2020-05-14
[kong] make it possible to only install CRDs in a Helm release


hbagdi:
If Kong is being installed multiple times in a single k8s cluster, it should be possible to install CRDs as a release and then install all Kong releases without CRDs.

Associated CRDs with any of the releases results in all other releases to be broken when the release that includes CRDs is removed.

rainest:
This is now effectively supported, but clunky: if you enable CRD installation and disable both the controller and Kong Pods, the chart does install the CRDs as a managed part of the release, and nothing else.

CRD management is inherently annoying because of the "delete CRD, delete all resources of that type footgun", which is kinda why Helm 3 takes a more hands-off approach (create CRDs if needed, but do not manage them along with the release) by default. That comes with its own problems related to CRD changes, but working within it is complicated so long as we support Helm 2 and Helm 3 both.

As-is, this is implemented, so closing this. We should revisit CRD management after removing Helm 2 support. #93 for that and related CRD discussion.

PR #136, opened 2020-05-14
[kong] add support for controller-only deployments

rainest:
Add a new `enabled` setting to values.yaml. When set to `true`, Kong and associated resources are rendered. When set to false, they are not.

This allows deploying the controller alone, which has been requested often.

CCed Satwant as he asked about this recently.

I'm not thrilled about the naming, and feel like we should eventually move the Kong-specific top-level configuration under a `kong` block. That's breaking, however, so we can't do it in the near future.

Thoughts on whether there's some better name we can use for this?

#### Checklist

### DIFF
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -310,7 +310,6 @@ directory.
 
 | Parameter                          | Description                                                                           | Default             |
 | ---------------------------------- | ------------------------------------------------------------------------------------- | ------------------- |
-| enabled                            | Enable or disable deploying Kong                                                      | `true`              |
 | image.repository                   | Kong image                                                                            | `kong`              |
 | image.tag                          | Kong image version                                                                    | `2.0`               |
 | image.pullPolicy                   | Image pull policy                                                                     | `IfNotPresent`      |
@@ -422,6 +421,7 @@ For a complete list of all configuration values you can set in the
 
 | Parameter                          | Description                                                                           | Default             |
 | ---------------------------------- | ------------------------------------------------------------------------------------- | ------------------- |
+| deployment.kong.enabled            | Enable or disable deploying Kong                                                      | `true`              |
 | autoscaling.enabled                | Set this to `true` to enable autoscaling                                              | `false`             |
 | autoscaling.minReplicas            | Set minimum number of replicas                                                        | `2`                 |
 | autoscaling.maxReplicas            | Set maximum number of replicas                                                        | `5`                 |
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -35,6 +35,8 @@ $ helm install kong/kong --generate-name --set ingressController.installCRDs=fal
   - [Configuration method](#configuration-method)
   - [Separate admin and proxy nodes](#separate-admin-and-proxy-nodes)
   - [Standalone controller nodes](#standalone-controller-nodes)
+  - [CRDs only](#crds-only)
+  - [Example configurations](#example-configurations)
 - [Configuration](#configuration)
   - [Kong Parameters](#kong-parameters)
     - [Kong Service Parameters](#kong-service-parameters)
@@ -267,9 +269,9 @@ helm install admin-only -f shared-values.yaml -f only-admin.yaml kong/kong
 ### Standalone controller nodes
 
 The chart can deploy releases that contain the controller only, with no Kong
-container, by setting `enabled: false` in values.yaml. There are several
-controller settings that must be populated manually in this scenario and
-several settings that are useful when using multiple controllers:
+container, by setting `deployment.kong.enabled: false` in values.yaml. There
+are several controller settings that must be populated manually in this
+scenario and several settings that are useful when using multiple controllers:
 
 * `ingressController.env.kong_admin_url` must be set to the Kong Admin API URL.
   If the Admin API is exposed by a service in the cluster, this should look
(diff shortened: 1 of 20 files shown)


satwant17 (commented):
~~we should highlight/comment  kong_url and publish_service properties in the value file like we do for the admin token~~
~~kong_url: http://kong-admin.kong.svc:80 **admin api url in the cluster** 
publish_service: kong/kong-proxy **proxy service **~~
Covered in the standalone readme section

satwant17 (approved):

hbagdi:
This seems like a hack and a bad one. Such global conditional flags usually cause more trouble in the long run and provide a terrible user-experience.
Considering #135 will have to implemented in future, we should stick to granular enabled settings.

Reading from the diff, I see:
- We need to tweak the deployment, this is a major one and we need to solve this
- Each Ingress needs to be enabled independently, it seems like this is already the case
- Each Service needs to be enabled independently, it seems like this is already the case

After reading through the values.yaml file, I figured out why you did this. It seems like everything specific to Kong is being governed at the root level and hence adding an `enabled` at the root level makes sense.

I don't think adding a root level `kong` is the ideal solution in the long run. It is bound to confuse more and doesn't have much gains.

What instead we can do is think of this change in terms of change in behavior of the single deployment resource. So what if we do: introduce a property to tweak deployment behavior: `deployment.kong.enabled`, this property only toggles the kong container off, and that's it. In future, the annotations related to the deployment resource can be moved under it. We could also allow users to not inject the wait for db init container if they explicitly want to disable it, add sidecar containers to this deployment and keep them off by default. I understand this is a change in how we usually think about "enabled". Thoughts?

rainest:
I agree that the lack of structure is bad, but it's the lack of structure we already have. There are a lot of things scattered around the root level that apply to the Deployment that could be better gathered together like so:

```
deployment:
  enabled: true
  annotations: {}
  replicas: 2
  pods:
    tolerations: []
    labels: {}
    annotations {}
    kong:
      enabled: true
    ingressController:
      enabled: true
```
However, I don't think we should move to that new style piecemeal. Controlling the Kong container via `.Values.deployment` and the controller via `ingressController.enabled` is confusing: like resources should have similar config.

I think having the standalone `deployment` block is good, since there are things attached to it that aren't at the container level, but still think we should have a `kong` section: there are currently several top-level settings that apply to the Kong container rather than the deployment (e.g. `readinessProbe`) and both Kong and the controller containers have enough settings that it makes sense to break them out rather than nest them under the `deployment` block. I'm ambivalent on whether to place their enable flag goes under their own block or under `deployment` as shown above.

This satisfies #135 by disabling both along with the existing "this isn't Helm 2" flag:

```
$ helm template --include-crds=true iggy --set ingressController.installCRDs=false --set enabled=false --set ingressController.enabled=false /tmp/symkong | grep kind:                                      
kind: CustomResourceDefinition
    kind: KongConsumer
kind: CustomResourceDefinition
    kind: KongCredential
kind: CustomResourceDefinition
    kind: KongPlugin
kind: CustomResourceDefinition
    kind: KongClusterPlugin
kind: CustomResourceDefinition
    kind: KongIngress
kind: CustomResourceDefinition
    kind: TCPIngress
        kind:
    kind: ""
```
This isn't perfectly exhaustive--there are things like pdb.yaml that I haven't explicitly conditioned on the Deployment, but it should catch all the defaults. We can make it explicit if desired.

That's in kind of a weird place as far as configuration. It uses the existing flag for Helm 2, even though it's nested under the controller settings and makes the if condition a mess (whatever, we probably won't need to touch it again and Helm 2 will eventually go away). For Helm 3 I don't think we can really add a dedicated "install CRDs" toggle--it doesn't appear that you can place template directives in CRDs at all. Something like this renders nothing:

```
{{- if .Values.enabled }}
# enabled
{{- else }}
# disabled
{{- end }}
```

So I don't think there's any way to handle CRD-only with the standard `crds/` handling outside having a combination of settings that disables all other templates. However, it looks like we will want to subchart them regardless:

> There is not support at this time for upgrading or deleting CRDs using Helm.

https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#some-caveats-and-explanations

I believe that adds new additions, but won't update, for instance, KongPlugin to include `configFrom`.

hbagdi:
>  I'm ambivalent on whether to place their enable flag goes under their own block or under deployment as shown above.

Although confusing, I'd recommend doing a `deployment.kong.enabled`. 


> So I don't think there's any way to handle CRD-only with the standard crds/ handling outside having a combination of settings that disables all other templates.

Let's do the disabling all the templates part. Any risks with that?

charts/kong/README.md
```diff
@@ -244,6 +245,28 @@ helm install proxy-only -f shared-values.yaml -f only-proxy.yaml kong/kong
 helm install admin-only -f shared-values.yaml -f only-admin.yaml kong/kong
 ```
 
+### Standalone controller nodes
+
+The chart can deploy releases that contain the controller only, with no Kong
+container, by setting `enabled: false` in values.yaml. There are several
```
hbagdi: This is out out of date with the new changes that were made in this PR.

charts/kong/templates/custom-resource-definitions.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if and .Values.ingressController.enabled .Values.ingressController.installCRDs -}}
+{{- if (or (and .Values.ingressController.enabled .Values.ingressController.installCRDs) (and (not (or .Values.deployment.kong.enabled .Values.ingressController.enabled )) .Values.ingressController.installCRDs)) -}}
```
hbagdi, line 6: Please add a comment in the code to explain this horrendous if condition and why we are doing that too. It will be very easy to forget the why part in future.

hbagdi: I'm myself having a hard time understanding this.

rainest:
FP for the release. b6723aa has changes for the comments.

hbagdi (approved):

PR #138, opened 2020-05-21
[kong] remove default securityContext

### LINKED ISSUE #137: Remove securityContext by default in the chart
Currently, Kong's default deployment contain a securityContext section:
https://github.com/Kong/charts/blob/master/charts/kong/values.yaml#L474

This was due to limitations in Kong which prevented to run it as non-root by default. This is not the case anymore so this should be removed to remove friction in deployment the controller on OpenShift platform.

see https://github.com/Kong/kubernetes-ingress-controller/issues/672

rainest:
Removes the default security context (which ran the container as UID 1000). This does not appear to be necessary and removing it improves compatibility with OpenShift.

  - fixes #137 

#### Checklist


hbagdi (approved):
Upgrade guide should include a note for this (not in scope of this PR).

rainest:
Shouldn't need it--there's no change in behavior with it gone, nor any harm leaving it in place if you already had it from an old values.yaml.

PR #145, opened 2020-05-27
[kong] release 1.6.0

rainest:
Releases version 1.6.0 of the Kong chart.

#### Checklist


hbagdi (approved):
Can we call out the removal of `uid` from securityContext in the upgrade doc?
Someone can run into a problem if they are using an older version of Kong.

Other than that, this looks good. Please feel free to merge after the above change. Thanks!

PR #156, opened 2020-06-12
[kong] Add support for SSL connections to Postgres

CristianPupazan:

Chart deployment fails when SSL is forced on Postgres. This PR is adding the ability to set the client certificate for Postgres SSL connections. 

  - fixes #5423

In order to test the change:

Create a `pg_secret.yml` file with the following content, adding your base-64 encoded certificate:

```
apiVersion: v1
kind: Secret
metadata:
  name: kong-pg-cert
data:
  pg-cert.pem: Q0xJRU5UIENFUlRJRklDQVRFIFRPIEJFIElOU1RFUlRFRCBIRVJF
```
Create the certificate:

`kubectl create -f pg_secret.yml`

Update `values.yml` accordingly:

```
pgSSL:
  enabled: true
  pgCertSecretName: "kong-pg-cert"
  pgCertSecretDataKeyName: "pg-cert.pem"
```

#### Checklist

### DIFF
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -172,6 +172,15 @@ the support has now been dropped due to stability issues.
 You can still deploy Cassandra on your own and configure Kong to use
 that via the `env.database` parameter.
 
+### Enabling SSL for Postgres connections
+
+For Postgres SSL connections you will need to provide the client certificate.
+This can be done via a secret that needs to be created in advance. More information about kubernetes secrets can be found
+[here](https://kubernetes.io/docs/concepts/configuration/secret/).
+
+Once the secret has been created, SSL can be enabled using the `pgSSL.enabled` parameter. The secret name and the key 
+name can be specified via the `pgCertSecretName` and `pgCertSecretDataKeyName` parameters. 
+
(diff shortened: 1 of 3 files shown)


rainest:
https://github.com/Kong/kong/issues/5423 is a limitation in the underlying libraries we use to handle TLS connections within Lua code. That's being addressed upstream, and we'll add configuration for Postgres client cert handling once it's available.

[lua_ssl_trusted_certificate](https://github.com/Kong/kong/blob/master/kong.conf.default#L1037-L1044) instead sets the location of the certificate bundle used to validate server certificates. That can be handled using https://github.com/Kong/charts/blob/kong-1.7.0/charts/kong/values.yaml#L231 (which mounts under `/etc/secrets/SECRET_NAME`) if needed, though often you'll only need the default CA bundle path provided by your distribution (we have an unrelated ticket open internally to set this by default in our image builds).

Closing as this can be handled using existing values.yaml parameters: it does not require a dedicated section and we generally try to avoid creating dedicated values.yaml sections unless it's not possible to implement that functionality otherwise.

PR #160, opened 2020-06-23
[kong] add support for hybrid mode

### LINKED ISSUE #142: Add support for hybrid mode of Kong

rainest:

Add basic support for hybrid mode:

* Add a new "cluster" service configuration section, default off.
* Add a new cluster service template.
* Add a new cluster ingress template.
* Update service settings docs in values.yaml to include cluster.
* Add Hybrid Mode section to the deployment guides.

Fix #142 

* No Enterprise examples or considerations yet, since Enterprise support for hybrid mode is still pending release. I expect configuration will largely be the same, so this should require new examples, but nothing else.
* No manual testing of the Ingress template, which is basically copy-paste from the manager/gui template minus the `enterprise.enabled` gate. I don't expect most users will use it (cross-cluster hybrid mode is theoretically doable, but complicated, and should probably be handled through some sort of VPN rather than Ingress), but it's there in case.
* Hybrid's use of `lua_ssl_trusted_certificate` presents a challenge for any user that needs to set that for other reasons. This is difficult to do without a significant revamp of how certificate trust is handled on the core end. See https://konghq.atlassian.net/browse/FTI-1618?focusedCommentId=29116 for the internal request.

#### Checklist


charts/kong/README.md
```diff
@@ -288,6 +293,124 @@ scenario and several settings that are useful when using multiple controllers:
...
+deployments](https://docs.konghq.com/2.0.x/hybrid-mode/) as of Kong 2.0.0 and
+Kong Enterprise 2.1.0. These deployments split Kong nodes into control plane
+(CP) nodes, which provide the admin API and interact with the database, and
+data plane (DP) nodes, which provide the proxy and receive configuration from
+control plane nodes.
+
+The chart does not support deploying control and data plane nodes in a single
+release. Users should [create separate releases for each node
+type](#separate-admin-and-proxy-nodes), i.e. use separate control and data
+plane values.yamls that are installed separately.
```
mflendrich: Suggested reword (to focus this section on "what hybrid mode is" and "how to achieve hybrid mode":

> [Hybrid mode](https://docs.konghq.com/2.0.x/hybrid-mode/) is a deployment pattern supported by Kong as of Kong 2.0.0 and Kong Enterprise 2.1.0. Deploying Kong in hybrid mode means splitting Kong nodes into **control plane (CP) nodes** (which provide the admin API and interact with the database), and **data plane (DP) nodes** (which provide the proxy and receive configuration from control plane nodes).
> 
> You can deploy Kong in hybrid mode by [creating separate Helm releases for each node type](#separate-kong-functions-across-nodes): one Helm release for control plane nodes, and one for data plane nodes.

charts/kong/README.md
```diff
@@ -288,6 +293,124 @@ scenario and several settings that are useful when using multiple controllers:
...
+
+* If `.Values.env.role="control_plane"`, the release will create a Service
+  named `RELEASE-NAME-kong-cluster`.
```
mflendrich: I'm unable to find a causal link between setting `.Values.env.role` to `control_plane` and the release creating such a service. How does it happen? Why is it important for the user?

Where is the `.Values.env.role` field used in the chart? I'm unable to find any such place 🤔

I suggest defining `.Values.env.role` in the chart-level `values.yaml` and documenting it there.

charts/kong/README.md
```diff
@@ -288,6 +293,124 @@ scenario and several settings that are useful when using multiple controllers:
...
+* Cluster certificates are not yet generated automatically. You must create a
+  certificate/key pair from a local instance, create a Secret containing them,
+  add the Secret to `.Values.secretVolumes`, and add
+  `cluster_cert`/`cluster_cert_key` values indicating its location under
+  `/etc/secrets` to `env`.
+* Note that data plane nodes do not default to using the default port (8005)
+  for `cluster_control_plane`. You should use `RELEASE-NAME-kong-cluster:8005`
+  as the `cluster_control_plane` value if using the default port.
```
mflendrich: What I think might be worth documenting is a closed and unambiguous list of steps that the user could blindly follow in order to deploy Kong in hybrid mode. I can see that all the steps (certificate generation, uploading secrets to k8s, updating values yamls, `helm install`) are loosely defined in this doc, but a simple list of steps is something I'd consider useful.

charts/kong/README.md
```diff
@@ -288,6 +293,124 @@ scenario and several settings that are useful when using multiple controllers:
...
+
+#### Certificates
+
+Hybrid mode uses TLS to secure the CP/DP node communication channel, and
+requires certificates for it. You can generate these either using `kong hybrid
+gen_cert` on a local Kong installation or using OpenSSL:
```
mflendrich, line 320: Suggested reword to avoid ambiguous pronouns and to specify the course of action:

> Hybrid mode secures the CP/DP node communication channel using TLS with client and server certificates as means of authentication. In order to deploy Kong in hybrid mode, you need to:
> 1. generate appropriate certificates,
> 1. wrap the newly generated certificates in Kubernetes secrets and apply them to the cluster, and
> 1. tell the Helm chart where the certificates are by specifying secret names in chart values.

Please verify the suggestion for factual correctness.

charts/kong/README.md
```diff
@@ -288,6 +293,124 @@ scenario and several settings that are useful when using multiple controllers:
...
+
+```
```
mflendrich: nit: be consistent about the use of `bash` in "```" blocks

charts/kong/README.md
```diff
@@ -288,6 +293,124 @@ scenario and several settings that are useful when using multiple controllers:
...
+the cert on the Ingress. This may be necessary if you run CP and DP nodes in
+separate clusters depending on how you expose the clusters' networks to one
+another.
+
+#### Data plane node configuration
+
+Data plane configuration also requires the certificate and `role`
+configuration, and the database should always be set to `off`. You must also
+trust the cluster certificate and indicate what hostname/port it should use to
+find control plane nodes.
```
mflendrich: misleading pronouns:

> **You** must also trust the client certificate

> indicate what hostname/port **it** should use

suggestion: replace those pronouns with "Kong"

charts/kong/README.md
```diff
@@ -343,13 +466,17 @@ individual services: see values.yaml for their individual default values.
...
 * `status`
 
 `status` is intended for internal use within the cluster. Unlike other
 services it cannot be exposed externally, and cannot create a Kubernetes
 service or ingress. It supports the settings under `SVC.http` and `SVC.tls`
 only.
 
+`cluster` is used on hybrid mode control plane nodes. It does not support the
+`SVC.http.*` settings, as cluster communications must be TLS-only.
+
```
mflendrich: suggestion: explain what kind of data the `cluster` service is used for. Imagine that a security engineer is reading this and is trying to tell whether the business (data plane) data may flow through this channel.

charts/kong/example-values/README.md
```diff
@@ -23,6 +23,14 @@ change this value to `true` if you use Helm 2).
   installs Kong for Kubernetes with Kong Enterprise with the ingress controller
   in PostgreSQL. It enables all Enterprise services.
 
+* [minimal-kong-hybrid-control.yaml](minimal-kong-hybrid-control.yaml) and 
+  [minimal-kong-hybrid-data.yaml](minimal-kong-hybrid-data.yaml) install
+  separate releases for hybrid mode control and data plane nodes, using
+  the built-in PostgreSQL chart on the control plane release. They require
+  some pre-work to create certificates and configure the control plane
+  location. See comments in the file headers for details. Note that you should
+  install the control plane release before the data plane release.
```
mflendrich: > They require **some** pre-work

(duplicate) suggestion: specify the pre-work as a list of steps to follow blindly

mflendrich: > Note that you should install the control plane release before the data plane release.

1. this is potentially problematic - not all ways to install Helm releases will guarantee that, especially if helm releases are managed by GitOps. For example, [helm-operator](https://github.com/fluxcd/helm-operator) won't allow you to install helm releases in a specific order.

2. consider making this requirement more visible if it's important for deployment success. Possibly as a clear warning in the "Hybrid mode" section.

charts/kong/values.yaml
```diff
@@ -128,6 +128,40 @@ status:
...
+  #  service.beta.kubernetes.io/aws-load-balancer-proxy-protocol: "*"
+
+  tls:
+    enabled: false
+    servicePort: 8005
+    containerPort: 8005
+    # Set a nodePort which is available if service type is NodePort
+    # nodePort: 32443
+
+  type: NodePort
```
mflendrich: Why `NodePort` and not `ClusterIP`?

mflendrich:
My overall feeling (not specific to this PR) about documentation in this repo is that it could have more of an explicit focus on:
- an end user not very familiar with Kong
- an end user not very familiar with Helm
- a user who's in a hurry and needs something to "just" work, not attempt to understand every single bit of configuration

charts/kong/README.md
```diff
@@ -1,4 +1,3 @@
-## Kong for Kubernetes
```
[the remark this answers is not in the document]
rainest: Unintended, this was me apparently hitting a bad sequence of keys in vim and then not spotting it in diff review. Thanks for catching it!

charts/kong/README.md
```diff
@@ -229,11 +229,16 @@ Kong can be configured via two methods:
...
+also split releases for hybrid mode CP/DP nodes, split proxy and Developer
+Portal nodes, etc.*
+
 Users may wish to split their Kong deployment into multiple instances that only
-run some of Kong's services, e.g. where some nodes only run the proxy and other
-only run the admin API, or where some nodes only run Developer Portal services.
-These require separate Helm releases (i.e. you run `helm install` once for
-every instance type you wish to create).
+run some of Kong's services (i.e. you run `helm install` once for every
+instance type you wish to create).
```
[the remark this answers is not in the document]
rainest: This is intentional since there's not really a great way to name this feature such that it's obvious why you'd use it. Separating admin and proxy is the most common based on anecdotal experience from community questions, so I used that as the title to grab those users' attention.

Historically the others (separating the Enterprise services) was something we'd also expect questions to come through customer support and sales engineering, who can direct those users to the right section, whereas the community users need to find it on their own. Hybrid doesn't match that, but it has its own dedicated section linking back here, so that should suffice for discoverability.

charts/kong/README.md
rainest: It used to, and I then removed it in favor of making the service and role stuff separate. That entails additional steps, but doesn't have any hidden functionality inconsistent with the other services' configuration, which I think is a reasonable tradeoff.

The original strategy made a bit more sense when I initially started work on this (a long long time ago--there was a significant pause due to priority shifts elsewhere), and the new strategy makes more sense in light of changes since to service handling.

Everything under `.Values.env` is translated into a `KONG_` environment variable, corresponding to [kong.conf](https://github.com/Kong/kong/blob/master/kong.conf.default) settings, so those generally aren't documented separately on their own (we do mention them in the context of specific configurations that need them, like here). There's a [rather large template helper](https://github.com/Kong/charts/blob/kong-1.7.0/charts/kong/templates/_helpers.tpl#L393-L581) that combines those values with others we set automatically from other sections of values.yaml. Those are then rendered in the container `env` sections of the deployment and migrations jobs, e.g. [for the main deployment](https://github.com/Kong/charts/blob/kong-1.7.0/charts/kong/templates/deployment.yaml#L75).

charts/kong/example-values/README.md
rainest: Useful to do so, but not required--it's mostly to make sure the DP nodes are able to become live and ready, since they need to initiate communication with a CP node to begin handling traffic. They do retry if that fails, so this is mostly guidance to bring them up more smoothly.

Adding clarifying notes that explain why you should do that.

charts/kong/README.md
rainest: Kong and friends are, for better or worse, fairly modular: you can and users do combine different features in surprising and unexpected ways, and there's not necessarily a single path that works for every environment and every configuration combination. This is indeed a source of user confusion, but one that's difficult to avoid given how flexible the product is, so documentation, especially higher-level README stuff, tends towards general rather than verbatim instructions, and assumes some level of familiarity to constrain length and avoid duplication (the latter isn't necessarily a problem on its own, but does turn into one when we need to update docs for new versions across the board--changes are fairly frequent and we do run into issues with doc rot).

The content under `example-values/` is our closest equivalent to this, since while we provide basic step-by-step instructions in the earlier sections of this README, once you get past basics (install Helm, add our repo, create license and Docker Secrets if using Enterprise) the configurations start to diverge quite a bit. Since values.yamls more or less encapsulates everything after, we provide those as templates with instructions for filling in the remaining required items--those were a specific request from sales engineering to fill in the "steps" between initial setup work and release installation.

charts/kong/README.md
rainest: Noted, though outside our area of concern--if the protocol details and contents of data transferred do come up in customer questions, it's something that should live on the main doc site. There's some overlap between content there and here, but content here should skew more towards chart-specific stuff. This is mostly to avoid doc rot: the main doc site gets more thoroughly reviewed during Kong releases and is better able to ensure it keeps up to date with changes as they arrive. That's harder for us since we don't track the Kong release versions: chart releases happen on their own schedule and apply to multiple Kong releases and variants, so the docs here are lighter on the specifics.

charts/kong/values.yaml
rainest: Pretty much arbitrary--we use it as the default elsewhere, though I don't think we have a solid reason to do so. That default can and probably should be re-evaluated, though it's out of scope for this.

charts/kong/README.md
rainest: Agree that the "it" in the first sentence is ambiguous, but I don't think we need to try and clarify further--whether the reference is back to "Hybrid mode", "communication channel", or "TLS" isn't crucial to understand (arguably, any interpretation is correct); the main point is that you need certs.

The rest I try to show through the exact commands and configuration used to reach the end state, i.e. if you use the `openssl` and `kubectl create secret` examples as shown, the values.yaml examples later will load the resulting certificate Secret.

charts/kong/templates/ingress-cluster.yaml
```diff
@@ -0,0 +1,34 @@
+{{- if .Values.deployment.kong.enabled }}
```
hbagdi: See comment in values.yaml; we should consider not introducing this right now.

charts/kong/templates/service-kong-cluster.yaml
```diff
@@ -0,0 +1,51 @@
...
+As such, it does not exist in the default values.yaml and is omitted from the
+initial render/do not render logic at the top, which is different from other
+Service templates.
+*/ -}}
+{{- if .Values.deployment.kong.enabled }}
+{{- if and .Values.cluster.enabled .Values.cluster.tls.enabled -}}
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ template "kong.fullname" . }}-cluster
```
hbagdi, line 14: `cluster` is a terrible name and it is unfortunate that Kong decided to go with it. It is far too generic to have any meaning. Can we rename this to be `control` or `control-plane` instead?

charts/kong/README.md
```diff
@@ -343,13 +461,17 @@ individual services: see values.yaml for their individual default values.
...
+* `cluster`
 * `status`
 
 `status` is intended for internal use within the cluster. Unlike other
 services it cannot be exposed externally, and cannot create a Kubernetes
 service or ingress. It supports the settings under `SVC.http` and `SVC.tls`
 only.
 
+`cluster` is used on hybrid mode control plane nodes. It does not support the
+`SVC.http.*` settings, as cluster communications must be TLS-only.
```
hbagdi: Consider using blockquote in markdown to make this stand out:
```
> `cluster` is used ...
```

Not a blocker.

charts/kong/README.md
```diff
@@ -288,6 +294,118 @@ scenario and several settings that are useful when using multiple controllers:
 Standalone controllers require a database-backed Kong instance, as DB-less mode
 requires that a single controller generate a complete Kong configuration.
 
+### Hybrid mode
```
hbagdi, line 297: Out of scope for this PR but we should consider moving such documentation to it's own markdown page. Just leaving a note. This is probably already on your mind.

charts/kong/README.md
```diff
@@ -288,6 +294,118 @@ scenario and several settings that are useful when using multiple controllers:
...
+```bash
+openssl req -new -x509 -nodes -newkey ec:<(openssl ecparam -name secp384r1) \
+  -keyout /tmp/cluster.key -out /tmp/cluster.crt \
+  -days 1095 -subj "/CN=kong_clustering"
+```
+
+You must then place these certificates in a Secret:
+
+```bash
+kubectl create secret tls kong-cluster-cert --cert=/tmp/cluster.crt --key=/tmp/cluster.key
```
hbagdi, line 331: We are using the same certificate and key pair for control and data nodes. Please consider generating a single CA certificate, two keys and then corresponding signed certificate.

hbagdi (changes requested):

charts/kong/values.yaml
mflendrich: If the `NodePort` is not used anywhere, then I strongly suggest changing this to `ClusterIP`:
* Unnecessary `NodePort`s quickly dry out a scarce resource (TCP port space) in larger clusters
* Security: a `NodePort` is an additional attack vector comparing to `ClusterIP`

[edit: wasn't aware of @hbagdi's comment at the moment of writing]

charts/kong/README.md
mflendrich: My main concern here is that this section leaves the user with an understanding that they need to generate and upload certificates, but (if the user doesn't have enough willpower to read the following section as well) there will be no mention of a need to point the helm release to the uploaded secret. This is not obvious for a reader IMO.

I suggest at the very least making a note at the end of this section (_Certificates_) to read that the user should also set the secret name as described in the two following sections (_Control plane node configuration_ and s/Control/Data/)

charts/kong/README.md
mflendrich: accidentally marked as resolved?

charts/kong/README.md
mflendrich: Thanks for the explanation. Created PR #163 to extend the comment for `.Values.env` with a hint about the `KONG_` translation.

charts/kong/templates/service-kong-cluster.yaml
rainest: IMO we cannot--I'm pretty strongly against tools on top of Kong using their own names for Kong concepts or overriding defaults, from the standpoint that inconsistency will be even more confusing. If it's a concern, it needs to be addressed with the core team, not here.

If we have that conversation with them, we should probably also bring up the odd default port, which follows the `800X` pattern for plaintext traffic even though it's TLS and would use `844X` following the convention set by the others.

charts/kong/README.md
mflendrich: the former case not updated - mistake?

I think that _"You"_ is not appropriate here because it's probably the server (and not the human reading this doc which is pointed at by "you") who needs to trust that client certificate.

charts/kong/README.md
rainest: This also falls into territory where I don't really agree with how this works in Kong itself, but don't feel the chart/chart docs is the place it should be addressed. This is designed to track the instructions in https://docs.konghq.com/2.0.x/hybrid-mode/#generating-certificatekey-pair, which produce a single cert/key pair and uses it for both client auth on DP nodes and server auth on CP nodes.

I do agree that we should create a CA and issue separate certs for each node, but that should be addressed in a future Kong release that the chart then follows--we shouldn't establish our own behavior independently.

charts/kong/example-values/README.md
```diff
@@ -23,6 +23,20 @@ change this value to `true` if you use Helm 2).
...
+  built-in PostgreSQL chart on the control plane release. They require some
+  pre-work to [create certificates](https://github.com/Kong/charts/blob/master/charts/kong/README.md#certificates)
+  and configure the control plane location. See comments in the file headers
+  for additional details.
+
+  Note that you should install the control plane release before the data plane
+  release if possible: data planes must be able to talk with a control plane
+  node before they can come online. Data plane nodes will retry this for a time
+  before evicted by liveness checks, however, so starting them second isn't
+  strictly required so long as the control nodes come online shortly after.
```
mflendrich: maybe possible to write this with fewer words?

```suggestion
  Note that data plane pods will crash-loop with liveness checks until they can 
  access the control plane. Therefore it's recommended (but not strictly necessary) 
  to start the control plane before the data plane.
```

charts/kong/README.md
rainest: Maybe--it's something I would defer absent evidence it's needed from the field. If we start getting a lot of questions about the details of the protocol, it's something we should add; if it doesn't come up often, then it probably isn't needed. That information may be of interest to a specialized audience, but someone just trying to get hybrid mode running shouldn't need to know detail beyond it using TLS and requiring TLS configuration.

Brevity is desirable, and I'll avoid including finer details if I think it's likely most readers would likely gloss over them anyway. My personal experience from the field (and in the comments here :smile:) is that complaints are more often directed towards a lack of simple instructions, where my response is "the simple instructions _are_ there, but they're alongside a lot of other detail, and I can understand why it's easy to miss them when looking at the large document". Because of that, if I think I can reasonably omit something and still guide readers to my goal (here, that they can spin up a hybrid mode cluster), I will.

mflendrich: > the simple instructions are there, but they're alongside a lot of other detail, and I can understand why it's easy to miss them when looking at the large document

Imagine a case where a Kubernetes user (new to Kong) comes to this repo to try out Kong in hybrid mode.
- Is it a likely scenario?
- Where are the "simple instructions" for this case? Most of my clarification requests originate from the fact that I'm consistently failing to locate those simple instructions, per my definition of simplicity (which can be roughly summarized by "contained or, at least, listed with links in one place")

Maybe I'm bikeshedding or missing the scope here and it's appropriate to curb this discussion. But my (newcomer's) perception is pretty clear that the documentation does not help to get up and running easily (except for a trivial case of "everything by default", but even in that case it doesn't apprently give any hint about how to make that "default" Kong do anything useful, like proxy to a vanilla echoserver present in the cluster).

In other words, one of the frameworks classifies documentation into four categories: how-tos, tutorials, references and explanations. This document serves as a good reference. I miss how-tos and tutorials for Kong Helm charts.

charts/kong/README.md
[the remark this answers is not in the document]
rainest: Like many of the comments elsewhere, I'd need to see evidence from elsewhere that it's a problem, or agree that the proposed change still meets the intent of the original also, and I don't here.

My expectation is that users will often read it starting from a specific question (here, "how do I separate my Kong admin and proxy into separate instances?"), and the section title needs to make it discoverable for them, rather than trying to achieve some ideal of title correctness. There's no official name for these as a group--we don't refer to "Kong functions" or "Kong services" much elsewhere, and instead refer to them individually by their own names, e.g. in the main doc site where they [are adjacent, but not under a unified header](https://i.imgur.com/V2z5Sr7.png). We know they're similar because we're familiar with the implementation, but can't expect that users necessarily do.

To try and spin an analogy, were this document a pharmacy, we (the pharmacy owners) likely would not have shelf sections labeled "μ-opioid receptor agonists" and "non-steroidal anti-inflammatory COX enzyme inhibitors" even though those are entirely accurate and technically specific descriptions--customers are much more likely to find what they're looking for if those sections are "gastric distress relief" and "pain relief", even though the "pain relief" section neither contains all known pain relievers, nor do its contents solely relieve pain. We also have staff (SEs and CSEs, and community forums) available to direct customers if they need something in those sections for a purpose other than its most common use--much like we can direct users looking to separate out Portal services or manage their CP/DP nodes, the pharmacy staff can direct customers to the NSAID aisle if they need something to deal with a fever.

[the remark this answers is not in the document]
mflendrich: Alright. I think we're hitting the "reference" vs "how-to" difference here, which I've explained in https://github.com/Kong/charts/pull/160#discussion_r445179942.

charts/kong/README.md
rainest: Yeah--the whole README is quite long at this point, and splitting some of the finer details into separate references is something we should do, just not now. Created #164.

charts/kong/README.md
rainest: That level of detail and the linear step order would be a good candidate for the guide section mentioned in #164. For this document I prefer to present the requirements briefly through examples that work together and build on it as we receive questions from users, rather than try to account for every possible hypothetical question in advance.

mflendrich: The problem my comment originated from is that I read this section as inaccurate. It is not sufficient to **generate and apply** certs in order to use the hybrid mode. It is required and sufficient to **pass existing cert secret names to the chart**. Right? Currently this section says the former and I believe it should say the latter.

Current wording feels misleading to me (it may leave the user with an assumption that the chart can „find” the cert secret somehow, for example if the name follows some convention.

charts/kong/README.md
rainest: > likely

Ideally no. The core product here is Kong, rather than the chart, and we sort of want users to come in through konghq.com and docs.konghq.com, since that's where all the flashy marketing content and basic introductions live.

The README.md here has some of that, because people do use it as a starting point in practice, but there are challenges to making it an effective entrypoint because there are so many branches: when you throw in "ah, but what if Enterprise?", "ah, but what if the controller is on?", "ah, but what if you use an HTTP load balancer versus a TCP load balancer?", etc.. There are a _lot_ of complicating factors that can and do come up, so when asked, "could we provide simpler instructions?" my response is often "yes, if you want to reduce the product feature set and allowed feature combinations" or "yes, if you want to focus on a single set of features in a fairly prescribed environment, which isn't the purpose of this document".

README.md is a general overview that seeks to cover many of these at a high level, so it can't also be specific. That's where (future) guides would come in, since they can more easily state up front that the guide specifically covers an Enterprise hybrid configuration without a controller, and then instructs users to create a TCP load balancer for the proxy.

Recognizing that it does have a dual purpose, there are simple instructions, but they're very basic, and specifically avoid those branching paths. That's why the [install instructions](https://github.com/Kong/charts/tree/master/charts/kong#install) and [Enterprise instructions](https://github.com/Kong/charts/tree/master/charts/kong#kong-enterprise) comprise only must-do, your-cluster-will-not-start-at-all-regardless-of-other-configuration-if-you-ignore-this steps, to present those absolute requirements with as little extra info as possible, since we want users (and particularly less savvy users) to be able to get to solid baseline before proceeding.

charts/kong/example-values/README.md
rainest: Made a different change to instead keep roughly the same length but remove as much K8S-specific terminology as possible--ideally users do know what liveness checks and backoff are, but many are less experienced and don't.

charts/kong/README.md
rainest: No action on this one, leaving as-is based on discussion elsewhere.

charts/kong/values.yaml
```diff
@@ -128,6 +128,40 @@ status:
...
+
+  # Kong cluster ingress settings.
+  # TODO: have left this in for now, but there is no corresponding template, and it doesn't work
+  # unsure if it makes sense to have a standard ingress setup, since it's of questionable use and
+  # the hybrid mTLS presumably cannot work through an ingress proxy without considerable work
+  ingress:
```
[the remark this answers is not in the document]
rainest: I initially didn't because I remembered the client TLS stuff, then forgot that and re-added it, and have now been reminded to remove it again. Explanatory docs added also.

charts/kong/README.md
rainest: That section changed a fair bit for other reasons, so if there are still changes needed, it should be in a new comment.

charts/kong/README.md
rainest: Usually if I only update something in part and resolve it means I only agreed with part--I won't always call out stuff I didn't do explicitly if I didn't think it was a big deal. This is keeping with https://github.com/Kong/docs.konghq.com/blob/master/STYLEGUIDE.md#personal-pronouns

This is more "you must do something" as a call to action; the longer implied form is "you must configure the data plane node to trust the certificate". The configuration example after shows exactly what to do in the actual config, so I'm less concerned with the exact specificity and absolute correctness of the presentation before.

rainest:
Some additional notes on the omnichange e74c11a5abe8ada54661a83c5134a44bfd19114a commit:

* I'd forgotten to actually set the cluster listen earlier :facepalm: This worked because Kong sets a default one if the role is set, but lacking it means you wouldn't be able to change the port.
* Features not supported by the cluster listen are now more documented in comments, though I still don't have configuration for them--I loathe to place stuff in values.yaml that says "this looks like valid YAML config, but it's only here to say that you can never change it. That causes some issues for the renderer, since it previously always expected that block, and none exists unless it's there in the default values.yaml or a user values.yaml, hence 94212d3184ce53507bbcdf1feaeac1a7ee4d82d2

For practical testing of footguns, I tried adding an `http` and `ingress` block. The latter just doesn't do anything, since its templates no longer exist. The former works normally, which is to say it doesn't, but Kong will open the listen without complaint. This is kinda good (configuring `http` doesn't break anything on its own, it's just pointless) and kinda bad (if you don't configure a TLS listen or try and point the DP to an HTTP listen, clustering will fail, with no indication other than the DP logs. That's kinda hard to really address with the split release paradigm--the DP only knows what it wants to talk to, and doesn't have any way to know in advance if that target is misconfigured.

hbagdi:
> The former works normally, which is to say it doesn't, but Kong will open the listen without complaint.

Please open a Github issue in `kong/kong` with reproducible steps for this one.

charts/kong/templates/_helpers.tpl
```diff
@@ -59,29 +59,37 @@ Generic tool for creating KONG_PROXY_LISTEN, KONG_ADMIN_LISTEN, etc.
...
-    {{- $listenConfig := merge $listenConfig .http -}}
-    {{- $_ := set $listenConfig "address" (default "0.0.0.0" .address) -}}
-    {{- $httpListen := (include "kong.singleListen" $listenConfig) -}}
-    {{- $unifiedListen = append $unifiedListen $httpListen -}}
+  {{/* Some services do not support these blocks at all, so these checks are a
+       two-stage "is it safe to evaluate this?" and then "should we evaluate
+       this?"
+  */}}
+  {{- if .http -}}
+    {{- if .http.enabled -}}
```
hbagdi, line 67: nit: make this simpler with `if and .http .http.enabled`

charts/kong/templates/_helpers.tpl
```diff
@@ -59,29 +59,37 @@ Generic tool for creating KONG_PROXY_LISTEN, KONG_ADMIN_LISTEN, etc.
...
-    */}}
-    {{- $listenConfig := dict -}}
-    {{- $listenConfig := merge $listenConfig .tls -}}
-    {{- $parameters := append .tls.parameters "ssl" -}}
-    {{- $_ := set $listenConfig "parameters" $parameters -}}
-    {{- $_ := set $listenConfig "address" (default "0.0.0.0" .address) -}}
-    {{- $tlsListen := (include "kong.singleListen" $listenConfig) -}}
-    {{- $unifiedListen = append $unifiedListen $tlsListen -}}
+  {{- if .tls -}}
+    {{- if .tls.enabled -}}
```
hbagdi, line 77: same nit here

charts/kong/values.yaml
```diff
@@ -127,7 +127,33 @@ status:
...
+# available on other services.
+#
+# The cluster service cannot be exposed through an Ingress, as it must perform
+# TLS client validation directly and is not compatible with TLS-terminating
+# proxies. If you need to expose it externally, you must use "type:
+# LoadBalancer" and use a TCP-only load balancer (check your Kubernetes
+# provider's documentation, as the configuration required for this varies).
+cluster:
+  # If you want to specify annotations for the cluster service, uncomment the following
+  # line, add additional or adjust as needed, and remove the curly braces after 'annotations:'.
```
hbagdi, line 146: Line 10 in the service template:
```
{{- if and .Values.cluster.enabled .Values.cluster.tls.enabled -}}
```

`cluster.enabled` is not defined here. That seems like a bug?

hbagdi (commented):
Nits and one possible bug.

We need to add CI test for this. I do understand that it is not possible with the current CI setup.
Please work with @jackkav who has worked on CI for other charts invovling multiple phases (Hint: use kind and a couple of shell scripts).
Figuring out a CI test plan is part of the scope of this PR and is a blocker to merge this PR in.
The CI test itself can be implemented after this PR is merged in (but before the GA release).

charts/kong/templates/_helpers.tpl
rainest: gotemplate attempts to check both always if they're within an `and` rather than failing fast if the first condition isn't met, so it null errors if this isn't nested.

rainest:
To confirm, the bug in question is the one on the Kong side regarding not really enforcing some of the hybrid mode requirements, or something else on our end?

I agree we should definitely add CI for this (as well as upgrades, which has been a backlog item for a while), though we may want to push that out to a minor release with manual testing for this one--mostly I want us to be able to get something available for hybrid mode alongside the 2.1 Kong releases.

Given that Kong 2.1 is soon and that the other stuff that we want for chart 2.x will be a considerable amount of work, I propose that we merge this without CI and roll a 1.8 chart release for it (without any of the 2.x breaking changes), and then have hybrid CI as a requirement for chart 2.x. @hbagdi @mflendrich does that sound like a reasonable plan?

hbagdi:
> To confirm, the bug in question is the one on the Kong side regarding not really enforcing some of the hybrid mode requirements, or something else on our end?

The bug is fixed with https://github.com/Kong/charts/pull/160/commits/6da1a9f067f685d5ea8660b5baddc10590a71045.

Regarding pushing the testing off, I'm usually not in favor of it because it takes some amount of discipline to actually get to it later on (something more important will quickly take priority).
On the other hand, I think it is important to release a chart that has hybrid mode compatibility around the same time that 2.1 comes out.
We can push testing off for this one but then lets keep the release small so that there are no other regressions involved.

hbagdi (approved):
Looks pretty good.
If there are issues, let's iterate. This is an exciting feature.

rainest:
Force push to resolve a minor conflict introduced by merger of #169, since both added new environment variables at the same place.

PR #163, opened 2020-06-24
[kong] values.yaml: document ".env" behavior

mflendrich:

materializes the knowledge from review comment https://github.com/Kong/charts/pull/160#discussion_r445056396

none

#### Checklist

### DIFF
--- a/charts/kong/values.yaml
+++ b/charts/kong/values.yaml
@@ -25,6 +25,9 @@ deployment:
 # -----------------------------------------------------------------------------
 
 # Specify Kong configurations
+# This chart takes all entries defined under `.env`, transforms them into into KONG_*
+# environment variables (interpreted by Kong as configuration value overrides) and
+# passes them to Kong containers.
 # Kong configurations guide https://docs.konghq.com/latest/configuration
 # Values here take precedence over values from other sections of values.yaml,
 # e.g. setting pg_user here will override the value normally set when postgresql.enabled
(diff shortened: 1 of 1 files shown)


charts/kong/values.yaml
```diff
@@ -25,6 +25,9 @@ deployment:
 # -----------------------------------------------------------------------------
 
 # Specify Kong configurations
+# This chart takes all entries defined under `.env`, transforms them into into KONG_*
+# environment variables (interpreted by Kong as configuration value overrides) and
+# passes them to Kong containers.
 # Kong configurations guide https://docs.konghq.com/latest/configuration
 # Values here take precedence over values from other sections of values.yaml,
 # e.g. setting pg_user here will override the value normally set when postgresql.enabled
```
```suggestion
# Specify Kong configuration
# This chart takes all entries defined under `.env` and transforms them into into `KONG_*`
# environment variables for Kong containers.
# Their names here should match the names used in https://github.com/Kong/kong/blob/master/kong.conf.default
# See https://docs.konghq.com/latest/configuration also for additional details
# Values here take precedence over values from other sections of values.yaml,
# e.g. setting pg_user here will override the value normally set when postgresql.enabled
```

rainest (changes requested):
Suggestion touches up some of the other stuff here and removes some of the finer details.

Users shouldn't really care that these take precedence over actual kong.conf values, since those are generally irrelevant for Helm installs (you'd have to modify templates to place a non-default kong.conf on the filesystem, which almost nobody will or should need to do in practice), so we can gloss over that part. 

Use of the kong.conf-style names is sort of a holdover from traditional deployments where kong.conf use is common. Our documentation usually references it/the names used in it rather than environment variables, and all documentation for those settings lives in kong.conf for the most part, hence why we focus on those even though ultimately we do translate them to environment variables for the actual deployment.

Issue #152, opened 2020-06-08
[kong] RBAC for single-namepsace deployments


rainest:
See the scenario described in https://discuss.konghq.com/t/clusterrole-and-clusterrolebinding-creation/6435/2?u=traines

Using `--watch-namespace` should allow for Kong deployments that do not create any ClusterRole for their ServiceAccount, instead placing all permissions into the namespaced Role. However, this is not tested and should be coordinated with the controller (not strictly required, but it should probably have its own Role-only example manifest that the chart tracks).

CRDs are always cluster-wide, but are easier to manage separately (as-is, Helm 3 *always* manages them separately, and only has the ability to create them alongside the release for convenience--the CRDs are not part of the release). The (Cluster)RoleBinding, by comparison, is tied to a specific ServiceAccount, so it needs to be part of the release for the template to bind it correctly.

Proposed changes:

- Add a new "single namespace" configuration property somewhere in values.yaml (probably top-level, since there are some cases where we use the ServiceAccount for non-controller configuration).
- If the new property is disabled, deploy RBAC resources as they are now.
- If the new property is enabled, do not create a ClusterRole or ClusterRoleBinding, and include all permissions that would normally exist in the ClusterRole in the Role, minus KongClusterPlugin (there may be other cluster resources that we'd normally have access to and can't use in a single namespace context).

This is probably contingent on https://github.com/Kong/kubernetes-ingress-controller/issues/717

andrevtg:
This would be awesome. I am currently able to use Kong in restrained environments (like Okteto) by patching both the chart and the ingress controller code (as mentioned in https://github.com/Kong/kubernetes-ingress-controller/issues/717).

rainest:
Handled via watchNamespaces with controller 2.0+

Issue #73, opened 2020-03-13
[kong] proposed breaking changes for v2.0.0


rainest:
This issue collects proposed breaking changes for the next major release of the Kong Helm chart. It will be updated over time based on the status of the proposals.

If you are a chart user and believe a proposed change would be incompatible with your environment, please let us know in the comments.

Checked items are code complete and expected to arrive in the new version. Unchecked items are as of yet undecided, and may or may not have some WIP code available.

# Active/open proposals
- [ ] Remove support for Helm 2. Maintaining backwards compatibility introduces additional complexity (e.g. #33) and limits from using new Helm 3 features. Migrating is fairly simple, and Helm 3 can coexist with Helm 2 deployments, so we do not believe this should be a major concern in most environments.
- [x] Remove support for multiple proxy ingress hostnames (https://github.com/Kong/charts/blob/kong-1.3.1/charts/kong/values.yaml#L111-L127). This allows unifying the Ingress templates. We believe that this feature is primarily of use in environments where Kong is used alongside a non-Kong ingress controller, and the proxy is not a LoadBalancer service. We have not encountered many such configurations in practice.
- [x] Remove support for unique admin API listener configuration. Support for the standard configuration format was added in #72. This allows using a single Service template for Kong services.
- [x] Remove support for dedicated Portal session and authentication sections of values.yaml. #55  

# Rejected proposals
- [ ] Remove support for existing dedicated `http`/`tls` configuration blocks (https://github.com/Kong/charts/blob/kong-1.3.1/charts/kong/values.yaml#L111-L127) for services in favor of a generic listener configuration that supports stream listens as well. << stream listens are distinct enough that it doesn't really make sense to consolidate configuration of them. We've consolidated most of the underlying template logic to simplify development, but there's no compelling reason to change the UX here also.

PR #166, opened 2020-06-29
clusterroles become roles for restricted clusters

andrevtg:

This is somewhat related to https://github.com/Kong/charts/issues/152

#### Checklist

This PR makes ClusterRoles changeable into Roles for single-namespace deployments is restricted clusters.

### DIFF
--- a/charts/kong/values.yaml
+++ b/charts/kong/values.yaml
@@ -302,7 +302,11 @@ ingressController:
   rbac:
     # Specifies whether RBAC resources should be created
     create: true
-    # Restricted clusters won't allow clusterroles
+    # Create cluster roles
+    # Cluster roles allow the controller to use resources from all namespaces for Kong configuration
+    # Depending on your K8S cluster configuration, you may not be able to create cluster roles, and can
+    # disable this setting if so. This limits the controller to resources in the namespace it's installed in,
+    # and should be used along with "ingressController.env.watch_namespace: INSTALL_NAMEPSACE"
     clusterroles:
       enabled: true
 
--- a/charts/kong/values.yaml
+++ b/charts/kong/values.yaml
@@ -302,6 +302,9 @@ ingressController:
   rbac:
     # Specifies whether RBAC resources should be created
     create: true
+    # Restricted clusters won't allow clusterroles
+    clusterroles:
+      enabled: true
 
   serviceAccount:
     # Specifies whether a ServiceAccount should be created
--- a/charts/kong/templates/controller-rbac-resources.yaml
+++ b/charts/kong/templates/controller-rbac-resources.yaml
@@ -59,11 +59,17 @@ subjects:
     namespace: {{ .Release.Namespace }}
 ---
 apiVersion: rbac.authorization.k8s.io/v1beta1
+{{- if .Values.ingressController.rbac.clusterroles.enabled }}
 kind: ClusterRole
 metadata:
+  name:  {{ template "kong.fullname" . }}
+{{- else }}
+kind: Role
+metadata:
+  name:  {{ template "kong.fullname" . }}-role
+{{- end }}
   labels:
     {{- include "kong.metaLabels" . | nindent 4 }}
-  name:  {{ template "kong.fullname" . }}
 rules:
   - apiGroups:
       - ""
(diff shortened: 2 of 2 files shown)


charts/kong/values.yaml
```diff
@@ -302,6 +302,9 @@ ingressController:
   rbac:
     # Specifies whether RBAC resources should be created
     create: true
+    # Restricted clusters won't allow clusterroles
+    clusterroles:
+      enabled: true
```
```suggestion
    # Create cluster roles
    # Cluster roles allow the controller to use resources from all namespaces for Kong configuration
    # Depending on your K8S cluster configuration, you may not be able to create cluster roles, and can
    # disable this setting if so. This limits the controller to resources in the namespace it's installed in,
    # and should be used along with "ingressController.env.watch_namespace: INSTALL_NAMEPSACE"
    clusterroles:
      enabled: true
```

rainest (changes requested):
Looks more or less fine as-is; have requested a change in the values.yaml doc. Please also add an entry to the README for the new setting: https://github.com/Kong/charts/blame/kong-1.7.0/charts/kong/README.md#L395-L416

@hbagdi this is kinda incongruous with the controller since the equivalent doesn't exist in the flat manifests. However, I don't know if we'd want to provide equivalents there or instruct users to change the CR/CRB manually if they need it, in which case the discrepancy is a non-issue. Do you think we should have a corresponding change there to add this somehow before this is released in Helm?

andrevtg (commented):
sounds much better this way, sorry for my lazy docs

charts/kong/values.yaml
```diff
@@ -302,6 +302,13 @@ ingressController:
   rbac:
     # Specifies whether RBAC resources should be created
     create: true
+    # Create cluster roles
+    # Cluster roles allow the controller to use resources from all namespaces for Kong configuration
+    # Depending on your K8S cluster configuration, you may not be able to create cluster roles, and can
+    # disable this setting if so. This limits the controller to resources in the namespace it's installed in,
+    # and should be used along with "ingressController.env.watch_namespace: INSTALL_NAMEPSACE"
+    clusterroles:
+      enabled: true
```
hbagdi, line 311: Can we instead opt for an enum?
```yaml
rbac:
  scope: "cluster" # allows values are "cluster" and "namespace"
```

The reasoning here is that although RBAC is not heavily configurable today, I expect users to be able to tweak RBAC resource heavily in future as the controller is going to get smarter about adjusting it's feature-set based on the RBAC permissions it has. Also, the controller is going to be compatible with multiple API groups:
- core k8s
- configuration.konghq.com
- Service APIs
- Knative
- Possibly some more

The scope of permissions and permissions themselves could be configurable and going down the `clusterroles.enabled` nesting path makes the values.yaml more verbose than it needs to be.

hbagdi:
> @hbagdi this is kinda incongruous with the controller since the equivalent doesn't exist in the flat manifests. However, I don't know if we'd want to provide equivalents there or instruct users to change the CR/CRB manually if they need it, in which case the discrepancy is a non-issue. Do you think we should have a corresponding change there to add this somehow before this is released in Helm?

It is completely fine (and even encouraged) for the Helm chart to have more features than the YAML manifests in KIC repo, but every feature in YAML manifest should be covered in the Helm chart.

The YAML manifests are meant as examples only.
(I know they are used as is in production but that's not what we should encourage)

charts/kong/values.yaml
rainest: Perhaps something like this, allowing us to add the others later?

```
# Settings related to Kubernetes RBAC
rbac:
  scope:
    # namespace-level roles and bindings. should always be enabled
    namespace: true
    # cluster-level roles and bindings. should be disabled in more restricted environments,
    # where Kong administrators do not have access to resources across the K8S cluster,
    # or do not want to grant K8S cluster-level permissions to the Kong cluster
    # when disabled, the chart creates equivalent namespaced roles as best possible
    cluster: true
```

I'd like to both clarify the Kong/K8S distinction (since both have RBAC) and also avoid using a single string to encapsulate all the toggles, since it seems like there may be many in the future. We wouldn't want something like `"clusteralsocontrolleralsoPSPalsoKnative"`.

Setting all to scopes false is equivalent to the currently proposed `rbac.create: false`. We should consider moving `ingressController.rbac.create` to an `rbac.controller` setting for chart 2.0, or otherwise rethink how we split K8S RBAC settings out from their related sections (`ingressController` and `podSecurityPolicy`) if at all.

Kinda torn on whether to allow users to toggle namespace-level roles off--I can't think of much reason to do so, but we kinda allow you to do so currently if I remember how the controller settings work

hbagdi: > I'd like to both clarify the Kong/K8S distinction (since both have RBAC) and also avoid using a single string to encapsulate all the toggles, since it seems like there may be many in the future. We wouldn't want something like "clusteralsocontrolleralsoPSPalsoKnative".

That was not my point. My point was rather to have `scope` as an enum of `cluster` and `namespace`. Fine-grained permissions or not part of `scope`. 

Whatever solution we decide should be forward compatible with the following features that _migth_ come in the ingress controller:
- Support for multiple namespaces (current it is either one or all namespaces)
- Support for fine-grained RBAC policies

> Setting all to scopes false is equivalent to the currently proposed rbac.create: false. We should consider moving ingressController.rbac.create to an rbac.controller setting for chart 2.0, or otherwise rethink how we split K8S RBAC settings out from their related sections (ingressController and podSecurityPolicy) if at all.

please track this in #73 

> Kinda torn on whether to allow users to toggle namespace-level roles off--I can't think of much reason to do so, but we kinda allow you to do so currently if I remember how the controller settings work

What do you mean by namespace-level role here? I'm not sure what you mean.

rainest: That sounds fine, `scope: "cluster"` then.

I just meant a plain Role, as opposed to cluster-scoped ClusterRole.

mkorbi:
hey @andrevtg  we are looking forward to seeing this commit, are you going to deliver the requested adjustment at the Readme soon? Thank you!

PR #175, opened 2020-07-24
[kong] add example Enterprise DB-less config

rainest:
#### What this PR does / why we need it:

Adds a beta example for K4K8S using the kong-enterprise-edition image.

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [x] PR is based off the current tip of the `next` branch and targets `next`, not `master`
- [x] New or modified sections of values.yaml are documented in the README.md
- [x] Title of the PR and commit headers start with chart name (e.g. `[kong]`)

PR #179, opened 2020-08-04
[kong] use bash wait-for-postgres implementation

rainest:
Replaces the existing busybox+netcat wait-for-postgres implementation with a centos+bash implementation. This allows for easier drop-in replacements of other waitImage options (namely the Red Hat UBI).

See discussion in https://github.com/Kong/kong-operator/issues/24 as to why this should be promoted from the `fork/redhat` branch into mainline releases.

The script is stored in a static ConfigMap template to work around some issues I encountered getting it to run correctly when placed directly in the Job definition as the initContainer command. This probably isn't a strict requirement, but it's a lightweight workaround and I'm not too concerned with getting rid of it absent some compelling reason to.

Example failure when DNS lookups fail:

```
waiting for db - trying ohno:5432
bash: ohno: Name or service not known
bash: /dev/tcp/ohno/5432: Invalid argument
waiting for db - trying ohno:5432
```
Example failure (or several checks before succeeding) when DNS succeeds:

```
waiting for db - trying example.com:5432
waiting for db - trying example.com:5432
waiting for db - trying example.com:5432
```
Although there's no direct relation between them, we may wish to merge and do a patch release of the example #175 before this. We can alternately add both and do a minor release, since the example can still be used with older versions of the chart.

#### Checklist

### DIFF
--- a/charts/kong/values.yaml
+++ b/charts/kong/values.yaml
@@ -398,8 +398,8 @@ postgresql:
 # -----------------------------------------------------------------------------
 
 waitImage:
-  repository: centos
-  tag: latest
+  repository: bash
+  tag: 5
   pullPolicy: IfNotPresent
 
 # update strategy
--- a/charts/kong/values.yaml
+++ b/charts/kong/values.yaml
@@ -398,7 +398,7 @@ postgresql:
 # -----------------------------------------------------------------------------
 
 waitImage:
-  repository: busybox
+  repository: centos
   tag: latest
   pullPolicy: IfNotPresent
 
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -456,8 +456,8 @@ directory.
 | migrations.preUpgrade              | Run "kong migrations up" jobs                                                         | `true`              |
 | migrations.postUpgrade             | Run "kong migrations finish" jobs                                                     | `true`              |
 | migrations.annotations             | Annotations for migration jobs                                                        | `{"sidecar.istio.io/inject": "false", "kuma.io/sidecar-injection": "disabled"}` |
-| waitImage.repository               | Image used to wait for database to become ready                                       | `busybox`           |
-| waitImage.tag                      | Tag for image used to wait for database to become ready                               | `latest`            |
+| waitImage.repository               | Image used to wait for database to become ready                                       | `bash`              |
+| waitImage.tag                      | Tag for image used to wait for database to become ready                               | `5`                 |
 | waitImage.pullPolicy               | Wait image pull policy                                                                | `IfNotPresent`      |
 | postgresql.enabled                 | Spin up a new postgres instance for Kong                                              | `false`             |
 | dblessConfig.configMap             | Name of an existing ConfigMap containing the `kong.yml` file. This must have the key `kong.yml`.| `` |
--- a/charts/kong/UPGRADE.md
+++ b/charts/kong/UPGRADE.md
@@ -66,8 +66,15 @@ script](https://github.com/Kong/charts/blob/kong-1.9.0/charts/kong/templates/wai
 to perform the same connectivity check. The default `waitImage.repository` value
 is now `bash` rather than `busybox`.
 
+The Helm upgrade cycle requires this script be available for upgrade jobs. On
+existing installations, you must first perform an initial `helm upgrade --set
+migrations.preUpgrade=false --migrations.postUpgrade=false` to chart 1.9.0.
+Perform this initial upgrade without making changes to your Kong image version:
+if you are upgrading Kong along with the chart, perform a separate upgrade
+after with the migration jobs re-enabled.
+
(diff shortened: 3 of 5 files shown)


charts/kong/values.yaml
```diff
@@ -398,7 +398,7 @@ postgresql:
 # -----------------------------------------------------------------------------
 
 waitImage:
-  repository: busybox
+  repository: centos
```
mflendrich: Consider using `bash:5` instead of `centos:latest` because the former is 13.7MB and the latter is 215MB.

If you choose to go with centos though, I strongly recommend `centos:8` instead of `centos:latest` because major RHEL/CentOS releases come with surprising breaking changes normally.

mflendrich (changes requested):
A suggestion regarding images used, and avoiding the `latest` tag. Otherwise OK :+1:

charts/kong/values.yaml
rainest: Sounds reasonable. I originally chose centos as probably the closest open-source public equivalent to the Red Hat UBI (which we use for the Red Hat certified stuff), but as we only care about bash and I don't expect differences in bash (fingers crossed on CI passing it), smaller image for that is good.

`latest` was a holdover from our busybox tag--it's probably more or less arbitrary since it seems unlikely that the bash component would change in a way that breaks this; I don't have particularly strong feelings either way for the basic utility image (if `latest` does end up breaking something for that, it's probably something we do need to investigate, because HOW!?), but as pinned is pretty much always the safer choice, seems reasonable to move to that.

Possible future security issues in an older image is the only concern, but bash vulnerabilities are the sort of thing that usually gets wide enough coverage that we hopefully shouldn't miss them.

hbagdi:
This change will have a heavy impact on existing users and can be considered even a breaking change, but this is where definitions of impact of any change gets blurry.

There are two roadblocks that users will run into:
1. Some users don't give their clusters access to any public docker registry for obvious security reasons. These users will have errors if they haven't explictly defined waitImage.
2. The other set of users who have defined waitImage will run into an issue where the bash command will fail, meaning the initContainer will fail.

The simplest solution would have been to use Kong's docker image if it shipped with bash and it's entrypoint could be easily overridden. I don't know if either of that is true or not.

Since I don't have a better solution either, let's move ahead but please add in a callout in the upgrade guide and the changelog.

rainest:
> Some users don't give their clusters access to any public docker registry for obvious security reasons. These users will have errors if they haven't explictly defined waitImage.

No change here really--busybox is also a public image, so while the specific image changes, the security concerns are more or less the same.

> The other set of users who have defined waitImage will run into an issue where the bash command will fail, meaning the initContainer will fail.

This could cause issues in the event that someone created a custom image and still expects to use the old netcat-based test, and furthermore doesn't have bash available on their image (or has it at a different location than the bash image). I don't think there's any great way to handle that beyond placing instructions in UPGRADE.md, so will add those.

The Kong images (at least centos and alpine) do have bash available, and I think the entrypoint script should handle them appropriately, though they could run into the same issue where the bash executable isn't at the same locatio. That's probably fine: it should be possible to let `$PATH` evaluation handle it and https://github.com/Kong/docker-kong/blob/master/centos/docker-entrypoint.sh (should be the same across our other platforms) does allow passing non-standard commands through the entrypoint--anything that's not a Kong command it just `exec`s verbatim. Maybe worth considering that, but as we haven't received user complaints about the existing busybox dependency (the RHM requirements to use Red Hat images only is the first instance I can recall where it was an issue), I'm skeptical that it's something we need.

As such, and given that the Kong images have the same issue for existing custom image+netcat users (they have bash, but not netcat), I think swapping bash for busybox as this has currently is fine.

hbagdi:
> No change here really--busybox is also a public image, so while the specific image changes, the security concerns are more or less the same.

I should have been more clear. Users run the public docker image of any referenced image and duplicate it in their private Docker registry after running through their security checks. With this change, users have an expectation that the upgrade will go smooth but it will fail if the `bash` docker image is not present in their private docker registry.

rainest:
> With this change, users have an expectation that the upgrade will go smooth but it will fail if the bash docker image is not present in their private docker registry.

Gotcha. The new UPGRADE.md content now covers this.

rainest:
Doc bug caught when upgrading existing: if you're upgrading from a version without this, the pre-upgrade job is created before the ConfigMap, and blocks on it. Not as ideal as a way around it without a ConfigMap (that was put in place to work around some weird issues I was having with writing it directly into the container `command`), but it's simple enough to avoid this by running a chart-only upgrade first with the jobs explicitly disabled.

Curious if there's maybe a way to skip the jobs automatically if there's no Kong version change, but probably not worth the effort since these issues are fairly infrequent and we can't necessarily rely on users doing independent chart/Kong version upgrades, even though we recommend it.

hbagdi (approved):
I'm further disappointed by Red Hat today.

PR #181, opened 2020-08-05
Updated values.yaml to include smtp_mock and smtp_auth_type

felderi:
Updated values.yaml to include smtp_mock and smtp_auth_type

  - fixes #

#### Checklist
- [ ] PR is based off the current tip of the `next` branch and targets `next`, not `master`

### DIFF
--- a/charts/kong/values.yaml
+++ b/charts/kong/values.yaml
@@ -578,6 +578,8 @@ enterprise:
     smtp_auth_type: nil
     smtp_ssl: nil
     smtp_starttls: true
+    smtp_mock: on
+    smtp_auth_type: nil
     auth:
       # If your SMTP server does not require authentication, this section can
       # be left as-is. If smtp_username is set to anything other than an empty


felderi:
Updated values.yaml to include smtp_mock and smtp_auth_type

hbagdi:
@felderi Please sign the CLA.

charts/kong/values.yaml
```diff
@@ -578,6 +578,8 @@ enterprise:
     smtp_auth_type: nil
     smtp_ssl: nil
     smtp_starttls: true
+    smtp_mock: on
+    smtp_auth_type: nil
```
hbagdi, line 582: This is already present on line 578. Please remove this.

charts/kong/values.yaml
```diff
@@ -578,6 +578,8 @@ enterprise:
     smtp_auth_type: nil
     smtp_ssl: nil
     smtp_starttls: true
+    smtp_mock: on
```
hbagdi, line 581: Please quote `on` (so make it "on"). YAML's boolean interpretation set is large and and this will be interpreted as a boolean `true` instead of a string `on`.
https://yaml.org/type/bool.html

hbagdi (changes requested):

rainest:
These sections exist as a compromise solution where we have Enterprise settings that are perhaps... overly aggressive in their enforcement of valid config. RBAC and SMTP settings have some strict checks where enabling/disabling some settings without doing the same for others results in Kong aborting startup. I'd somewhat like to do away with them entirely (everything they set could be set through `env` instead), but as-is they exist to avoid repeated redeploys while trying to recall the 4-5 settings you need to adjust in concert.

The downside is that they're sources of confusion, where it looks like you can/should add other related settings under them but cannot actually do so: their related templates require changes as well; it's not possible to add new settings in values.yaml alone.

My ideal solution would be to have graceful failures instead, i.e. where Kong will log a warning about config that's probably wrong or will interfere with non-critical functionality only, but that'd need to be handled through the Fast Track and Interfaces teams.

SMTP_MOCK is in fact also already present, it's just the `enterprise.smtp.enabled` key, which has the dual role of setting that and removing any of the other SMTP settings in mock mode: https://github.com/Kong/charts/blob/kong-1.8.0/charts/kong/templates/_helpers.tpl#L523-L542

Going ahead and closing this. @felderi if there was some issue that prompted it that we'd still need to address, let me know.

PR #185, opened 2020-08-26
[kong] add cluster telemetry service

### LINKED ISSUE #184: [kong] readme error
Hi, just a small one:

`SVC.ingress.hosts` (array) in helm values table should read `SVC.ingress.hostname` (singular), at least for the admin ingress.

rainest:
- Add support for the cluster telemetry service. This service is similar to the cluster service, but is used to ship Vitals data from Enterprise DP nodes.

  - fixes #184 (minor doc rollup fix unrelated to the main PR)
  - fixes an issue reported internally, where DP nodes with Vitals enabled would complain about cluster service failures. These are in fact cluster telemetry service failures, and do not interfere with the main cluster service. Inbound core changes will make those error reports less confusing, as they will indicate the specific failing endpoint.

- Also adds a cluster port to the deployment. It was previously omitted, and while omitting it doesn't break communication, it does misreport things (the port was open, but not declared).
- The Service name is too long for port names, so this uses a shortened `clustert-tls` name in the Deployment, whereas most stuff uses `clustertelemetry-tls`
- We don't currently have automated hybrid tests. At present, the `hybrid` namespace on our test cluster contains a pair of `silly-rabbit` releases that have Enterprise CP/DP nodes that share configuration and Vitals data. 

#### Checklist

### DIFF
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -514,7 +514,7 @@ authentication, which cannot pass through an ingress proxy).
 | SVC.externalTrafficPolicy          | k8s service's externalTrafficPolicy. Options: Cluster, Local                          |                     |
 | SVC.ingress.enabled                | Enable ingress resource creation (works with SVC.type=ClusterIP)                      | `false`             |
 | SVC.ingress.tls                    | Name of secret resource, containing TLS secret                                        |                     |
-| SVC.ingress.hosts                  | List of ingress hosts.                                                                | `[]`                |
+| SVC.ingress.hostname               | Ingress hostname                                                                      | `""`                |
 | SVC.ingress.path                   | Ingress path.                                                                         | `/`                 |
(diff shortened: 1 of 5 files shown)


charts/kong/README.md
```diff
@@ -488,7 +502,9 @@ only.
 `cluster` is used on hybrid mode control plane nodes. It does not support the
 `SVC.http.*` settings (cluster communications must be TLS-only) or the
 `SVC.ingress.*` settings (cluster communication requires TLS client
-authentication, which cannot pass through an ingress proxy).
+authentication, which cannot pass through an ingress proxy). `clustertelemetry
```
hbagdi:
```suggestion
authentication, which cannot pass through an ingress proxy). `clustertelemetry`
```

hbagdi (approved):
`cluster` is a horrible name but it is what it is in this case.

PR #190, opened 2020-08-31
adding Kong Enterprise values.yaml file to repo

marcoam:
It adds a _values.yaml_ file for Kong Enterprise. By having this file, it will simplify our instructions [here](https://docs.konghq.com/enterprise/2.1.x/kong-for-kubernetes/install-on-kubernetes/). This request came directly from our prospects/customers who mentioned that the current process we have (modifying a KongCE-values.yaml to make it EE-compatible) is not efficient and worst of all, error prone.

### DIFF
--- a/charts/kong/kong-enterprise-values.yaml
+++ b/charts/kong/kong-enterprise-values.yaml
@@ -1,0 +1,737 @@
+# Default values for Kong Enterprise's Helm Chart
+# Declare variables to be passed into your templates.
+#
+# Sections:
+# - Deployment parameters
+# - Kong parameters
+# - Ingress Controller parameters
+# - Postgres sub-chart parameters
+# - Miscellaneous parameters
+# - Kong Enterprise parameters
+
+# -----------------------------------------------------------------------------
+# Deployment parameters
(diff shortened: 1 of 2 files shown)


rainest:
We provide these under https://github.com/Kong/charts/tree/main/charts/kong/example-values as (a) there are a variety of Enterprise (and community) deployment patterns, so a one-size-fits-all approach to providing a values.yaml isn't feasible (we do provide a single main values.yaml, but because it's necessary in order to provide default values at all), and (b) focus only on the values.yaml settings they need, rather than the whole set (most are Enterprise/community-agnostic).

Internally, DOCS-897 has background on the current state of this, which is a bit in flux because of 2.1 changes, but does definitely still need links in the docs site. Will close this in favor of that, though there are some additional questions in that Jira re whether our existing examples are what we need and how much/if they need to change in light of product changes elsewhere.

https://github.com/Kong/charts/pull/191 has minor changes to the version; we missed the version bumps in the older examples initially, so just be aware that those still refer to 1.5 for the time being (will be fixed soon).

PR #193, opened 2020-09-02
Add namespace to generated manifests

### LINKED ISSUE #192: Is there a reason why some templates does not contain namespace?
Hello,

is there a reason why some of the templates (e.g. `service-kong-admin`, `service-kong-proxy`) do not contain namespace (I would expect something like `namespace: {{ .Release.Namespace }}` in template) whereas others (`controller-service-account`) do? 

When using `helm template` with this chart to generate manifests, namespaces are not present which complicates further usage. It would be nice to have them for all the templates (I can prepare a PR if needed).

Regards,
Martin.

mdostal-hci:
It explicitely adds namespace to generated manifests (e.g. by `helm template`), not to rely on automated Helm inclusion. See https://github.com/Kong/charts/issues/192 for more detail.

fixes #192 

#### Checklist

### DIFF
--- a/charts/kong/templates/admission-webhook.yaml
+++ b/charts/kong/templates/admission-webhook.yaml
@@ -10,6 +10,7 @@ apiVersion: admissionregistration.k8s.io/v1beta1
 {{- end }}
 metadata:
   name: {{ template "kong.fullname" . }}-validations
+  namespace: {{ .Release.Namespace }}
   labels:
     {{- include "kong.metaLabels" . | nindent 4 }}
 webhooks:
@@ -38,6 +39,7 @@ apiVersion: v1
 kind: Service
 metadata:
   name: {{ template "kong.service.validationWebhook" . }}
+  namespace: {{ .Release.Namespace }}
   labels:
     {{- include "kong.metaLabels" . | nindent 4 }}
 spec:
@@ -54,6 +56,7 @@ apiVersion: v1
 kind: Secret
 metadata:
   name: {{ template "kong.fullname" . }}-validation-webhook-keypair
+  namespace:  {{ .Release.Namespace }}
   labels:
     {{- include "kong.metaLabels" . | nindent 4 }}
 type: kubernetes.io/tls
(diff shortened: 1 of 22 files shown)


rainest:
Hey, more places to tag @eskibars !

This hits a vein similar to the end of https://github.com/Kong/kubernetes-ingress-controller/pull/844#pullrequestreview-490946197, where we have most of our more complex tooling in this Helm chart, but have some users that don't want all of Helm's overhead.

Chart modifications to support Helm use via `helm template` (which only renders manifests from templates, and does not handle lifecycle management) may be an imperfect middle ground: `helm template` allows for workflows that are similar to what Kustomize provides, while leveraging our existing templates for more advanced functionality. We have have encountered users who ran into issues with this approach in the past, as using `helm template` to render templates and then apply them outside Helm's lifecycle management isn't really supported as a first-class/expected use case by Helm itself (I may be wrong there--checking with their devs to see how much that is or isn't the case), so we may need to proceed with caution and/or better document gotchas if we go the `helm template` route.

I've historically been reticent to add documentation or template functionality targeted at `helm template` use as such, though in this particular case the change isn't something that would cause obvious problems when using `helm install` and doesn't require much maintenance going forward, so I'd be pretty okay merging this if we increase endorsement of `helm template` use while acknowledging some of the rough edges.

@mdostal-hci though this doesn't relate to the PR content itself, you're a demonstrably good person to ask know that we know you do use the chart that way: what other stumbling blocks have you encountered, if any? In particular, how do you handle migration jobs? Do you find that you need to do much beyond provide `--is-upgrade` and extract them from the larger body of rendered templates?

Side note: approving but not merging yet--mostly just keeping this open because I'm interested in the broader discussion between `helm template` users, myself, and members of our product strategy team. We don't have an imminent chart release; this should go in for the next one, however.

rainest (approved):

hbagdi (approved):
As Travis said, we don't have a lot of feedback from users who do the `helm template` workflow. We are happy to incorporate such improvements and even document it further if that helps the community.

rainest:
Meh, go figure, other PR created conflicts. Merged manually in 700d2879d7a0f7f9e6a6138788a439b3292216e1 to avoid anything similar going forward. We'll keep the stuff for product management in mind, as we have other open tickets for that elsewhere.

PR #204, opened 2020-09-17
[kong] release 1.10.0 and reintegrate 1.9.0 commits

mflendrich:
This PR releases the Kong chart version 1.10.0 and reintegrates `next` into `main`.

Reintegration of `next` into `main` is necessary because during the 1.9.0 release some commits were copied over to `main` from `next` by cherry-picking/rebasing, and therefore commits from `next` are not integrated into `main`. After this PR, all commits from `next` become ancestors of `main`.

The reintegration is a no-op from the perspective of file content - it only affects the relationships between commits in the commit tree.


rainest (approved):
IIRC this was additional weirdness stemming from either the master/main switch or unrelated issues with main-only CI jobs. Hopefully shouldn't crop again on future releases.

rainest:
Ah, well, spoke too soon--thought this looked good, but it looks like GitHub is indicating that no rebase+merge is allowed now that it's past the approval requirement. For obvious reasons we don't want squash here--not sure if we can instead merge commit safely and proceed with updating next to include it after. My git brain is not bigbrain enough to know the best approach here.

mflendrich:
Rebasing always mean creating copies of commits. We don't want that here - instead, we do want the commits from `next` to become ancestors of `main`. This is only possible with either a merge commit, or a `--ff` merge.

Did the former - created a merge commit.

PR #208, opened 2020-10-05
[kong] Add custom extra configmaps/secrets

### LINKED ISSUE #196: How can i mount extra volume from host to Kong Pod using this helm chart ?
We want to put logs generated by Kong pod into one volume (external storage mounted on host).  I understand best practice is to have log on stdout and stderr then ship it to some external logs tool like elastic but for now our appliacation needs log at mounted storage. any help for this issue appriciated

mikemenaker:
Add extraConfigMaps and extraSecrets fields to values file that will be added to volumes/volumeMounts template helpers.

Adds ability to mount arbitrary config maps/secrets

closes #196

#### Checklist
- [ X ] PR is based off the current tip of the `next` branch and targets `next`, not `main`
- [ X ] New or modified sections of values.yaml are documented in the README.md
- [ X ] Title of the PR and commit headers start with chart name (e.g. `[kong]`)

### DIFF
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -609,6 +609,9 @@ For a complete list of all configuration values you can set in the
 | serviceMonitor.namespace           | Where to create ServiceMonitor                                                        |                     |
 | serviceMonitor.labels              | ServiceMonitor labels                                                                 | `{}`                |
 | serviceMonitor.targetLabels        | ServiceMonitor targetLabels                                                           | `{}`                |
+| extraConfigMaps                    | ConfigMaps to add to mounted volumes                                                  | `[]`                |
+| extraSecrets                       | Secrets to add to mounted volumes                                                     | `[]`                |
+
 
 #### The `env` section
 
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -350,10 +350,6 @@ The name of the service used for the ingress controller's validation webhook
   {{- if .subPath }}
   subPath: {{ .subPath }}
   {{- end }}
-
-  {{- if .readOnly }}
-  readOnly: {{ .readOnly }}
-  {{- end }}
 {{- end }}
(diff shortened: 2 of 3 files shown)


rainest (changes requested):
Woo, finally got a chance to play around with this after Kong Summit stuff. We had some history issues with our release process as part of that, so you'll want to merge in the latest `next` tip.

Generally looks good as far as mounting stuff, but I have one change request and one question that's probably also a change request. Simpler one first!

For the comments, I'd suggest making the inline comments a double-comment with some brief explanation, so that removing the leading `#` doesn't include anything that will parse as YAML other than the the example string itself, e.g. 


```
#   subPath: my-subpath # Optional, if you wish to mount a single key and not the entire ConfigMap
```
instead of:

```
#   subPath: my-subpath [OPTIONAL]
```
More complex one second: do you know of any reason to include the `readOnly` field? That looks like maybe some confusion introduced by the K8S docs and volume mount spec, which address all types of volume mounts equally. As best I can tell, R/W ConfigMap or Secret mounts aren't actually allowed, so I'd say remove that setting (and the associated template content) unless you know of some edge case that'd require it still.

ConfigMap/Secret mounts appear to be [read-only always](https://github.com/kubernetes/kubernetes/issues/62099#issuecomment-527286207), and K8S will just ignore silently ignore that setting for them. Per the other issue comments, this behavior is itself a source of confusion, but apparently exists as a legacy holdover from when K8S _did_ allow you to modify mounted ConfigMaps and Secrets.

mikemenaker:
I updated the fork with the requested changes, I'll merge/rebase the last 'next' branch tomorrow morning.

rainest (approved):

rainest:
Bizarre. CI keeps failing because a Postgres-backed instance cannot connect to Postgres, but I cannot replicate this locally with the same test values on a fresh install/namespace.

Possibly something awry with GH actions at the moment; will try tests again on Monday.

PR #209, opened 2020-10-06
[kong] release 1.11.0 to next

rainest:
- Updates the chart to 1.11.0.
- Bumps controller version to 1.0.
- Updates CRDs to controller 1.0 version.
- Adds changelog and upgrade documentation.
- Fixes formatting and ToC issues from 1.10.0.
- Integrates commits from main that were not already in next.

For 1.10, we merged main into next and then added several 1.10 changes directly to main, and the tip of next was behind main.

For this PR, I've merged main into the release branch, which incorporates those changes. Using a rebase and merge for this merge into next and the subsequent merger of next into main should bring the branches back into alignment.

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -1,10 +1,43 @@
 # Changelog
 
+## 1.11.0
+
+### Breaking changes
+
+* Kong Ingress Controller 1.0 removes support for several deprecated flags and
+  the KongCredential custom resource. Please see the [controller changelog](https://github.com/Kong/kubernetes-ingress-controller/blob/main/CHANGELOG.md#breaking-changes)
+  for details. Note that Helm 3 will not remove the KongCredential CRD by
+  default: you should delete it manually after converting KongCredentials to
+  [credential Secrets](https://github.com/Kong/kubernetes-ingress-controller/blob/next/docs/guides/using-consumer-credential-resource.md#provision-a-consumer).
+  If you manage CRDs using Helm (check to see if your KongCredential CRD has a
+  `app.kubernetes.io/managed-by: Helm` label), perform the credential Secret
+  conversion **before** upgrading to chart 1.11.0 to avoid losing credential
+  configuration.
+* The chart no longer uses the `extensions` API for PodSecurityPolicy, and now
+  uses the modern `policy` API. This breaks compatibility with Kubernetes
+  versions 1.11 and older.
+  ([#195](https://github.com/Kong/charts/pull/195))
+
+### Improvements
+
+* Updated default controller version to 1.0.
+* The chart now adds namespace information to manifests explicitly. This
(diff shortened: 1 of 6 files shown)


charts/kong/crds/custom-resource-definitions.yaml
```diff
@@ -1,71 +1,3 @@
...
-kind: CustomResourceDefinition
-metadata:
-  name: kongcredentials.configuration.konghq.com
-spec:
-  group: configuration.konghq.com
-  version: v1
-  scope: Namespaced
-  names:
-    kind: KongCredential
-    plural: kongcredentials
```
hbagdi, line 45: My initial instinct here is to not remove this because it means users can't use this resource even for older versions of the controller.
Let's remove it for now and we can add it back in if such a request comes up. I doubt it will come up.

charts/kong/CHANGELOG.md
```diff
@@ -1,10 +1,43 @@
...
+  conversion **before** upgrading to chart 1.11.0 to avoid losing credential
+  configuration.
+* The chart no longer uses the `extensions` API for PodSecurityPolicy, and now
+  uses the modern `policy` API. This breaks compatibility with Kubernetes
+  versions 1.11 and older.
+  ([#195](https://github.com/Kong/charts/pull/195))
+
+### Improvements
+
+* Updated default controller version to 0.10.0.
```
hbagdi: Please remove this line.

hbagdi (changes requested):

charts/kong/crds/custom-resource-definitions.yaml
rainest: Eh, it's kind of in a weird spot. In modern Helm, using the default crds directory behavior, removing it does nothing for existing releases. It's only actually removed when you follow the UPGRADE.md instructions to destroy it.

Old behavior and/or intentionally managing CRDs with Helm 3 (`installCRDs: true`) does destroy it, which is why I tried to write instructions that stress that and provide several ways to check. The riskiness of that management method is, AFAIK, why Helm 3 switched to the new method. It has its own drawbacks, but it's much safer.

That does remind me that I need to instruct people to apply after though--they won't get the status fields otherwise.

charts/kong/CHANGELOG.md
rainest: It remains, but it now uses the correct version for this release.

rainest:
Along with the new fix commit, CHANGELOG.md now indicates the correct default controller version for this release and UPGRADE.md now includes https://github.com/Kong/charts/pull/209/files#diff-58a65045f2365289ab5a8958aa3aebe4R92-R102

That apply is safe even without doing the KongCredential work (though you should still use the KongCredential work), as applying a file without KongCredential just won't touch it/will only add status to the still-supported CRDs.

hbagdi (approved):

PR #219, opened 2020-10-29
Add annotations for initmigrations job

seanson:
This enables users to specify optional specific annotations for the init migrations job. This allows more control when deploying Kong via systems other than standard Helm installs, like ArgoCD. The use case for us is to allow setting some custom annotations that allow us to clean up this job after running.

I've additionally clarified the documentation for existing `.migrations.annotations` to state that they modify the job's pod annotations and not the job itself.

#### Checklist

### DIFF
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -469,7 +469,7 @@ directory.
 | migrations.preUpgrade              | Run "kong migrations up" jobs                                                         | `true`              |
 | migrations.postUpgrade             | Run "kong migrations finish" jobs                                                     | `true`              |
 | migrations.annotations             | Annotations for migration job pods                                                    | `{"sidecar.istio.io/inject": "false", "kuma.io/sidecar-injection": "disabled"}` |
-| migrations.initAnnotations         | Annotations for init migration job                                                    | `{}`                |
+| migrations.jobAnnotations          | Additional annotations for migration jobs                                             | `{}`                |
 | waitImage.repository               | Image used to wait for database to become ready                                       | `bash`              |
 | waitImage.tag                      | Tag for image used to wait for database to become ready                               | `5`                 |
 | waitImage.pullPolicy               | Wait image pull policy                                                                | `IfNotPresent`      |
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -468,7 +468,8 @@ directory.
 | env                                | Additional [Kong configurations](https://getkong.org/docs/latest/configuration/)      |                     |
 | migrations.preUpgrade              | Run "kong migrations up" jobs                                                         | `true`              |
 | migrations.postUpgrade             | Run "kong migrations finish" jobs                                                     | `true`              |
-| migrations.annotations             | Annotations for migration jobs                                                        | `{"sidecar.istio.io/inject": "false", "kuma.io/sidecar-injection": "disabled"}` |
+| migrations.annotations             | Annotations for migration job pods                                                    | `{"sidecar.istio.io/inject": "false", "kuma.io/sidecar-injection": "disabled"}` |
+| migrations.initAnnotations         | Annotations for init migration job                                                    | `{}`                |
 | waitImage.repository               | Image used to wait for database to become ready                                       | `bash`              |
 | waitImage.tag                      | Tag for image used to wait for database to become ready                               | `5`                 |
 | waitImage.pullPolicy               | Wait image pull policy                                                                | `IfNotPresent`      |
--- a/charts/kong/templates/migrations-post-upgrade.yaml
+++ b/charts/kong/templates/migrations-post-upgrade.yaml
@@ -13,6 +13,9 @@ metadata:
   annotations:
     helm.sh/hook: "post-upgrade"
     helm.sh/hook-delete-policy: "before-hook-creation"
+  {{- range $key, $value := .Values.migrations.jobAnnotations }}
+    {{ $key }}: {{ $value | quote }}
+  {{- end }}
 spec:
   template:
     metadata:
--- a/charts/kong/templates/migrations-pre-upgrade.yaml
+++ b/charts/kong/templates/migrations-pre-upgrade.yaml
@@ -13,6 +13,9 @@ metadata:
   annotations:
     helm.sh/hook: "pre-upgrade"
     helm.sh/hook-delete-policy: "before-hook-creation"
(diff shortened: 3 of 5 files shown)


rainest (changes requested):
tl;dr You'll want to follow the pattern used for `migrations.annotations`: the new value for Job annotations should apply to all migration jobs, i.e. in the [pre-upgrade](https://github.com/Kong/charts/blob/main/charts/kong/templates/migrations-pre-upgrade.yaml) and [post-upgrade](https://github.com/Kong/charts/blob/main/charts/kong/templates/migrations-post-upgrade.yaml) templates also. Please name the new value `migrations.jobAnnotations`.

There shouldn't be a reason to have separate annotations sets for each type of migration Job, as they usually have similar needs (e.g. you want to avoid sidecar injection on _all_ migration Pods, or want to clean up _all_ migrations Jobs, not just init or pre-upgrade migrations) and, worst case, don't run at the same times anyway.

We'd ideally use the same naming conventions as we use for the main deployment, which has separate values for `deploymentAnnotations` (which adds annotations to the Deployment itself) and `podAnnotations` (which adds annotations to the Deployment's Pod template, same as the existing `migrations.annotations` value). However, renaming `migrations.annotations` (to have `migrations.jobAnnotations` and `migrations.podAnnotations`) would be a breaking change and that it's easy enough to use `migrations.jobAnnotations` and add a comment indicating that `migrations.annotations` adds annotations to the Job's Pods (as you've done) for now.

This particular use case (deleting old Jobs without a cron job or similar somewhere) has come up before, and we've historically said that Helm's native functionality (deleting them through lifecycle management) is fine. We'd wanted to use native K8S functionality for this for both workflows that use Helm lifecycle management and those that do not, as [it does exist](https://kubernetes.io/docs/concepts/workloads/controllers/job/#ttl-mechanism-for-finished-jobs), but that feature is de facto unavailable because it's behind an alpha gate, and alpha features aren't readily available for a significant fraction of users because managed platforms don't necessarily allow them.

I cannot find much information on promotion of that TTL mechanism to beta or GA, unfortunately, and it doesn't appear to have moved much since I initially found it. The original [graduation criteria](https://github.com/kubernetes/enhancements/blob/master/keps/sig-apps/0026-ttl-after-finish.md#graduation-criteria) don't indicate much beyond SIG Applications wanting to gather user feedback, but it's unclear whether they have: searching for issues about it mostly brings up user reports indicating difficulty enabling the alpha flag, and searching the community Slack chat history mostly turns up other users asking whether there were plans to promote it past alpha without responses, so my best assessment at this time is that development of the feature and user feedback gathering are stalled, and it likely will not be available beyond alpha for the forseeable future.

seanson:
> There shouldn't be a reason to have separate annotations sets for each type of migration Job, as they usually have similar needs (e.g. you want to avoid sidecar injection on _all_ migration Pods, or want to clean up _all_ migrations Jobs, not just init or pre-upgrade migrations) and, worst case, don't run at the same times anyway.

There are a few use cases for individual annotations for each type of migration Job, systems like ArgoCD allow some granular control around which resources get synced first based on their annotation which one would want to set for the init migrations job but not the other jobs. Would it be worth calling out an individual key for each job and an additional key for the shared annotations:
```
migrations:
  ...
  jobAnnotations: {}
  initJobAnnotations: {}
  preUpgradeJobAnnotations: {}
  postUpgradeJobAnnotations: {}
```

rainest:
I can see the rationale for that, but it's orthogonal to other goals. We want to avoid values.yaml configuration that requires configuring similar items (e.g. migrations Jobs, services for the Kong container) individually: that's why we have templates for the manifests that reuse values content and boilerplate YAML.

We have some leeway with what we can add for `helm template` workflows, but it's limited. Something like the the recent change to [set namespaces explicitly](https://github.com/Kong/charts/pull/193) is fine because it still reuses the value across the board, and adding separate annotation injection points for the Job and Job Pod template is fine because it's useful for both releases managed by Helm and releases that aren't (for example, the original "don't inject mesh sidecars" use case applies to either). Something like this that'd require Helm users to configure the same annotation per Job is much harder to justify.

What may work is instead leveraging other template features to accomplish this via other means, though it'd depend on the specifics of what ArgoCD can do and your workflow. We do provide [a mechanism to disable each of the migrations jobs](https://github.com/Kong/charts/blob/main/charts/kong/values.yaml#L266-L270), so possibly re-running the template command with them disabled piecemeal and a different Job annotation value each time would work?

For example, running a template command with `--set migrations.preUpgrade=false --set migrations.postUpgrade=false` renders _only_ the init migrations job. There's one weird caveat here, however: we actually have an undocumented `migrations.init` value, but there's either a bug in the template logic or template engine, and setting it to `false` doesn't actually override the default, and Helm will render it anyway. However, we also condition it on whether Helm is performing an initial install, and template respects `--is-upgrade`: running something like `helm template --is-upgrade --set migrations.postUpgrade=false` renders the pre-upgrade Job only.

The other option, which is perhaps a bit more sane, is to patch the resources after they're rendered. I have less experience with Kustomize in general, but it seems possible, and Kustomize is built into modern versions of kubectl (the `-k` flag for `kubectl apply -f rendered-template -k /path/to/kustomizations`).

You should then be able to use a [label selector](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/patchMultipleObjects.md) to apply an [annotation transformer](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/transformerconfigs/README.md#annotations-transformer). Helm renders each of those jobs with a unique label:

```
app.kubernetes.io/component: init-migrations
app.kubernetes.io/component: post-upgrade-migrations
app.kubernetes.io/component: pre-upgrade-migrations
```
So while I don't have a complete solution ready for that, it should be viable with the existing template output.

seanson:
Given the individual annotation configuration is undesirable, I think the global `migrations.jobAnnotations` as suggested is a good fit. This would allow us to configure the shared annotations and then dip into Kustomize output translations for individual job management. This adds extra complexity on our side but it sounds like it's better for the Helm chart overall.

rainest (approved):

rainest:
Addendum to the previous comment: I figured out why the hidden `migrations.init` toggle wasn't working, and have opened #247 to address it. Once that's released (presumably in either a future 1.13.1 or 1.14.0), you'll be able to pass `--set migrations.init=false` to disable it rather than relying on the `--is-upgrade` workaround.

PR #223, opened 2020-11-02
[kong] Add runAsUser parameter to ingress controller

buzzsurfr:
Adds the ability to configure the user ID for the ingress controller. In this example, this is useful for connecting kong to AWS App Mesh. By specifying the user ID, the kong traffic will not be filtered by the envoy proxy.

  - None

Per conversation with @shapirov103.

#### Checklist

### DIFF
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -570,6 +570,7 @@ section of `values.yaml` file:
 | admissionWebhook.enabled           | Whether to enable the validating admission webhook                                    | false                                                                        |
 | admissionWebhook.failurePolicy     | How unrecognized errors from the admission endpoint are handled (Ignore or Fail)      | Fail                                                                         |
 | admissionWebhook.port              | The port the ingress controller will listen on for admission webhooks                 | 8080                                                                         |
+| securityContext                    | Set the securityContext for ingress controller                                        | `{}`                                                                         |
 
 For a complete list of all configuration values you can set in the
 `env` section, please read the Kong Ingress Controller's
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -560,7 +560,6 @@ section of `values.yaml` file:
 | image.tag                          | Version of the ingress controller                                                     | 0.9.1                                                                        |
 | readinessProbe                     | Kong ingress controllers readiness probe                                              |                                                                              |
 | livenessProbe                      | Kong ingress controllers liveness probe                                               |                                                                              |
-| runAsUser                          | User ID (UID) of the ingress controller                                               |                                                                              |
 | installCRDs                        | Create CRDs. **FOR HELM3, MAKE SURE THIS VALUE IS SET TO `false`.**  Regardless of value of this, Helm v3+ will install the CRDs if those are not present already. Use `--skip-crds` with `helm install` if you want to skip CRD creation.                 | true                                                                         |
 | serviceAccount.create              | Create Service Account for ingress controller                                         | true
 | serviceAccount.name                | Use existing Service Account, specify its name                                        | ""
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -23,40 +23,28 @@ $ helm install kong/kong --generate-name --set ingressController.installCRDs=fal
 
 ## Table of contents
 
-- [Kong for Kubernetes](#kong-for-kubernetes)
-- [TL;DR;](#tldr)
-- [Table of contents](#table-of-contents)
 - [Prerequisites](#prerequisites)
-- [Important: Helm 2 vs Helm 3](#important-helm-2-vs-helm-3)
-    - [Helm 2](#helm-2)
-    - [Helm 3](#helm-3)
+- [Helm 2 vs Helm 3](#important-helm-2-vs-helm-3)
 - [Install](#install)
 - [Uninstall](#uninstall)
-- [FAQs](#faqs)
 - [Kong Enterprise](#kong-enterprise)
+- [FAQs](#faqs)
 - [Deployment Options](#deployment-options)
   - [Database](#database)
-    - [DB-less  deployment](#db-less-deployment)
   - [Runtime package](#runtime-package)
   - [Configuration method](#configuration-method)
   - [Separate admin and proxy nodes](#separate-admin-and-proxy-nodes)
   - [Standalone controller nodes](#standalone-controller-nodes)
   - [Hybrid mode](#hybrid-mode)
-    - [Certificates](#certificates)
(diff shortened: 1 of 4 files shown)


rainest:
@buzzsurfr or @shapirov103 do you have documentation that covers AWS App Mesh's need for this/proxy exemption behavior?

This seems a bit unusual as the controller often communicates inside the Pod (over localhost) only, and other mesh proxy exemptions we've seen use annotations to mark components that should not be proxied.

buzzsurfr:
I'm working on making the App Mesh documentation more explicit, but App Mesh utilizes the [Proxy Authorization](https://docs.aws.amazon.com/app-mesh/latest/userguide/proxy-authorization.html) feature within ECS (and the EKS VPC CNI), which includes a parameter to ignore specific UIDs or GIDs inside of a process namespace (e.g. ECS Task border, EKS Pod border). See [ProxyConfiguration](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_ProxyConfiguration.html) for the specific details.

By adding `runAsUser`, we could configure the Kong ingress controller to use a process ID that is ignored by the Proxy Authorization. It doesn't have to be `1337`, but I didn't see the harm in having envoy and the controller share a UID, and either way the helm chart enables that level of configuration.

@shapirov103 brought this to my attention at which point I realized we need to better document this, and I plan on enhancing this documentation.

rainest:
Allowing this via a UID override seems rather odd: I'd typically expect the type of requests made by the controller to be exempt from mesh proxying entirely, or be permitted through mesh ACL rules defined elsewhere.

- If the controller is running inside the same Pod as Kong, its communication with the Kong admin API is over localhost, and isn't proxied as such.
- If the controller is running in its own Pod, its communication with the Kong admin API is limited by whatever the mesh proxy enforces for the admin API's Service, and the mesh presumably has other means to allow that (something to indicate that Pod X is allowed to talk to Service Y via annotations or labels on those K8S resources).
- When communicating with the K8S API, the mesh proxy just isn't involved, as the K8S API provides its own means to enforce controls via the bound service account and its roles. If it is still handled through the mesh proxy, any mesh proxy restrictions rely on Pod-level abstractions, rather than inspecting or modifying the internals of containers running within it.

rainest (changes requested):
If we are going to implement something that relies on overriding the run as user, I request that it be flexible enough to allow other securityContext changes as well. I.e., rather than defining a securityContext block and populating individual settings within it from values, we allow you to define the entire block. This would work similar to what we offer for the Kong container:

https://github.com/Kong/charts/blob/kong-1.11.0/charts/kong/values.yaml#L527-L528
https://github.com/Kong/charts/blob/kong-1.11.0/charts/kong/templates/deployment.yaml#L239-L240

That approach (allowing users to choose whatever `securityContext` block they like) allows setting the `runAsUser` without preventing them from using other `securityContext` features.

shapirov103:
@rainest 

We are addressing use case #1: ingress controller and proxy are in the same pod. Based on the feedback from Claudio and from my experience deploying Kong with App Mesh, I see that there is communication between the ingress controller and the API server. This is not the communication between the proxy and the controller itself.

Tactically, for the Kong Ingress controller, this configuration option will disable egress filtering for the ingress controller container in the pod (consisting of multiple containers). This is a workaround at the moment that enables us to address the issue at the level of a single node in the mesh. There will be a better approach for this in the future https://github.com/aws/aws-app-mesh-roadmap/issues/241.

buzzsurfr:
I missed the `securityContext` block already in the values, I think we could use that and probably ~close/cancel this PR~ change our approach. We may still want a separate securityContext for ingress and service.

On the ingress controller:
>If the controller is running inside the same Pod as Kong, its communication with the Kong admin API is over localhost, and isn't proxied as such.

Does the ingress controller need to talk to the kubernetes API? When we tested with the ingress controller in the same pod as the ingress and app container, we had to exclude the controller from the proxying. What traffic does the ingress controller call (I presume it's the kubernetes API).

shapirov103:
The PR is needed since we want a separate securityContext for ingress and proxy. Applying the change to the proxy is counter to the intention (and breaks the deployment). I assume previous change request from @rainest applies: override will cover the full context.

buzzsurfr:
Commit has been updated to use a broad `securityContext` object specific for the ingress controller.

rainest:
Copying some questions I'd sent out elsewhere here: why would you need to override the UID for the controller container? Would you be unable to write a policy that exempts the existing UID?

That UID is static: we create user 1000 when building the image and then set Docker to use it, so I'd expect a policy exempting API server communication from processes run by UID 1000 would be fine: https://github.com/Kong/kubernetes-ingress-controller/blob/main/Dockerfile#L27

I wouldn't expect the controller to ever need any of the other features exposed by security contexts (enabling additional kernel capabilities or modifying SELinux policy), and we're not aware of other uses for overriding the UID.

buzzsurfr:
> why would you need to override the UID for the controller container?

This how App Mesh ignores specific traffic, by UID/GID, by destination IP, or by port. We chose to use UID since it does not require a lookup of the kubernetes service to get the IP:port.

~I could add a statement in our proxy configuration to ignore port `1000`. Is that UID used by the proxy and ingress controller, or only the ingress controller?~ I checked our proxy authorization config and we're bound to only have a single UID. We could potentially change the GID of the ingress controller to `1337` and ignore that, which shouldn't affect anything inside the container since it's single process. WDYT?

Would you be unable to write a policy that exempts the existing UID?

We could, but do _not_ want to exclude the kong proxy, so we need to make sure that it does not use UID `1000`.

shapirov103:
@rainest 

UID 1000 is pretty common and is very likely to cause collision in the service mesh. Given the current state we think that customers should be able to control it to avoid any undesired impact. 

The feature as it stands:
1. Does not impact the solution negatively
2. Provides clear path for integration with AWS App Mesh (we cannot override it on our end as it stands now).

I suppose there may be other use cases for security context control, so applying the change will improve potential extensibility (i can see that other than App Mesh you may not observe clear use cases atm). Please let us know if you have any concerns.

shapirov103:
After discussion with Kong product management: if the change has value for Kong in general (ability to override security context not just for proxy but also for the ingress controller) then it is a valuable contribution. Otherwise the change is not accepted by the Kong PM, due to the extra tuning options that may be confusing for the end users. For the App Mesh integration scenario: this approach should be treated as temporary, due to lack of more direct controls in the product to affect behavior which we will address on the AWS side.

rainest:
As best I know our partnerships team is good to go with the patch-after-install approach as an interim solution, and have the info they need to proceed with that. Closing this out as such--if that's not the case please let them or I know.

PR #229, opened 2020-11-14
[kong] Release 1.12.0 to next

### LINKED ISSUE #144: [kong] NOTES.txt warnings include extra newlines
The NOTES.txt template may include one or more [warnings](https://github.com/Kong/charts/blob/kong-1.5.0/charts/kong/templates/NOTES.txt#L13-L33) if performing an upgrade or install with deprecated configuration in values.yaml. Because of limitations in template newline chomping, this can print excess newlines.

Based on review of other charts, we can avoid this by moving string generation into a helper template that builds the complete text with string concatenation rather than templating each warning individually.

### LINKED ISSUE #227: [kong] update default worker count to 2
The Helm chart sets `nginx_worker_processes: "1"`. The original motivation for this default, against the kong default of `auto` was users seeing resource exhaustion when they installed Kong. This usually happened with k8s worker nodes containing a larger number of CPUs/vCPUs.
With the default of `1`, we run into the other problem where a lot of users run into severe performance problems, especially with KIC DB-less deployments, where a sync (POST /config) request can introduce significant latencies on proxy requests.
Given the number of external and internal reports on this specific, we should update the default value to `2`.

TODO:
- [ ] Update the default value to 2 in this repository
- [ ] Update the default value to 2 in Kong Ingress Controller's [repository](https://github.com/Kong/kubernetes-ingress-controller/blob/3a72f85f3baca0e630b2c117896136b34c10e2c3/hack/dev/dbless/base/kong.yaml#L26)

rainest:
Releases 1.12 to next. Includes several additional changes:

- Bump the default Enterprise tags to the latest 2.1 patch, 2.1.4.1.
- Increase the default worker count to 2.
- Handle deprecation warnings via a template to simplify and improve formatting.

  - fixes #227 
  - fixes #144

Re conversations about worker count increase and memory usage from planning:
- By default, the chart does not apply memory limits to Kong deployments. However, it did include commented example limits with an unreasonably low default limit (128MB). From personal experience, this limit is too low for _any_ deployment (even single-worker deployments), and would cause issues if applied. I've increased the commented example to 256MB, which is much more reasonable, though that's essentially a minimum lower bound for Kong to function at all rather than what I'd expect for most actual deployments. Non-demo/test environments that actually handle traffic will likely start around 1GB, and will usually have even higher limits.
- There was a brief interlude where additional workers allocated resulted in unexpected higher memory use due to removal of limits elsewhere. AFAIK we've since re-introduced those limits (https://github.com/Kong/kong/pull/5864), and increasing worker count no longer causes the consumption issues we observed in earlier 2.x releases, so I don't expect the chart change to cause issues.

We still want to explore resource usage guidelines further, but the minimums proposed here should be reasonable.

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -2,6 +2,29 @@
 
 ## 1.11.0
 
+### Improvements
+
+* Increased default worker count to 2 to avoid issues with latency during
+  blocking tasks, such as DB-less config updates. This change increases memory
+  usage, but the increase should not be a concern for any but the smallest
+  deployments (deployments with memory limits below 512MB).
+* Updated default Kong version to 2.2.
+  ([#221](https://github.com/Kong/charts/pull/221))
+* Updated default Kong Enterprise version to 2.1.4.1.
+* Added a means to mount extra ConfigMap and Secret resources.
+  ([#208](https://github.com/Kong/charts/pull/208))
+* Added configurable annotations for migration Jobs.
+  ([#219](https://github.com/Kong/charts/pull/219))
+* Added template for deprecation warnings to automate formatting and avoid
+  excess newlines.
+
+### Fixed
+
+* Upgrades no longer force auto-scaling Deployments back to the replica count.
+  ([#222](https://github.com/Kong/charts/pull/222))
+
+## 1.11.0
+
 ### Breaking changes
 
 * Kong Ingress Controller 1.0 removes support for several deprecated flags and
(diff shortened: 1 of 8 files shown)


mflendrich:
LGTM but:
- this does not fix #227 yet - after this PR, the checkbox below remains unsatisfied:
    > [ ] Update the default value to 2 in Kong Ingress Controller's repository
- for future PRs, please avoid mix features and release chores in one PR

Please merge without squashing, too.

mflendrich (approved):

mflendrich:
Part of #232.

PR #231, opened 2020-11-14
[kong] Add kong 'namespace' value

### LINKED ISSUE #230: [kong] Ability to choose the namespace to deploy kong
It would be great to be able to choose the namespace where to deploy kong using a helm value.

We would have `.Values.namespace` and by default, the namespace could remain `.Release.Namespace`. see the PR...

It can be useful in some use cases where you do not have control over `.Release.Namespace`... ex: kong being part of dependencies.

jiraguha:
solves  #230
  - fixes #230

#### Checklist

### DIFF
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -579,7 +579,7 @@ For a complete list of all configuration values you can set in the
 
 | Parameter                          | Description                                                                           | Default             |
 | ---------------------------------- | ------------------------------------------------------------------------------------- | ------------------- |
-| namespace                          | Namespace to deploy Kong                                                              | `""`                |
+| namespace                          | Namespace to deploy chart resources                                                   |                     |
 | deployment.kong.enabled            | Enable or disable deploying Kong                                                      | `true`              |
 | autoscaling.enabled                | Set this to `true` to enable autoscaling                                              | `false`             |
 | autoscaling.minReplicas            | Set minimum number of replicas                                                        | `2`                 |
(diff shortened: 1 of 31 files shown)


charts/kong/values.yaml
```diff
@@ -25,6 +25,9 @@ deployment:
   #   - name: sidecar
   #     image: sidecar:latest
 
+# Namespace to deploy Kong.
+namespace: ""
```
```suggestion
# Override namepsace for Kong chart resources. By default, the chart creates resources in the release namespace.
# This may not be desirable when using this chart as a dependency.
# namespace: "example"
```

charts/kong/README.md
```diff
@@ -579,6 +579,7 @@ For a complete list of all configuration values you can set in the
 
 | Parameter                          | Description                                                                           | Default             |
 | ---------------------------------- | ------------------------------------------------------------------------------------- | ------------------- |
+| namespace                          | Namespace to deploy Kong                                                              | `""`                |
```
rainest:
```suggestion
| namespace                          | Namespace to deploy chart resources                                                   |                     |
```

rainest (changes requested):
Template change looks fine. I've added suggestions that clarify where it applies (to all namespaced resources) and when to use it (when you don't want to use the release namespace).

Template logic treats the empty string and unset as equivalent for the default logic, though that's kind of a special case for strings--our convention is to comment out values that we leave up to template logic if nothing is set, so I've reworked that a bit also. End result is the same if you provide a non-empty string.

PR #236, opened 2020-12-08
[kong] clarify HPA requirements

rainest:
Adds values.yaml comments to clarify HPA requirements and `ingressController.resources` usage.

Thought it was possible to do this instead at the Pod level, since the controller doesn't consume a significant amount of resources and there's not much reason to handle it separately from the Kong container's requests, but it looks like that's a no go--docs sorta imply you can do it, but there's not actually any way to do that in the Pod spec.

#### Checklist

### DIFF
--- a/charts/kong/values.yaml
+++ b/charts/kong/values.yaml
@@ -368,7 +368,14 @@ ingressController:
     periodSeconds: 10
     successThreshold: 1
     failureThreshold: 3
+  # Remove "{}" and uncomment the block below to set controller resources
   resources: {}
+  # limits:
+  #  cpu: 100m
+  #  memory: 256Mi
+  # requests:
+  #  cpu: 50m
+  #  memory: 128Mi
 
 # -----------------------------------------------------------------------------
 # Postgres sub-chart parameters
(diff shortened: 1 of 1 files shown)


charts/kong/values.yaml
```diff
@@ -368,7 +368,14 @@ ingressController:
...
     successThreshold: 1
     failureThreshold: 3
+  # Remove "{}" and uncomment the block below to set controller resources
   resources: {}
+  # limits:
+  #  cpu: 100m
+  #  memory: 256Mi
+  # requests:
+  #  cpu: 50m
+  #  memory: 128Mi
```
mflendrich:
```suggestion
  resources: {}
  # Example reasonable setting for "resources":
  # resources:
  #   limits:
  #     cpu: 100m
  #     memory: 256Mi
  #   requests:
  #     cpu: 50m
  #     memory: 128Mi
```

Two comments here:
1. One does not edit this `values.yaml` (embedded in the chart) directly (and we don't recommend copying it entirely, but creating a minimal derivative), so the notion of manipulating parts of this file seems slightly misleading. I recommend an alternative of giving a complete example of the "resources" block.
2. Indentation: let's be consistent with the rest of the file on "2 spaces per level"

charts/kong/values.yaml
```diff
@@ -480,6 +487,8 @@ deploymentAnnotations:
   traffic.sidecar.istio.io/includeInboundPorts: ""
 
 # Enable autoscaling using HorizontalPodAutoscaler
+# When configuring an HPA, you must set resource limits on all containers via
```
mflendrich:
```suggestion
# When configuring an HPA, you must set resource requests on all containers via
```

HPA cares about requests, not limits.

mflendrich (approved):

PR #238, opened 2020-12-11
[kong] add nodeSelector to migration Jobs

rainest:
The nodeSelector value was previously only applied to the main Deployment. It should be applied to migration Jobs also, since they run Kong also, and have the same platform requirements.

  - fixes FTI-2152

#### Checklist

### DIFF
--- a/charts/kong/templates/migrations-post-upgrade.yaml
+++ b/charts/kong/templates/migrations-post-upgrade.yaml
@@ -60,7 +60,7 @@ spec:
       {{- include "kong.podsecuritycontext" . | nindent 8 }}
       {{- if .Values.nodeSelector }}
       nodeSelector:
-      {{ toYaml .Values.nodeSelector | indent 8 }}
+      {{- toYaml .Values.nodeSelector | nindent 8 }}
       {{- end }}
       restartPolicy: OnFailure
       volumes:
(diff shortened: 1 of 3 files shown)


charts/kong/templates/migrations-post-upgrade.yaml
```diff
@@ -58,6 +58,10 @@ spec:
         {{- include "kong.volumeMounts" . | nindent 8 }}
       securityContext:
       {{- include "kong.podsecuritycontext" . | nindent 8 }}
+      {{- if .Values.nodeSelector }}
+      nodeSelector:
+      {{ toYaml .Values.nodeSelector | indent 8 }}
```
mflendrich: Before this PR, all 3 edited files prefer `nindent` (with `{{-`) to `indent`ing - this PR appears to be the first introduction of `indent`.

Is there a rationale for breaking consistency? If not, could we do:

```suggestion
      {{- toYaml .Values.nodeSelector | nindent 8 }}
```
throughout this PR?

rainest: Not really. Copy-pasta from the deployment template; newline-chomping convention isn't consistent throughout the templates.

mflendrich (approved):
Conditional LGTM, assuming that the changes have been tested to work.

rainest:
> assuming that the changes have been tested to work.

It's manual since we don't have an exhaustive test values.yaml with all non-default settings. I set it to an example value to confirm that the indentation is correct/didn't result in invalid Job resources rendering.

PR #243, opened 2020-12-16
[kong] Add link to external secrets in the documentation

ragoncsa:

Adding link to external-secrets as a way to handle secrets on a more secure way to improve the documentation. This came up in a support ticket, it may be worth sharing in the docs.

#### Checklist

### DIFF
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -669,9 +669,10 @@ configuration can be placed under the `.env` key.
 
 ### Prerequisites
 
-Note that the steps below prescribe the creation of Kubernetes secrets with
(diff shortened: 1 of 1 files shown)


rainest:
Leaving this out as it pertains to general cluster management, not Kong or this chart specifically. The security concerns for Secret resources are covered in the base Kubernetes documentation: 

https://kubernetes.io/docs/concepts/configuration/secret/
https://kubernetes.io/docs/concepts/security/overview/#infrastructure-security

We're not prescriptive on the use of Secret storage in etcd versus external brokers.

PR #247, opened 2021-01-05
[kong] fix migrations.init

### LINKED ISSUE #235: Migrations job created in kubernetes using kong/kong chart

rainest:
Makes the hidden `migrations.init` toggle actually work, and fix the template ignoring the legacy `runMigrations` toggle for the init-migrations job.

More details in the commit message, but tl;dr you can't actually use `$someVar | default true`.

  - fixes #235 

#### Checklist

### DIFF
--- a/charts/kong/templates/migrations.yaml
+++ b/charts/kong/templates/migrations.yaml
@@ -1,10 +1,19 @@
 {{- if .Values.deployment.kong.enabled }}
 {{- if .Release.IsInstall -}}
-{{/* .migrations.init isn't normally exposed in values.yaml. In testing,
-however, adding it with value "false" didn't actually disable this job. Not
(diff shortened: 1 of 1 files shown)


mflendrich:
Any reason against dropping the runMigrations flag whatsoever?

If there's concern that somebody might upgrade and fail to notice that runMigrations is no longer supported by us, we might add a check that fails template rendering if that flag is not unset.

Otherwise, with two parallel feature flag sets covering the same thing, we fall prey to logical errors.

rainest:
The current backlog of deprecated config (including that) will be removed entirely for 2.0. We didn't want to break the old config immediately.

mflendrich (approved):
Lgtm assuming that the dichotomy (as discussed in comments) goes away in 2.0.

PR #249, opened 2021-01-07
[repo] replace old Helm stable with Bitnami

rainest:
Per https://helm.sh/blog/new-location-stable-incubator-charts/ the
previous location used here is now unavailable, breaking releases.

We furthermore no longer actually use the dependency in the old repo in
the actual chart, so changing this to the Bitnami repo rather than the
new Helm repo.

See failure at https://github.com/Kong/charts/runs/1664313187?check_suite_focus=true

CI change only; going direct to main.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -58,7 +58,7 @@ jobs:
 
       - name: Add dependency chart repos
         run: |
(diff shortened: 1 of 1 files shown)


mflendrich (approved):
It looks like a regression that we couldn't have prevented with tests (more so by watching registry announcements).

Issue #252, opened 2021-01-14
Invalid pod annotation for user-provided webhook certs


mflendrich:
Per https://github.com/Kong/charts/issues/126

> you can specify your own certificate, but the chart will always add the checksum annotation regardless

Current behavior: when the user provides a custom admission controller webhook certificate, an annotation on KIC pods includes a checksum of another (auto-generated) certificate, which differs from run to run, causing unnecessary pod restarts
Desired behavior: when the user provides a custom webhook certificate, subsequent helm updates shoud not modify the cert checksum annotation, and therefore not lead to a rolling restart of pods

cc @agaffney who originally reported this issue

rainest:
Alright, this is more difficult than expected. I thought that you'd be able to just add your own Secrets with the names we use for the generated certificate and lookup would use that, but the chart attempts to [create new Secrets](https://github.com/Kong/charts/blob/kong-1.14.3/charts/kong/templates/admission-webhook.yaml#L38-L66) with that content, and they'll collide: 

> rendered manifests contain a resource that already exists. Unable to continue with install: Secret "release-kong-validation-webhook-ca-keypair" in namespace "whatever" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "release"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "whatever"

We can't condition the those blocks on the lookup succeeding because if the certs _were_ auto-generated, an upgrade that doesn't include those blocks will delete the Secrets.

Furthermore, functionality for setting your own certificate doesn't actually exist yet--the quoted bit from my old comment in this issue should have been a hypothetical about how we'd handle the checksum if you could. There are a couple things to keep in mind if we want to add that:
- The file and value certificate flags [are mutually exclusive](https://github.com/Kong/kubernetes-ingress-controller/blob/v1.1.1/cli/ingress-controller/main.go#L501-L507). We'd need to auto-detect either type of user configuration in controller env if we don't provide a dedicated setting to disable certificate generation.
- `caBundle` configuration for the hook is optional but absolute: if you do provide it, the API server won't trust webhook client certificates issued by the cluster CA.

[...]

PR #256, opened 2021-01-15
[kong] re-use admission certificates if present

### LINKED ISSUE #126: Kong pods get replaced on every update to helm release when admission controller is enabled
When changing something in the helm release values that generally shouldn't cause a restart of the Kong pods, they get restarted anyway when the admission controller is enabled. This appears to happen because the checksum for the admission webhook config is added as an annotation on the Kong pods, and the admission webhook config changes every time because a new CA/cert is being generated each time.

It would be nice to have an option to provide your own CA/cert for the admission webhook to prevent this unnecessary Kong restart. It seems that the cert is dependent on the CN generated from the release/service name which complicates things a bit, but it should still be possible.

### LINKED ISSUE #253: auto-generated webhook cert should not be overwritten by helm
Context: https://github.com/Kong/charts/issues/126#issuecomment-625459597

Current behavior: helm overwrites the auto-generated admission controller webhook cert `Secret` if it existed before update.
Desired behavior: if the auto-generated admission controller webhook cert `Secret` exists in the cluster, helm release/update should not try to overwrite it

cc @agaffney who originally reported #126

rainest:
Reuses admission webhook certificate Secrets if they are already present.

  - fixes #126 
  - fixes #253 

We currently generate our checksum by using the entire admission webhook template output as input to SHA256. For some reason I cannot determine, the checksum still changes with the new lookup functionality between installs and upgrades, although a diff between the two shows no difference in the generated resources. Subsequent upgrades do not change the checksum.

To avoid this, I've modified the checksum generator to extract the `tls.crt` lines from the template, concatenate those, and checksum that, since we shouldn't need to care about anything else. It's ugly, but it works.

IMO this de facto solves #252 also: the certificate and CA Secret names are predictable, and if you create those Secrets in advance of the install, it should use them instead. Need to check to confirm. We'd only need additional changes for that if we want to make the Secret names user-selectable.

#### Checklist

### DIFF
--- a/charts/kong/templates/admission-webhook.yaml
+++ b/charts/kong/templates/admission-webhook.yaml
@@ -1,7 +1,20 @@
-{{- if .Values.ingressController.admissionWebhook.enabled }}
-{{- $cn := printf "%s.%s.svc" ( include "kong.service.validationWebhook" . ) ( include "kong.namespace" . ) }}
+{{- if .Values.ingressController.admissionWebhook.enabled -}}
+{{- $cn := printf "%s.%s.svc" ( include "kong.service.validationWebhook" . ) ( include "kong.namespace" . ) -}}
 {{- $ca := genCA "kong-admission-ca" 3650 -}}
 {{- $cert := genSignedCert $cn nil nil 3650 $ca -}}
+{{- $certCert := $cert.Cert -}}
+{{- $certKey := $cert.Key -}}
+{{- $caCert := $ca.Cert -}}
+{{- $caKey := $ca.Key -}}
+
+{{- $caSecret := (lookup "v1" "Secret" (include "kong.namespace" .) (printf "%s-validation-webhook-ca-keypair" (include "kong.fullname" .))) -}}
+{{- $certSecret := (lookup "v1" "Secret" (include "kong.namespace" .) (printf "%s-validation-webhook-keypair" (include "kong.fullname" .))) -}}
+{{- if $certSecret -}}
+{{- $certCert = (b64dec (get $certSecret.data "tls.crt")) -}}
+{{- $certKey = (b64dec (get $certSecret.data "tls.key")) -}}
+{{- $caCert = (b64dec (get $caSecret.data "tls.crt")) -}}
+{{- $caKey = (b64dec (get $caSecret.data "tls.key")) -}}
+{{- end -}}
 kind: ValidatingWebhookConfiguration
 {{- if .Capabilities.APIVersions.Has "admissionregistration.k8s.io/v1" }}
 apiVersion: admissionregistration.k8s.io/v1
@@ -30,7 +43,7 @@ webhooks:
     - kongconsumers
     - kongplugins
   clientConfig:
-    caBundle: {{ b64enc $ca.Cert }}
+    caBundle: {{ b64enc $caCert }}
     service:
       name: {{ template "kong.service.validationWebhook" . }}
       namespace: {{ template "kong.namespace" . }}
@@ -55,12 +68,24 @@ spec:
 apiVersion: v1
 kind: Secret
 metadata:
+  name: {{ template "kong.fullname" . }}-validation-webhook-ca-keypair
+  namespace:  {{ template "kong.namespace" . }}
+  labels:
+    {{- include "kong.metaLabels" . | nindent 4 }}
+type: kubernetes.io/tls
+data:
+    tls.crt: {{ b64enc $caCert  }}
+    tls.key: {{ b64enc $caKey  }}
+---
+apiVersion: v1
+kind: Secret
+metadata:
   name: {{ template "kong.fullname" . }}-validation-webhook-keypair
   namespace:  {{ template "kong.namespace" . }}
   labels:
     {{- include "kong.metaLabels" . | nindent 4 }}
 type: kubernetes.io/tls
 data:
-  tls.crt: {{ b64enc $cert.Cert }}
-  tls.key: {{ b64enc $cert.Key }}
+  tls.crt: {{ b64enc $certCert }}
+  tls.key: {{ b64enc $certKey }}
 {{ end }}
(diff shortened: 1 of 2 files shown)


rainest:
Hold because this change requires Helm 3 and cannot go in until we've removed support for Helm 2 in chart 2.x. It should otherwise be fine to review (and if anyone can figure out why the install/upgrade checksum thing happens or some neater way of working around that, I'd be quite grateful).

As CI helpfully notes, we also need to upgrade CI to use Helm 3.

charts/kong/templates/deployment.yaml
```diff
@@ -29,7 +29,9 @@ spec:
     metadata:
       annotations:
         {{- if .Values.ingressController.admissionWebhook.enabled }}
-        checksum/admission-webhook.yaml: {{ include (print $.Template.BasePath "/admission-webhook.yaml") . | sha256sum }}
+        {{/* Generating a checksum from the entire template causes the checksum to change depending on whether you install or upgrade
```
shaneutt: :+1:

shaneutt (approved):

charts/kong/templates/admission-webhook.yaml
```diff
@@ -1,7 +1,20 @@
-{{- if .Values.ingressController.admissionWebhook.enabled }}
-{{- $cn := printf "%s.%s.svc" ( include "kong.service.validationWebhook" . ) ( include "kong.namespace" . ) }}
+{{- if .Values.ingressController.admissionWebhook.enabled -}}
+{{- $cn := printf "%s.%s.svc" ( include "kong.service.validationWebhook" . ) ( include "kong.namespace" . ) -}}
 {{- $ca := genCA "kong-admission-ca" 3650 -}}
 {{- $cert := genSignedCert $cn nil nil 3650 $ca -}}
+{{- $certCert := $cert.Cert -}}
+{{- $certKey := $cert.Key -}}
+{{- $caCert := $ca.Cert -}}
+{{- $caKey := $ca.Key -}}
```
mflendrich: Should this be in the `else` branch of the `if $certSecret`?

I don't see why we should generate certs if the secret is already present.

Maybe this is the reason for 
> why the install/upgrade checksum thing happens

?

charts/kong/templates/admission-webhook.yaml
```diff
@@ -54,13 +67,25 @@ spec:
...
 kind: Secret
+metadata:
+  name: {{ template "kong.fullname" . }}-validation-webhook-ca-keypair
+  namespace:  {{ template "kong.namespace" . }}
+  labels:
+    {{- include "kong.metaLabels" . | nindent 4 }}
+type: kubernetes.io/tls
+data:
+    tls.crt: {{ b64enc $caCert  }}
+    tls.key: {{ b64enc $caKey  }}
```
mflendrich, line 78:
```suggestion
    tls.crt: {{ b64enc $caCert }}
    tls.key: {{ b64enc $caKey }}
```
nit

charts/kong/templates/admission-webhook.yaml
```diff
@@ -1,7 +1,20 @@
...
 {{- $cert := genSignedCert $cn nil nil 3650 $ca -}}
+{{- $certCert := $cert.Cert -}}
+{{- $certKey := $cert.Key -}}
+{{- $caCert := $ca.Cert -}}
+{{- $caKey := $ca.Key -}}
+
+{{- $caSecret := (lookup "v1" "Secret" (include "kong.namespace" .) (printf "%s-validation-webhook-ca-keypair" (include "kong.fullname" .))) -}}
+{{- $certSecret := (lookup "v1" "Secret" (include "kong.namespace" .) (printf "%s-validation-webhook-keypair" (include "kong.fullname" .))) -}}
+{{- if $certSecret -}}
+{{- $certCert = (b64dec (get $certSecret.data "tls.crt")) -}}
```
mflendrich: The `cert` prefix in `certCert` and `certKey` feels odd to me. Maybe `webhookCert` and `webhookKey`?

mflendrich (commented):
Looks good in general; some comments attached.

Also, why is the test failing (on failure to find the `lookup` function)?

charts/kong/templates/admission-webhook.yaml
rainest: No particular reason. I didn't move it from its original location and it provides something to initialize the certificate variables to.

This shouldn't affect the checksum--as best I know, that's generated from the text output of the template only, so the variable initialization at the top shouldn't matter, as it doesn't output text and prunes all whitespace before `kind: ValidatingWebhookConfiguration`.

The result looks identical in the final output, though that makes some changes to the individual templates' output (it sorts resources at least--not sure of all the changes it makes). There's presumably something that's either pruned by that pass or something that's different when using it as a named template (the SHA input) versus as a rendered file (the content we pass to K8S and see in Helm debug output). Unfortunately Helm doesn't provide an easy way of seeing the former.

rainest:
> Also, why is the test failing (on failure to find the lookup function)?

It requires Helm 3. CI still uses Helm 2.

rainest:
CI now has a different problem!

> Error:  templates/controller-rbac-resources.yaml: the kind "rbac.authorization.k8s.io/v1beta1 Role" is deprecated in favor of "rbac.authorization.k8s.io/v1 Role"

https://github.com/rainest/charts has a PoC of the updated flow with a cheat: https://github.com/rainest/charts/commit/f4c7da0f919f69840b8a027d9a559ca486377f11

This should work with a proper fix for https://github.com/Kong/kubernetes-ingress-controller/issues/801. For roles, apparently just bumping the version is fine. CRDs require other changes, however.

rainest:
The checksum hack wasn't actually necessary--I thought about why we even had it, looked back in ancient history, and realized it was a workaround to intentionally redeploy the controller because we couldn't keep the certificate stable :facepalm: https://github.com/helm/charts/pull/20051

rainest:
Rebase following 2635265811bf6d0e334ee19307e9473e9be39a54. CI is now happy/this is no longer blocked on KIC#801.

PR #263, opened 2021-01-26
[repo] add CI to release kong 1-x patches

rainest:
Add a new workflow based on the main workflow for releasing Kong 1.x
chart patch releases after 2.0 is out. It still checks version bumps,
but targets the kong-1.x branch instead of main/master.

This goes into main first to exclude the kong-1.x branch from the default CI job. We need to merge it into main and next after.

The top-level configs (ct-main.yaml and such) will presumably eventually disappear from main and next entirely. The new chart-tester actions for Helm 3 won't use them, and the kong-1.x branch (which will continue to use the old-style workflow) only needs its config in its branch.

See https://github.com/rainest/charts/pull/6 and https://github.com/rainest/charts/actions/runs/513586226 for PoC.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -3,7 +3,7 @@ name: Release Charts
 on:
   push:
     branches:
-      - main
+    - kong-1.x
 
 jobs:
   lint-test:
@@ -20,7 +20,7 @@ jobs:
         uses: helm/chart-testing-action@v1.0.0
         with:
           command: lint
-          config: ct-main.yaml
+          config: ct-kong-1-x.yaml
 
       - name: Create kind cluster
         uses: helm/kind-action@v1.0.0-alpha.3
@@ -32,7 +32,7 @@ jobs:
         uses: helm/chart-testing-action@v1.0.0
         with:
           command: install
-          config: ct-main.yaml
+          config: ct-kong-1-x.yaml
   release:
     needs: lint-test
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


shaneutt (dismissed):

mflendrich (changes requested):
This his how I see branches in this repo:

- `main` is the mainline for `2.x` - the `HEAD` of this branch contains the latest `2.x` release + potentially unreleased fixes
- `next` is the bleeding edge `2.x` - the `HEAD` of this branch at any time contains what will be released as a subsequent minor `2.x` release
- `kong-1.x` is the mainline for `1.x` - the `HEAD` of this branch contains the latest `1.x` release + potentially unreleased fixes. This branch will be orphaned when we cease to support 1.x releases.

If this interpretation is accurate, then I don't understand why this config would be pushed to `main` or `next` at all. I believe that this PR should go to `kong-1.x` only.

rainest:
The `branches-ignore` updates at least need to go into main. Looking at that some more I think we also need to ignore `main` and `kong-1.x` on the non-main PR job also and give them their own PR workflows.

Using the non-main PR job is less of a problem for main (it does omit the version check, but shouldn't), but will actually break 1.x after migrating the non-main stuff to Helm 3.

mflendrich:
> The `branches-ignore` updates at least need to go into main.

Why so? I understand GH workflows to work by taking the `.github` directory in the commit being evaluated (on the branch being evaluated), and matching the branch name against `branches` and `branches-ignore`. I think that `main` doesn't play any role here, but maybe I'm missing something.

rainest:
You're correct; I got turned around while playing with my fork. Went through again and it works fine.

Revised version targets kong-1.x and includes config for it only.

mflendrich (approved):

PR #267, opened 2021-02-01
[kong] Move the end of the {{- else -}} block to correct position

michbeck100:
#### What this PR does / why we need it:

This PR moves the end of the {{- else -}} from https://github.com/Kong/charts/blob/main/charts/kong/templates/service-kong-proxy.yaml#L18 block to correct position. Otherwise the value of `proxy.ingress.tls` is added twice, which resolves to broken yaml.

#### Which issue this PR fixes
*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*
  - fixes #265

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [X] PR is based off the current tip of the `next` branch and targets `next`, not `main`
- [X] New or modified sections of values.yaml are documented in the README.md
- [X] Title of the PR and commit headers start with chart name (e.g. `[kong]`)

PR #268, opened 2021-02-01
[kong] fix legacy proxy ingress template

rainest:
I goofed on the block for legacy proxy Ingress objects. https://github.com/Kong/charts/blob/kong-1.13.0/charts/kong/templates/ingress-proxy.yaml#L22 no longer exists, but I left the end for it in, and that end caught the legacy block gate :facepalm:

The indent on the legacy stuff was also broken, since `toYaml` blocks can't have leading indents, as that will get added on top of the indent function for the first line only. Also should have been 2 to align the elements with the start of the array block. Gotemplate+YAML continue to be great fun.

CI doesn't catch these because secret references without corresponding secrets aren't accepted, so we lack tests for blocks with secret references.

  - fixes #267 

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 1.14.3
+
+### Fixed
+
+* Fix issues with legacy proxy Ingress object template.
+
 ## 1.14.2
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


shaneutt (approved):
LGTM :+1: 

Let's spin off a new issue to document our need for extra tests to catch this in the future if we haven't already.

PR #269, opened 2021-02-02
[kong] support daemonset deployment for kong pods

arjunkrishnasb:
This commit enables the clients to deploy the kong pods as daemonset, so one instance of pod is deployed in each node.

  - fixes #

#### Checklist

### DIFF
--- a/charts/kong/values.yaml
+++ b/charts/kong/values.yaml
@@ -19,6 +19,12 @@ deployment:
     # Setting this to false with ingressController.enabled=true will create a
     # controller-only release.
     enabled: true
+    # Deploy as daemon set
+  ## Setting this to true will deploy kong as a daemonset
+  ## Else by default it creates as a deployment
+  ##
+  # daemonset: true
+  ##
   ## Optionally specify any extra sidecar containers to be included in the deployment
   ## See https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#container-v1-core
   # sidecarContainers:
@@ -590,7 +596,7 @@ enterprise:
     smtp_admin_emails: none@example.com
     smtp_host: smtp.example.com
     smtp_port: 587
-    smtp_auth_type: ''
+    smtp_auth_type: nil
     smtp_ssl: nil
     smtp_starttls: true
     auth:
--- a/charts/kong/FAQs.md
+++ b/charts/kong/FAQs.md
@@ -85,3 +85,18 @@ This occurs if a `RELEASE-NAME-kong-init-migrations` Job is left over from a
 previous `helm install` or `helm upgrade`. Deleting it with
 `kubectl delete job RELEASE-NAME-kong-init-migrations` will allow the upgrade
 to proceed. Chart versions greater than 1.5.0 delete the job automatically.
+
+#### Can we deploy kong as a `Daemonset` ?
+Yes. You can deploy kong as a daemonset. Set `deployment.daemonset: true` to deploy the kong pods as a daemonset.
+Else by default, the pods are deployed as a deployment.
+
+#### How to deploy kong if the kubernetes cluster does not support Loadbalancer type service ?
+One way you can deploy the kong pods and expose it as `NodePort`.
+But the problem is we cannot access the ingress directly with well known ports http or https. 
+Instead we need to access this cluster using http[s]://<host-ip>:<node-port>/<some-path>.
+
+
+Another way is to deploy kong as Daemonset and expose the pods as hostPort. This is not recommended unless 
+it is pretty much needed because of the security reasons. But by this way, we can expose the well known ports
+80 and 443 of the pods as hostport and hence using the http://<host-ip> or https://<host-ip> we can directly
+access the ingress. If we have a loadbalancer in front of the hosts, then just http[s]://<lb-ip>/<some-path>
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -463,7 +463,7 @@ directory.
 | image.tag                          | Kong image version                                                                    | `2.0`               |
 | image.pullPolicy                   | Image pull policy                                                                     | `IfNotPresent`      |
 | image.pullSecrets                  | Image pull secrets                                                                    | `null`              |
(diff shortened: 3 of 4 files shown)


charts/kong/values.yaml
```diff
@@ -590,7 +596,7 @@ enterprise:
     smtp_admin_emails: none@example.com
     smtp_host: smtp.example.com
     smtp_port: 587
-    smtp_auth_type: ''
+    smtp_auth_type: nil
```
rainest, line 599: If nothing else, this change should be removed. It accidentally reverts https://github.com/Kong/charts/commit/c66fa64dc4944252ff650d042507d08fe24152e9, which we don't want.

rainest:
@arjunkrishnasb can you confirm the rationale for this? Do you have a need to deploy as a DaemonSet for reasons other than handling that particular network configuration?

This seems out of line with what we'd typically expect to support for exposing Kong to clients outside the cluster, and not a standard configuration we'd want to support.

Normally, I'd expect that you either have something that can provide LoadBalancer Services or use a NodePort to have kube-proxy direct inbound traffic to a kubelet hosting a Kong Pod. The latter doesn't provide you with access to reserved ports directly, so if you want that the standard K8S expectation is that you provision a load balancer outside the cluster, have it listen on the standard HTTP/HTTPS ports, and direct traffic to the NodePort. You'd then set [externalIPs](https://kubernetes.io/docs/concepts/services-networking/service/#external-ips) to the load balancer IP to manually provide status information.

I'd recommend [MetalLB](https://metallb.universe.tf/) or similar if your cluster doesn't already have support for LoadBalancer provisioning already.

shaneutt:
@rainest just a heads up, there's relevant context in https://github.com/Kong/kubernetes-ingress-controller/issues/682

arjunkrishnasb:
> @arjunkrishnasb can you confirm the rationale for this? Do you have a need to deploy as a DaemonSet for reasons other than handling that particular network configuration?
> 
> This seems out of line with what we'd typically expect to support for exposing Kong to clients outside the cluster, and not a standard configuration we'd want to support.
> 
> Normally, I'd expect that you either have something that can provide LoadBalancer Services or use a NodePort to have kube-proxy direct inbound traffic to a kubelet hosting a Kong Pod. The latter doesn't provide you with access to reserved ports directly, so if you want that the standard K8S expectation is that you provision a load balancer outside the cluster, have it listen on the standard HTTP/HTTPS ports, and direct traffic to the NodePort. You'd then set [externalIPs](https://kubernetes.io/docs/concepts/services-networking/service/#external-ips) to the load balancer IP to manually provide status information.
> 
> I'd recommend [MetalLB](https://metallb.universe.tf/) or similar if your cluster doesn't already have support for LoadBalancer provisioning already.


That is true. Initially we wanted to ease the deployment as host ports and directly do a TCP mapping to the backend hosts. This makes troubleshooting a bit easier than routing it through nodePort, because that gives introduces kube-proxy at the middle.

Hence i thought we could have daemonset, similar to how nginx offered.

debu99:
when this can be merged?

shaneutt:
> when this can be merged?

It looks like we're still waiting on some review changes and conflict resolution from the original contributor @arjunkrishnasb, once these things are taken care of there are no current blockers so it may be able to merge quickly after that.

Please feel free to share your use case and needs in this matter, end-user context can be helpful to us for priority :vulcan_salute:

charts/kong/values.yaml
```diff
@@ -19,6 +19,12 @@ deployment:
     # Setting this to false with ingressController.enabled=true will create a
     # controller-only release.
     enabled: true
+    # Deploy as daemon set
+  ## Setting this to true will deploy kong as a daemonset
+  ## Else by default it creates as a deployment
+  ##
+  # daemonset: true
+  ##
```
rainest, line 27: Indentation here is a bit wonky because of the `kong` block above. This applies to the entire replication controller, so it should be only one level below `deployment`

Also making the setting always defined with its default explicitly set. Leaving it undefined by default works since that makes it effectively `false`, but I'd generally prefer to state that explicitly:
```suggestion
  # Deploy as a DaemonSet. By default, the chart creates a Deployment.
  # Toggling this setting to true will use a DaemonSet instead
  daemonset: false
```

rainest:
@arjunkrishnasb I'm still a bit wary of this, but there does appear to be interest here and from issues elsewhere that indicate others want to use DaemonSets, so I suppose we should provide the option.

Do you have a bare metal Kong environment where you've set up a DaemonSet successfully, and did it require additional configuration beyond simply changing the replication controller type? https://kubernetes.github.io/ingress-nginx/deploy/baremetal/#via-the-host-network suggests such an environment would need other settings also.

Added another suggestion to the configuration. Can you add that along reverting the [accidental `smtp_auth_type` change](https://github.com/Kong/charts/pull/269/files#r577838875)?

The longer documentation should be moved to a section under https://github.com/Kong/charts/blob/main/charts/kong/README.md#deployment-options rather than an FAQs.md item. The naming of the FAQs doc is a bit off--in practice we use it to cover things that can go wrong and how to fix them, not for general configuration instructions.

arjunkrishnasb:
Hey Guys. Apologies for the late reply. Let me describe my scenario again. In our environment, k8s does not support loadbalancer type services. So we wanted to use an LB in the front and direct them to the worker nodes. If I expose it using node port, it introduces kubeproxy. But if i expose as node port, then i can just directly do netcat to troubleshoot basic connection and it is just easy for us. So we are deploying it as a daemonset.

This is the same setting we have for another ingress controller in our k8s environment. I am not sure what the use case for others. Based on that kindly repriortize it.

rainest (changes requested):
Can you send in an update with the requested changes to where this covered in documentation (under the README's _Deployment Options_ section rather than the FAQs) and the settings in values.yaml? I do want those at least before merging this.

The other part of my last comment was trying to understand whether deploying a DaemonSet alone handles what you needed or if you were aware of other settings that were necessary also--the NGINX controller docs around this indicate that such environments needed some changes to host network settings. It may be that you can control those with existing chart features, but I wasn't sure, hence why I was curious if you'd done anything else along with changing replication controller type when setting up your environment.

rainest:
Handled in #347

PR #278, opened 2021-02-12
chore(*) re-enable the sidecar injection for Kuma

nickolaev:
Kuma supports Jobs since 1.0.7, therefore no need to disable the sidecar injection for it.

  - fixes #

#### Checklist

### DIFF
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -468,7 +468,7 @@ directory.
 | env                                | Additional [Kong configurations](https://getkong.org/docs/latest/configuration/)      |                     |
 | migrations.preUpgrade              | Run "kong migrations up" jobs                                                         | `true`              |
 | migrations.postUpgrade             | Run "kong migrations finish" jobs                                                     | `true`              |
-| migrations.annotations             | Annotations for migration job pods                                                    | `{"sidecar.istio.io/inject": "false", "kuma.io/sidecar-injection": "disabled"}` |
+| migrations.annotations             | Annotations for migration job pods                                                    | `{"sidecar.istio.io/inject": "false" |
 | migrations.jobAnnotations          | Additional annotations for migration jobs                                             | `{}`                |
 | waitImage.repository               | Image used to wait for database to become ready                                       | `bash`              |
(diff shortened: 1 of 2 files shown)


nickolaev:
Putting this as a draft now, as I'd like @svenwal to confirm it first.

svenwal:
As discussed with @nickolaev the limitation removal works as expected so the migration jobs can finish even if Postgres database is part of the Mesh.

The bad news that afterwards the Kong pods will not start as they are getting to initContainers:
1. kuma-init
2. wait-for-db

The second one starts first and is unable to connect to Postgres if it is part of the mesh. kuma-init would start second but will never start.

We've also tested that - as we have been sure Postgres is up&running - removing the wait-for-db initContainer from the deployment manually brings up a fully running Kong (with Kong & Postgres being part of the mesh)

rainest:
@nickolaev re the issue Sven raised do you have a way to make your mutating webhook inject its initContainer first in the list? That seems necessary _in general_ for Kuma to be compatible with any init container that make network requests within the cluster.

nickolaev:
@rainest well unfortunately that is not a problem of init container ordering, it is a much wider conceptual problem. There is a lot written on the topic in other service meshe's issue. Essentially it boils down to:
 1. The dataplane of the Service Mesh is composed out of application proxies (Envoy being the most notorious one). 
 2. When the managed workloads that need to run "in the mesh", are deployed in Kubernetes, the patter that this proxy is deployed alongside the workload is called side-car
 3. Unlike the initContainer (which starts does its job and completes successfully in order for the rest of the Pod to continue starting) the side-car container is not really a Kubernetes core concept. Therefore there is no way for the Pod to "always" run the said side-car "proxy" during its whole lifecycle
 4. as a result when an init container needs to access a network resource within the mesh it can not, as the side-car is not yet there. There is no real way around this as the Pod initialization order mandates that the init containers are executed in order and only after the previous one succeeded. Only after all of them succeeded, the Pod will start its "app containers", side-car proxy being such "app container"

We have had a discussion with @mflendrich about this, and our proposal was to make the "wait-for-db" init container optional. Not sure if there are better options.

rainest:
Included in #285

PR #283, opened 2021-02-18
[kong] remove database waits

rainest:

Remove database readiness/migration completeness init containers entirely. Kong Pods _will_ enter CrashLoopBackoff after initial install, but will come online so long as the database comes online and there are no issues (e.g. invalid DB credentials) that block migrations.

Add configurable restart policies and backoff limits.

This addresses two concerns:
- Kong doesn't provide a good init option for migrations. There's no "simply check database connectivity command", so we use something else (bash or netcat) instead. We don't want this additional dependency.
- initContainers [don't play nicely with mesh sidecars](https://github.com/Kong/charts/pull/278#issuecomment-780439477).

We could make the init behavior more configurable, but after playing around with the options, I don't don't think the potential confusion from the temporary backoff state, primarily during the initial install, warrants the additional config surface, and is sufficiently handled with documentation and better control over the Job lifecycle.

Currently, the chart install sequence is:
1. The main Kong Deployment, initial migrations Job, and Postgres StatefulSet (if enabled) start simultaneously.
2. The init migrations Job includes an init container that performs an basic network connectivity test using bash. Once that succeeds, the Job exits the Init phase, enters Running, and creates its main container, which runs `kong migrations bootstrap`.
3. The main Kong Deployment includes an init container that runs `kong start; kong stop`, which won't succeed/exit Init until bootstrap finishes.

Upgrade Jobs also use the init strategy in (2). They use hooks, so the Deployment will not update until the pre-upgrade Job completes.

With this PR, the sequence is:
1. No change.
2. The main init migrations Job attempts to run `kong migrations bootstrap` immediately. If the database isn't available yet, it will crashloop until the database becomes available or until it exceeds the backoff limit.
3. The main Deployment starts and starts crashing until migrations complete, or crashes indefinitely if they fail to complete.

Upgrade jobs likely shouldn't enter backoff, since the database should already be running when they start.

If the database isn't managed by the chart, lack of a migration init container isn't much concern: database availability is the responsibility of the user, and the migrations job should be able to run immediately. When using the Postgres subchart, database availability is the responsibility of Helm, and there likely will be some delay. In practical testing, it comes online and migrations complete before the hard 6 minute backoff limit or CI wait period kill it.

#### Checklist


mflendrich:
Would addition of a DB connection retry mechanism in core solve this problem without regressing into `CrashLoopBackOff`s?

hbagdi:
-1 on removing these waits. Seeing failed/crashlooping containers provides a bad user experience to our users. This is more of an optics problem than a technical one, where users don't like failures for software that they are installing and getting started with.

> Kong doesn't provide a good init option for migrations. There's no "simply check database connectivity command", so we use something else (bash or netcat) instead. We don't want this additional dependency.

We have the option to add a patch in Kong to support this. I understand that is more involved but that is indeed the right solution. I don't see the urgency of removing these altogether.

> https://github.com/Kong/charts/pull/278#issuecomment-780439477

I see the problem here. Can we offer a knob to remove the wait-for-db init container? Kuma use cases are the minority at the moment and hence the tradeoff we are making here seems wrong.

nickolaev:
@svenwal for visibility

rainest:
> Can we offer a knob to remove the wait-for-db init container?

Thought I had linked the alternative but didn't. We can, it's just more template and config complexity than this option: https://github.com/Kong/charts/compare/next...feat/wait-for-revamp

I agree that the optics issue exists, but I think there's a reasonable case to accept it:

- While not all users will deploy alongside a mesh proxy, we assume that some will, and they'll be stuck with the backoff loop regardless of whether we make this a toggle. The imperfect optics are at least consistent, and don't require config changes to support mesh proxies.
- Installations/initial migrations are ideally infrequent, and we have options to explain odd behavior. Would the bad optics be more acceptable if we announce them more prominently, with a NOTES.txt message?

Another alternative is that we make the current lifecycle a special case for the built-in Postgres subchart only, and declare it incompatible with deployments alongside service meshes. While the subchart is great for quickly spinning up a test or demo environment, I'd recommend separately-managed databases for long-lived Kong deployments regardless.

The worst aspects of the optics go away if we can use a pre-install hook for initial migrations, as the main Kong Deployment won't start until that Job completes, and won't be stuck trying to start before migrations are done. We don't use a hook currently only because the built-in Postgres StatefulSet doesn't spawn until all pre-install hooks complete.

The migration Job init containers aren't as useful since failing entirely if the database isn't available makes sense--arguably moreso than getting stuck in init forever. While the init migrations probably will still restart while spawning alongside a new built-in Postgres, retry restarts are a normal Job feature, and they don't get stuck as long because they start as soon as Postgres comes up, whereas the main Deployment will build up more backoff time because it's waiting on both Postgres and migrations.

rainest:
Alright, bummer, played around with the pre-install option and I don't think it's viable. I forgot that we rely on stuff we create in the install (RBAC resources) other than Postgres, and I'm pretty sure that there's no way to create those in pre-install phase and keep them a managed part of the release without template duplication, which is :vomiting_face:

Checking with the Helm community to see if there's some novel solution to that, but I don't think there is any. Prior solutions are to use an initContainer instead, which would still have the same issue that prompted this originally.

PR #285, opened 2021-02-19
[kong] revamp database wait init containers

rainest:
The main rationale for these changes is similar to [the proposal to remove these waits entirely](https://github.com/Kong/charts/pull/283#issue-575317230). Based on discussion in that PR, we favor this approach instead.

- Allows users to disable database waits for compatibility with service meshes.
- Uses the Kong image for wait-for-postgres init containers by default. This is fine for our stock images, as they include bash. The wait image configuration is now optional, for custom Kong images that don't include bash.
- Disables the wait-for-postgres ConfigMap and mounts if it isn't actually in use.
- Re-enables mesh proxies on Jobs.
- Cleans up some DB documentation.
- Adds a missing labels block to the init migrations job.

Supersedes #283 (similar goal, different approach) and #278 (cherry-picks that commit and rewords the title to follow this repo's format).

#### Checklist

### DIFF
--- a/charts/kong/FAQs.md
+++ b/charts/kong/FAQs.md
@@ -85,3 +85,25 @@ This occurs if a `RELEASE-NAME-kong-init-migrations` Job is left over from a
 previous `helm install` or `helm upgrade`. Deleting it with
 `kubectl delete job RELEASE-NAME-kong-init-migrations` will allow the upgrade
 to proceed. Chart versions greater than 1.5.0 delete the job automatically.
+
+#### DB-backed instances do not start when deployed within a service mesh
+
+Service meshes, such as Istio and Kuma, commonly enforce restrictions on
+traffic between Pods and require that traffic pass through a sidecar proxy.
+These sidecars are not available for init containers.
+
+By default, this chart uses init containers to ensure that the database is
+online and has migrations applied before starting Kong. This provides for a
+smoother startup, but isn't compatible with service mesh sidecar requirements
+if both Kong and its database are in the mesh.
+
+Setting `waitImage.enabled=false` in values.yaml disables these init containers
+and resolves this issue. However, during the initial install, your Kong
+Deployment will enter the CrashLoopBackOff state while waiting for migrations
+to complete. It will eventually exit this state and enter Running as long as
+there are no issues finishing migrations, usually within 2 minutes.
+
+If your Deployment is stuck in CrashLoopBackoff for longer, check the init
+migrations Job logs to see if it is unable to connect to the database or unable
+to complete migrations for some other reason. Resolve any issues you find,
+delete the release, and attempt to install again.
--- a/charts/kong/FAQs.md
+++ b/charts/kong/FAQs.md
@@ -88,14 +88,14 @@ to proceed. Chart versions greater than 1.5.0 delete the job automatically.
 
 #### DB-backed instances do not start when deployed within a service mesh
 
-Service meshes, such as Istio and Kuma, commonly enforce restrictions on
-traffic between Pods and require that traffic pass through a sidecar proxy.
-These sidecars are not available for init containers.
+Service meshes, such as Istio and Kuma, if deployed in a mode that injects
+a sidecar to Kong, don't make the mesh available to `InitContainer`s,
+because the sidecar starts _after_ all `InitContainer`s finish.
 
 By default, this chart uses init containers to ensure that the database is
 online and has migrations applied before starting Kong. This provides for a
 smoother startup, but isn't compatible with service mesh sidecar requirements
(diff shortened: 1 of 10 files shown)


charts/kong/FAQs.md
```diff
@@ -85,3 +85,25 @@ This occurs if a `RELEASE-NAME-kong-init-migrations` Job is left over from a
...
+#### DB-backed instances do not start when deployed within a service mesh
+
+Service meshes, such as Istio and Kuma, commonly enforce restrictions on
+traffic between Pods and require that traffic pass through a sidecar proxy.
+These sidecars are not available for init containers.
+
+By default, this chart uses init containers to ensure that the database is
+online and has migrations applied before starting Kong. This provides for a
+smoother startup, but isn't compatible with service mesh sidecar requirements
+if both Kong and its database are in the mesh.
```
mflendrich:
```suggestion
if Kong is to access the database through the mesh.
```

"Being in the mesh" is fuzzy. "DB being accessed through the mesh" is less so.

charts/kong/templates/_helpers.tpl
```diff
@@ -365,10 +365,12 @@ The name of the service used for the ingress controller's validation webhook
   emptyDir: {}
 - name: {{ template "kong.fullname" . }}-tmp
   emptyDir: {}
+{{- if (and (eq .Values.env.database "postgres") .Values.waitImage.enabled) }}
```
mflendrich: The logic, as I understand it, is about "database being present", not about "database being of type postgres". Therefore, `!= off` rather than `== "postgres"` feels more appropriate to me. Please change the logic to the former unless there's a compelling reason not to do so.

(applies to other similar checks "if database is postgres" in this PR as well)

charts/kong/templates/_helpers.tpl
```diff
@@ -761,8 +763,14 @@ Environment variables are sorted alphabetically
...
   image: "{{ .Values.waitImage.unifiedRepoTag }}"
-{{- else }}
+{{- else if .Values.waitImage.repository }}
   image: "{{ .Values.waitImage.repository }}:{{ .Values.waitImage.tag }}"
+{{- else }} {{/* default to the Kong image */}}
+{{- if .Values.image.unifiedRepoTag }}
+  image: "{{ .Values.image.unifiedRepoTag }}"
+{{- else }}
+  image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+{{- end }}
```
mflendrich: This `if` ladder seems to be adding some duplicated and rather complex logic. Could this be avoided by something similar to the below:

- implementing a helper template (called e.g. `getRepoTag`) that takes an "image" object as scope, and returns either `.unifiedRepoTag` or a `{{.repository}}:{{.tag}}`
- using the `default` function, something like this: `image: {{ default (include "getRepoTag" .Values.Image) (include "getRepoTag" .Values.Image) }}`

?

charts/kong/templates/_helpers.tpl
rainest: There's a reason it's like this, though the check is a bit off. These waits, for the initial migrations, are only useful if Postgres being online or not is influenced by the chart lifecycle. This is only true if you use the sub-chart and Postgres spawns alongside the migration Jobs.

If you manage Postgres or Cassandra independently (all Cassandra instances are managed independently--we have no sub-chart for it), the chart lifecycle has no influence on whether the database is available or not. It either is, and migrations can start immediately, or it isn't, and they cannot. No amount of waiting helps if an independently-managed database isn't online; it makes sense to just let the failure happen immediately so users know they need to take action to make the database reachable. e5012beea782a95686f9256f65ba373dda966a74 is more correct for these, since it changes the check from "any Postgres" to "only sub-chart Postgres".

The Deployment waits, however, are to see if migrations have completed. Those apply to any database and are always determined by the chart lifecycle. I did, however, copy the wrong logic when adding the `waitImage.enabled` for them :facepalm: e4b534ee0da6fa8419b2e950031ee95b22630bf1 fixes that.

charts/kong/templates/_helpers.tpl
rainest: Yeah, good idea. This gets repeated in several locations and doesn't need to be. The tests for it were also incorrect. 3f64601

mflendrich (approved):
Looks good!

Before merging, please fix the error in the PR description (_labels_ -> _annotations_).

nickolaev:
Thank you to everyone involved in resolving this, seemingly small, yet blocking issue.

hbagdi:
@rainest Thanks for working through this multiple times!
Looks good from a user experience perspective.

Issue #272, opened 2021-02-09
ServiceMonitor does not expose status port if proxy service is not enabled


slawekww:
Let's assume that kong helm chart is used to deploy Control Plane in hybrid mode without proxy service enabled.

Example helm values:

```yaml
deployment:
  kong:
    enabled: true

admin:
  enabled: true
cluster:
  enabled: true
proxy:
  enabled: false
ingressController:
  enabled: false
manager:
  enabled: true
portal:
  enabled: true
portalapi:
  enabled: true
clustertelemetry:
  enabled: true

serviceMonitor:
  enabled: true
```

It caused that status port is not exposed via kong k8s service, not labeled "enable-metrics: true"
and ServiceMonitor registered target in Prometheus but no endpoints are available.

Helm chart version: 1.14.3

rainest:
This is set up this way since the ServiceMonitor differentiates between the service being monitored and where its metrics are available--they often aren't the same service. There are some other historical reasons why this is configured this way:

1. The ServiceMonitor predates the creation of hybrid mode.
2. The Prometheus plugin used to only provide metrics generated by the proxy: if you had no proxy enabled, you would not have any metrics. It added memory statistics (similar to what's exposed at `/status`) at a later point.
3. IIRC labeling multiple Services with the metrics label made the ServiceMonitor collect those metrics twice.
4. The status endpoint has no Service to label. This is intentional, since it's providing per-instance information that should be monitored at the individual Pod level, whereas a Service will send a request to a single ready Pod providing the Service.

The best solution here is probably to add a user-configurable labels section that places the metrics label on the proxy by default, since we can't easily account for all configurations automatically--users may also configure Portal-only or Manager-only nodes and want to collect memory metrics for them.

slawekww:
Let's focus on the following facts:
* services are served by processes run in pod replicas
* regardless what service is managed via pod, /status endpoint is used to check internal status of pod (livenessProbecheck)
* serviceMontor is used to expose custom pod metrics to Prometheus (basic metrics are managed by kube-api metrics)

@rainest 
If custom pod metrics for non-proxy service are valid and useful, it should be way to expose them via ServiceMonitor. Please kindly confirm if it is valid statement.

rainest:
It should be possible, yes--I'd just wanted to lay out the background for future reference.

[...]

PR #290, opened 2021-02-22
[kong] add user-configurable Service labels

rainest:
Adds configurable Service labels with a default proxy Service label.

Adds documentation for the Prometheus Operator integration.

Indirectly fixes https://github.com/Kong/charts/issues/272. Currently the ServiceMonitor does nothing if the proxy Service is disabled, as we hard-coded the monitor label on it and did not have any means for users to add labels themselves.

#### Checklist

### DIFF
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -53,6 +53,7 @@ $ helm install kong/kong --generate-name --set ingressController.installCRDs=fal
   - [RBAC](#rbac)
   - [Sessions](#sessions)
   - [Email/SMTP](#emailsmtp)
+- [Prometheus Operator integration](#prometheus-operator-integration)
 - [Changelog](https://github.com/Kong/charts/blob/main/charts/kong/CHANGELOG.md)
 - [Upgrading](https://github.com/Kong/charts/blob/main/charts/kong/UPGRADE.md)
 - [Seeking help](#seeking-help)
@@ -536,6 +537,7 @@ nodes.
 | SVC.ingress.path                   | Ingress path.                                                                         | `/`                 |
 | SVC.ingress.annotations            | Ingress annotations. See documentation for your ingress controller for details        | `{}`                |
 | SVC.annotations                    | Service annotations                                                                   | `{}`                |
+| SVC.annotations                    | Service labels                                                                        | `{}`                |
 
 #### Stream listens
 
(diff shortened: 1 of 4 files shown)


charts/kong/README.md
```diff
@@ -536,6 +537,7 @@ nodes.
 | SVC.ingress.path                   | Ingress path.                                                                         | `/`                 |
 | SVC.ingress.annotations            | Ingress annotations. See documentation for your ingress controller for details        | `{}`                |
 | SVC.annotations                    | Service annotations                                                                   | `{}`                |
+| SVC.annotations                    | Service labels                                                                        | `{}`                |
```
```suggestion
| SVC.labels                         | Service labels                                                                        | `{}`                |
```

mflendrich: The descriptions "service annotations" and "service labels" aren't very helpful because they essentially repeat the value name, and because the "service"/"SVC" term is overloaded here. Please say what effects seetting something there has.

An example of what I mean: "Annotations of the Kubernetes `Service` created for `SVC`"

charts/kong/README.md
```diff
@@ -805,6 +807,30 @@ If your SMTP server requires authentication, you must provide the `username` and
 
 By default, SMTP uses `AUTH` `PLAIN` when you provide credentials. If your provider requires `AUTH LOGIN`, set `smtp_auth_type: login`.
 
+## Prometheus Operator integration
+
+The chart can configure a ServiceMonitor resource to instruct the [Prometheus
+Operator](https://github.com/prometheus-operator/prometheus-operator) to
+collect metrics from Kong Pods. To enable this, set
```
mflendrich, line 814:
```suggestion
collect metrics from Kong pods. To enable this, set
```

nit - optional

charts/kong/README.md
```diff
@@ -805,6 +807,30 @@ If your SMTP server requires authentication, you must provide the `username` and
...
+`serviceMonitor.enabled=true` in values.yaml.
```
```suggestion
`serviceMonitor.enabled=true` in `values.yaml`.
```

charts/kong/values.yaml
```diff
@@ -78,10 +78,11 @@ admin:
   # Enterprise users that wish to use Kong Manager with the controller should enable this
   enabled: false
   type: NodePort
-  # If you want to specify annotations for the admin service, uncomment the following
-  # line, add additional or adjust as needed, and remove the curly braces after 'annotations:'.
+  # To specify annotations or labels for the admin service, remove the "{}" and add indented lines
+  # after, similar to the commented aws-load-balancer-proxy-protocol example
```
mflendrich:
```suggestion
  # To specify annotations or labels for the admin service, add them to the respective
  # "annotations" or "labels" dictionaries below.
```

Instead of asking users to manipulate opaque strings, we should be speaking about YAML concepts, right?

Applies to all repetitions.

charts/kong/README.md
rainest: The guidance in https://kubernetes.io/docs/contribute/style/style-guide/#use-upper-camel-case-for-api-objects is unfortunately less than clear re the "specifically to interacting with an API object" and "generally discussing an API object" distinction, and the examples don't clarify it much--one uses lowercase "pods" but others use camel case without an interaction. IME their docs tend to camel case API objects more often than not, so I generally use that throughout. Left as-is.

charts/kong/README.md
rainest: We explain the use of SVC above: https://github.com/Kong/charts/blame/kong-1.14.5/charts/kong/README.md#L482-L496

The description is brief, but that's true of most of these--a lot of them are just standard K8S Service features that don't need additional explanation specific to our chart. Leaving these as-is.

mflendrich (approved):

PR #292, opened 2021-02-23
[kong] release 1.15.0 to next

### LINKED ISSUE #291: [kong] Bump Kong to 2.3

rainest:
Bumps Kong image versions and releases 1.15.0 to next.

  - fixes #291 

This is blocked on:
- Approval and merger of #290.
- Review of the compatibility concerns from #291 

The content of this PR is not expected to change if those are complete and we don't see a reason to hold off on the 2.3 bump.

#### Checklist

### DIFF
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -690,11 +690,10 @@ Kong is going to be deployed.
 
 #### Kong Enterprise Docker registry access
 
-Next, we need to setup Docker credentials in order to allow Kubernetes
-nodes to pull down Kong Enterprise Docker images, which are hosted in a private
-registry.
+Kong Enterprise versions 2.2 and earlier use a private Docker registry and
+require a pull secret. If you use 2.3 or newer, you can skip this step.
 
-You should received credentials to log into https://bintray.com/kong after
+You should have received credentials to log into https://bintray.com/kong after
 purchasing Kong Enterprise. After logging in, you can retrieve your API key
 from \<your username\> \> Edit Profile \> API Key. Use this to create registry
 secrets:
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -691,7 +691,7 @@ Kong is going to be deployed.
 #### Kong Enterprise Docker registry access
 
 Kong Enterprise versions 2.2 and earlier use a private Docker registry and
-require a pull secret. If you use 2.3 or newer, you can skip this step.
+require a pull secret. **If you use 2.3 or newer, you can skip this step.**
 
 You should have received credentials to log into https://bintray.com/kong after
 purchasing Kong Enterprise. After logging in, you can retrieve your API key
--- a/charts/kong/ci/test5-values.yaml
+++ b/charts/kong/ci/test5-values.yaml
@@ -1,8 +1,11 @@
 # This tests the following unrelated aspects of Ingress Controller
 # - ingressController deploys with a database
-# - TODO database-backed proxy works _without_ stream listens
-#   This test is a duplicate of test2-values.yaml to handle the https://github.com/Kong/charts/issues/295
-#   workaround when the socket file is _not_ created
+# - TODO remove this test when https://github.com/Kong/charts/issues/295 is solved
+#   and its associated wait-for-db workaround is removed.
+#   This test is similar to test2-values.yaml, but lacks a stream listen.
+#   wait-for-db will _not_ create a socket file. This test ensures the workaround
+#   does not interfere with startup when there is no file to remove.
+
 ingressController:
   enabled: true
   installCRDs: false
--- a/charts/kong/ci/test5-values.yaml
+++ b/charts/kong/ci/test5-values.yaml
@@ -1,6 +1,8 @@
 # This tests the following unrelated aspects of Ingress Controller
 # - ingressController deploys with a database
-# - stream listens work
+# - TODO database-backed proxy works _without_ stream listens
+#   This test is a duplicate of test2-values.yaml to handle the https://github.com/Kong/charts/issues/295
+#   workaround when the socket file is _not_ created
 ingressController:
   enabled: true
   installCRDs: false
@@ -29,15 +31,6 @@ proxy:
     hostname: proxy.kong.example
     annotations: {}
     path: /
-# - add stream listens
-  stream:
-  - containerPort: 9000
-    servicePort: 9000
-    parameters: []
-  - containerPort: 9001
-    servicePort: 9001
-    parameters:
-    - ssl
 
 # - PDB is enabled
 podDisruptionBudget:
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -492,12 +492,15 @@ The name of the service used for the ingress controller's validation webhook
 {{- end -}}
 
 {{- define "kong.wait-for-db" -}}
+{{ $sockFile := (printf "%s/stream_rpc.sock" (default "/usr/local/kong" .Values.env.prefix)) }}
 - name: wait-for-db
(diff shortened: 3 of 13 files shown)


shaneutt (approved):

rainest:
b4d21f3 is an alternative workaround that just removes the problem socket file. After [further review of why this occurs](https://github.com/Kong/kong/issues/6827#issuecomment-785318497), we may just be stuck with this for a while, since the ideal fix requires NGINX-level code changes.

Still not entirely sure why the command's `kong stop`/SIGTERM isn't actually removing the socket file before the Docker-level SIGQUIT fires, but that's my best guess as to what happens. Simply removing the file is the historical workaround for the upstream issue, so we may as well go with that until we get an updated NGINX version.

charts/kong/templates/_helpers.tpl
```diff
@@ -494,7 +494,9 @@ The name of the service used for the ingress controller's validation webhook
   imagePullPolicy: {{ .Values.image.pullPolicy }}
   env:
   {{- include "kong.env" . | nindent 2 }}
-  command: [ "/bin/sh", "-c", "until kong start; do echo 'waiting for db'; sleep 1; done; kong stop" ]
+{{/* TODO: the rm command here is a workaround for https://github.com/Kong/kong/issues/6827
+     It should be removed once that's fixed */}}
```
mflendrich: Please file a bug and link to it from the TODO.

charts/kong/templates/_helpers.tpl
```diff
@@ -494,7 +494,9 @@ The name of the service used for the ingress controller's validation webhook
...
+  command: [ "/bin/sh", "-c", "until kong start; do echo 'waiting for db'; sleep 1; done; kong stop; rm {{ .Values.env.prefix | default "/usr/local/kong" }}/stream_rpc.sock" ]
```
mflendrich: If the `rm`ed file doesn't exist, `sh -c` will exit with non-zero status and therefore mark the initcontainer as failed. Or am I missing something?

mflendrich (changes requested):

charts/kong/templates/_helpers.tpl
rainest: Indeed it does not. Good catch--the only CI check that used the DB also included stream listens to try and economize on CI runs. That made sense since stream listen functionality _should_ be independent of whether the instance is DB-backed or not, but that's no longer the case.

rainest:
4f0891b left separate for now to show the diff more easily. We should rebase on top of next after #290 is in and squash it into  b4d21f3.

The hell. test2 isn't deleting the socket file anymore in CI, though it is when I run with that set of values manually.

rainest:
Went ahead and rebased to next since several changes have since gone in there. The post-review diff is: [292-review-diff.txt](https://github.com/Kong/charts/files/6045019/292-review-diff.txt)

This check originally used a shell-native file check:

```
command: [ "/bin/sh", "-c", "until kong start; do echo 'waiting for db'; sleep 1; done; kong stop; if [ -f `realpath {{ $sockFile }}` ]; then rm {{ $sockFile }}; echo '{{ $sockFile }} removed'; else echo '{{ $sockFile }} noop'; ls {{ .Values.env.prefix }}; fi"]
```
For some reason, this doesn't work in CI, though it did work for me locally and in a test cluster install. The stat variant works everywhere.

charts/kong/templates/_helpers.tpl
```diff
@@ -492,12 +492,15 @@ The name of the service used for the ingress controller's validation webhook
...
+{{ $sockFile := (printf "%s/stream_rpc.sock" (default "/usr/local/kong" .Values.env.prefix)) }}
 - name: wait-for-db
   image: {{ include "kong.getRepoTag" .Values.image }}
   imagePullPolicy: {{ .Values.image.pullPolicy }}
   env:
   {{- include "kong.env" . | nindent 2 }}
-  command: [ "/bin/sh", "-c", "until kong start; do echo 'waiting for db'; sleep 1; done; kong stop" ]
+{{/* TODO: the rm command here is a workaround for https://github.com/Kong/charts/issues/295
+     It should be removed once that's fixed */}}
+  command: [ "/bin/sh", "-c", "until kong start; do echo 'waiting for db'; sleep 1; done; kong stop; if stat {{ $sockFile }} > /dev/null; then rm {{ $sockFile }}; else true; fi"]
```
mflendrich:
```suggestion
  command: [ "/bin/sh", "-c", "until kong start; do echo 'waiting for db'; sleep 1; done; kong stop; rm -fv {{ $sockFile | squote }}"]
```

`-f` makes `rm` _ignore nonexistent files and arguments_ and _never prompt_. Both seem appropriate in this case. Also without the `if` the code is less verbose.
Also quoted the sockFile. Without quotes, if the filename contains a space character, it will be expanded into multiple arguments for `rm`.

charts/kong/README.md
```diff
@@ -690,11 +690,10 @@ Kong is going to be deployed.
 
 #### Kong Enterprise Docker registry access
 
-Next, we need to setup Docker credentials in order to allow Kubernetes
-nodes to pull down Kong Enterprise Docker images, which are hosted in a private
-registry.
+Kong Enterprise versions 2.2 and earlier use a private Docker registry and
+require a pull secret. If you use 2.3 or newer, you can skip this step.
```
```suggestion
Kong Enterprise versions 2.2 and earlier use a private Docker registry and
require a pull secret. *If you use 2.3 or newer, you can skip this step*.
```

nit

charts/kong/ci/test5-values.yaml
```diff
@@ -0,0 +1,43 @@
+# This tests the following unrelated aspects of Ingress Controller
+# - ingressController deploys with a database
+# - TODO database-backed proxy works _without_ stream listens
```
mflendrich: What's the deal with the TODO here?

charts/kong/ci/test5-values.yaml
```diff
@@ -0,0 +1,43 @@
...
+#   This test is a duplicate of test2-values.yaml to handle the https://github.com/Kong/charts/issues/295
+#   workaround when the socket file is _not_ created
```
mflendrich: I don't understand this comment.
Does `test5-values.yaml` (identical to `test2-values.yaml`) verify that `test2` has not leaked the socket by retrying the same operations and failing if the socket already exists?

If so:
```suggestion
#   This test is a duplicate of test2-values.yaml to handle the https://github.com/Kong/charts/issues/295
#   in order to check if test2 hasn't leaked a socket file. If it did, this test
#   will fail because the file already exists.
```

mflendrich (commented):
LGTM pending fixes

charts/kong/ci/test5-values.yaml
rainest: Rephrased to clarify the TODO condition and purpose of the test.

This new case was more necessary when the workaround used a conditional test, before your suggestion to use `rm -f`. With the prior iteration, I wanted to test both workaround cases, i.e.
- A DB-backed configuration with a stream listen creates a socket file in wait-for-db. The workaround removes it, wait-for-db exits successfully, and the main Kong container starts correctly. test2 already confirmed this.
- A DB-backed configuration with _no_ stream listen creates no socket files in wait-for-db. The workaround does not fail if no file is present, wait-for-db exits successfully, and the main Kong container starts correctly. No existing test case covered this; test5 now does.

test2 and test5 don't interact with one another, but we can't simultaneously test the workaround behavior for configurations that create the file and those that do not, and no existing test covered the latter.

The workaround is arguably safer with the `rm -f` variant (we rely on known behavior of `rm` rather than our own script), so we may be okay to remove it entirely--I'm ambivalent, but keeping it errs on the side of caution. Comment is hopefully clearer now though.

mflendrich (approved):

PR #296, opened 2021-02-25
Update README.md

svenwal:
Added a note on README that pull secrets are not needed anymore since Kong Enterprise 2.3

### DIFF
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -684,7 +684,7 @@ Set the secret name in `values.yaml`, in the `.enterprise.license_secret` key.
 Please ensure the above secret is created in the same namespace in which
 Kong is going to be deployed.
 
-#### Kong Enterprise Docker registry access
+#### Kong Enterprise Docker registry access (only for Kong <2.3)
 
 :bulb: This is needed only for Kong Enterprise versions older than 2.3 - from 2.3 onwards, the image can be downloaded without those pullSecrets.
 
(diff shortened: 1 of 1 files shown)


charts/kong/README.md
```diff
@@ -686,6 +686,8 @@ Kong is going to be deployed.
 
 #### Kong Enterprise Docker registry access
```
```suggestion
#### Kong Enterprise Docker registry access (only for Kong <2.3)
```

charts/kong/README.md
```diff
@@ -686,6 +686,8 @@ Kong is going to be deployed.
...
 
+:bulb: This is needed only for Kong Enterprise version <=2.2 - since 2.3 the image can be downloaded without those pullSecrets.
```
```suggestion
:bulb: This is needed only for Kong Enterprise versions older than 2.3 - from 2.3 onwards, the image can be downloaded without those pullSecrets.
```

This is because, technically, "2.2.0.1" is not "<= 2.2". My understanding is that for any 2.2.x.y these steps must be followed - please correct me if I'm wrong.

charts/kong/README.md
svenwal: You English is better than mine  :)

mflendrich: Probably not :slightly_smiling_face:  - rewording comments are just about leaving no doubt about the corner cases.

This particular comment makes it evident from the header of the section that for modern Kong versions it's ok to gloss over it.

mflendrich (approved):
Thank you for improving the docs here! :rocket:

PR #300, opened 2021-02-26
[kong] release 1.15.0

### LINKED ISSUE #264: Implement requests/limits for all containers
Hi there!

Kong chart have the possibility to define resource quotas for the main pods but don't have an option to inheritance this resource for the migration job and neither for the batch init container kong.wait-for-postgres. 

If we have a namespace with resourceQuotas, as example:

```
---
apiVersion: v1
kind: ResourceQuota
metadata:
  name: kong-rq-test
  namespace: kong
spec:
  hard:
    requests.cpu: '1'
    requests.memory: '2Gi'
    limits.cpu: '2'
    limits.memory: '4Gi'
```

When we try to install Kong in this environment, Kubernetes will be stuck int the migration-job:

`failed quota: kong-rq-test: must specify limits cpu,limits.memory,requests.cpu,requests.memory`

If  we inheritance the resources from the main container, this migrations will past but then we'll be stuck in the init continer.

Options:

1. Define limit range for the namespace, in this way all pod without resources defined will be use this default. Problems, what if the limit range defined is to minimal and the migrations job and batch pod cant run.
2. Inheritance the resource for migrations and define resources for batch.

Do you think that this issue needs to be resolved by the chart or is an administrator issue?

### LINKED ISSUE #272: ServiceMonitor does not expose status port if proxy service is not enabled

### LINKED ISSUE #291: [kong] Bump Kong to 2.3

rainest:
Releases 1.15.0

  - fixes #272 
  - fixes #264 
  - fixes #291 

#### Checklist

### DIFF
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -471,8 +471,8 @@ directory.
 | migrations.annotations             | Annotations for migration job pods                                                    | `{"sidecar.istio.io/inject": "false", "kuma.io/sidecar-injection": "disabled"}` |`
 | migrations.jobAnnotations          | Additional annotations for migration jobs                                             | `{}`                |
 | waitImage.enabled                  | Spawn init containers that wait for the database before starting Kong                 | `false              |
-| waitImage.repository               | Image used to wait for database to become ready                                       | `bash`              |
-| waitImage.tag                      | Tag for image used to wait for database to become ready                               | `5`                 |
+| waitImage.repository               | Image used to wait for database to become ready. Uses the Kong image if none set      |                     |
+| waitImage.tag                      | Tag for image used to wait for database to become ready                               |                     |
 | waitImage.pullPolicy               | Wait image pull policy                                                                | `IfNotPresent`      |
 | postgresql.enabled                 | Spin up a new postgres instance for Kong                                              | `false`             |
 | dblessConfig.configMap             | Name of an existing ConfigMap containing the `kong.yml` file. This must have the key `kong.yml`.| `` |
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -468,7 +468,7 @@ directory.
 | env                                | Additional [Kong configurations](https://getkong.org/docs/latest/configuration/)      |                     |
 | migrations.preUpgrade              | Run "kong migrations up" jobs                                                         | `true`              |
 | migrations.postUpgrade             | Run "kong migrations finish" jobs                                                     | `true`              |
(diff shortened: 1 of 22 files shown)


charts/kong/README.md
```diff
@@ -591,6 +591,7 @@ For a complete list of all configuration values you can set in the
 | readinessProbe                     | Kong readiness probe                                                                  |                     |
 | livenessProbe                      | Kong liveness probe                                                                   |                     |
 | lifecycle                          | Proxy container lifecycle hooks                                                       | see `values.yaml`   |
+| terminationGracePeriodSeconds      | Related to lifecycle hook                                                             | 30                  |
```
mflendrich, line 597:
```suggestion
| terminationGracePeriodSeconds      | Sets the [termination grace period](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#hook-handler-execution) for Deployment pods | 30                  |
```

charts/kong/values.yaml
```diff
@@ -456,8 +456,13 @@ livenessProbe:
 lifecycle:
   preStop:
     exec:
+      # Note kong quit has a default timeout of 10 seconds
       command: ["/bin/sh", "-c", "/bin/sleep 15 && kong quit"]
 
+# terminationGracePeriodSeconds is closely related to the lifecycle preStop hook
```
mflendrich, line 486:
```suggestion
# Sets the termination grace period for pods spawned by the Kubernetes Deployment.
```

mflendrich (approved):
non-blocking: present documentation for terminationGracePeriodSeconds does not appear useful to me, specifically:
- it does not say on which pods it specifies the grace period
- it does not explain what the grace period is
The suggestion tries to fix that.

Also, the PRs to `next` being fast-forward-merged/rebase-and-merged (instead of squashed, or merged with a merge commit) took me some time to figure out, because they lose the PR context (well, GH is able to recover that context by connecting these commits to PRs).

For me, the history would be more readable with either merge commits (if the sequence of commits within the PR matters) or squash-and-merges referencing PR number otherwise.

PR #302, opened 2021-03-01
[kong] clarify terminationGracePeriodSeconds docs

rainest:
Doc changes suggested in #300.

Pulled out separately since I didn't want to add new stuff in the merge from next to main in #300 

#### Checklist


mflendrich (changes requested):
I don't think we should be doing pushes to `main` which are not version releases because this is at odds with the release process, and has a chance to produce surprising results (two different states of the `kong` chart having the same version number - which one should the chart zip include? Should the zip for version 1.15.0 change over time?)

Therefore :-1:  - I'd rather not see pushing this change to `main` without a version bump, and instead recommend merging to `next`.

Also,

> [x] PR is based off the current tip of the next branch and targets next, not main

seems inaccurate.

rainest:
Target was wrong, should have been next.

mflendrich (approved):

Issue #313, opened 2021-03-11
Invalid template syntax 


ttony:
When you are using helm sdk and error:

[ERROR] Have some error in upgrading helm chart : parse error at (igw-cp/charts/kongHybrid/charts/kong/templates/service-kong-proxy.yaml:13): "-" 

service-kong-proxy.yaml:

```javascript
{{ if (not (hasKey .Values.proxy.ingress "hosts"))  -}}
```
so, to fix it temporary. i remove the "-" at the end as:

```javascript
{{ if (not (hasKey .Values.proxy.ingress "hosts"))  }}
```

rainest:
Can you provide more details or an example program we can replicate this? To confirm, are you rendering the chart using [the Go SDK](https://helm.sh/docs/topics/advanced/#go-sdk) instead of the CLI helm tool? Does this issue happen with the current next branch of the chart as well?

This particular section of the template is for handling legacy proxy ingress configuration. We do test both variants currently, via 
https://github.com/Kong/charts/blob/kong-1.15.0/charts/kong/ci/test4-values.yaml#L53-L57 (for legacy) and https://github.com/Kong/charts/blob/main/charts/kong/ci/test1-values.yaml#L39-L43 (non-legacy).

When I run `helm template`, it doesn't complain with either, but I believe it performs some additional post-processing on the output, since the resulting Ingress and Service definitions get separated. It's possible that doesn't happen when using the SDK, though it's not clear why that newline chomp would cause an issue.

The legacy handling will go away entirely in the next release of the chart (which is currently staged in next). I'd want to target any fix for that instead if the issue is still present.

ttony:
Hi Rainest, 

we can close this for time being as i can see the code has been lot of changes. let's close this.

Issue #270, opened 2021-02-03
Chart test verifies the TLS-secured ingress scenario


rainest:
See https://github.com/Kong/charts/pull/268 for an example uncaught issue.

We cannot currently test templates that require a Secret, namely:
- Ingress TLS configuration
- ~Enterprise instances, which need a registry pull secret and a license~ out of scope

## Proposed solution

https://github.com/Azure/k8s-create-secret should work for this. For certificates we can just use a garbage certificate. Enterprise configuration would require actual valid credentials.

## Acceptance criteria
- [ ] The "TLS-secured ingress endpoint" use case is tested by chart tests

sandromodarelli:
closing as stale

PR #317, opened 2021-03-18
[kong] add user defined volumemount,volume and initContainers

khaikong:
Ability to add additional init containers before initialize kong pods. This will give ability to load custom plugins source file without using needing creation of custom image. It also give flexibility to add extra initcontainers before running kong pods.
It also add ability to add custom volumes and volumemounts. To be used with pods

  - fixes #

#### Checklist


khaikong:
lint test failed here..
When I ran `helm lint` this is what I see
```
[kong] helm lint                                                                                feature/scb_initcontainers  ✭
==> Linting .
[ERROR] Chart.yaml: appVersion should be of type string but it's of type float64
[WARNING] templates/controller-rbac-resources.yaml: rbac.authorization.k8s.io/v1beta1 Role is deprecated in v1.17+, unavailable in v1.22+; use rbac.authorization.k8s.io/v1 Role
[WARNING] templates/controller-rbac-resources.yaml: rbac.authorization.k8s.io/v1beta1 RoleBinding is deprecated in v1.17+, unavailable in v1.22+; use rbac.authorization.k8s.io/v1 RoleBinding
[WARNING] templates/controller-rbac-resources.yaml: rbac.authorization.k8s.io/v1beta1 ClusterRole is deprecated in v1.17+, unavailable in v1.22+; use rbac.authorization.k8s.io/v1 ClusterRole
[WARNING] templates/controller-rbac-resources.yaml: rbac.authorization.k8s.io/v1beta1 ClusterRoleBinding is deprecated in v1.17+, unavailable in v1.22+; use rbac.authorization.k8s.io/v1 ClusterRoleBinding

Error: 1 chart(s) linted, 1 chart(s) failed
```
and suggestion, should I ignore the `WARNING`

ttony:
@khaikong  thank you for the help. Look forward to the PR.

charts/kong/values.yaml
```diff
@@ -24,6 +24,15 @@ deployment:
   # sidecarContainers:
   #   - name: sidecar
   #     image: sidecar:latest
+  # initContainers:
+  # - name: initcon
+  #   image: initcon:latest
+  # userDefinedVolumes:
```
```suggestion
  # userDefinedVolumes:
```

charts/kong/templates/_helpers.tpl
```diff
@@ -355,6 +355,12 @@ The name of the service used for the ingress controller's validation webhook
 
 {{- end -}}
 
+{{- define "kong.userDefinedVolumes" -}}
+  {{- if .Values.deployment.userDefinedVolumes }}
+  {{- toYaml .Values.deployment.userDefinedVolumes }}
+  {{- end }}
```
rainest: Indentation here also looks wrong--IIRC `toYaml` doesn't play nicely with in-line indents and will indent the first line only.

These should probably all have no indent, similar to what we have for the other volume helpers, e.g. https://github.com/Kong/charts/blob/kong-1.15.0/charts/kong/templates/_helpers.tpl#L366-L370. Not sure, but CI should confirm--if it's wrong, the result will be a rendering error.

rainest (changes requested):
Several additional change requests that didn't fit well into in-line comments:

- The README updates should include these new settings under the settings list also, around https://github.com/Kong/charts/blame/kong-1.15.0/charts/kong/README.md#L587
- Please add a user-defined initContainer to [one of the CI cases](https://github.com/Kong/charts/tree/next/charts/kong/ci). test3 should be fine (we need to reorganize them, but that's outside the scope of this PR, so I've chosen that one arbitrarily). We're only looking to test rendering really, so something that should be guaranteed to work (e.g. a `bash:latest` container that runs `true`) should suffice.
- Ditto for the user-defined volumes/mounts; those should be tested via CI also. Although https://github.com/Kong/charts/issues/270 is still unsolved, you should be able to create an `emptyDir` volume to avoid that.

The linter error is just because of the extra space in values.yaml. You're seeing some additional errors when running locally because you're using a newer version of Helm, though for some reason it's _not_ complaining about the extra space. You can ignore the additional errors, since we're not bumping the version used in CI for a bit for unrelated reasons.

charts/kong/templates/service-kong-proxy.yaml
```diff
@@ -10,6 +10,7 @@
 {{- include "kong.service" $serviceConfig }}
 {{ if .Values.proxy.ingress.enabled }}
 ---
+{{ if (not (hasKey .Values.proxy.ingress "hosts"))  }}
```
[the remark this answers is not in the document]
rainest: Ah, okay, this was added because of https://github.com/Kong/charts/issues/313#issuecomment-802245433

Regardless, that line should no longer be here--it's gone in next, so if SDK usage still breaks with the current next template (https://github.com/Kong/charts/blob/887c5d29e37081e1b33e400750a2b0bb4754304f/charts/kong/templates/service-kong-proxy.yaml#L12-L13), any fix should be targeted to that, not the now-removed legacy handler.

charts/kong/templates/_helpers.tpl
khaikong: fixed

ttony:
@khaikong  thanks for the fix.

rainest:
@khaikong can you revert 522b157? That added a different variant of the old content, and  4b90c91 was the fix I actually wanted there.

@ttony are you able to [rebase and reauthor](https://www.git-tower.com/learn/git/faq/change-author-name-email/) your commits to place them under the email associated with your account? The CLA check won't recognize them otherwise.

Deleted the example since we want those to just focus on different Kong environment types (major things like hybrid mode, ingress controller use, etc.). For features that can be used in any deployment topology like these, the examples in the base values.yaml suffice, and are easier to maintain (since there's only the one file to update).

Other than that, we should be all set.

ttony:
@khaikong,  I have rebase the commits from @rainest  suggestion. Can you pls help me to pull my changes to this PR? thenk you

khaikong:
Seems like everything is good to go. @rainest please one last check before we merge

rainest (approved):

PR #323, opened 2021-03-25
[kong] look up the CRD version that's available in the cluster

### LINKED ISSUE #322: Chart broken on Kubernetes <1.16 due to lookup using apiextensions.k8s.io/v1 CRD
Hi.

When using the chart with a cluster in version 1.15, the helm install fails with this error message:
```
lookup_func.go:111: [ERROR] unable to retrieve resource list for: apiextensions.k8s.io/v1 , error: the server could not find the requested resource
lookup_func.go:85: [ERROR] unable to get apiresource from unstructured: apiextensions.k8s.io/v1, Kind=CustomResourceDefinition , error the server could not find the requested resource
Error: template: kong/templates/custom-resource-definitions.yaml:21:25: executing "kong/templates/custom-resource-definitions.yaml" at <lookup "apiextensions.k8s.io/v1" "CustomResourceDefinition" "" "kongplugins.configuration.konghq.com">: error calling lookup: unable to get apiresource from unstructured: apiextensions.k8s.io/v1, Kind=CustomResourceDefinition: the server could not find the requested resource
```
This is using Helm 3.3.4, the last version of Helm compatible with the cluster.

The cause of the error is here https://github.com/Kong/charts/blob/kong-2.0.0-rc.1/charts/kong/templates/custom-resource-definitions.yaml#L21, directly using the `apiextensions.k8s.io/v1` API without first checking its availability.

The chart advertises compatibility with Kube 1.12+ so it looks like a bug :).

rainest:
The original implementation of this lookup in #305 assumed that the CRD API version was `apiextensions.k8s.io/v1`. This is the case on 1.16+, but earlier versions still use `apiextensions.k8s.io/v1beta1`, and the attempt to look up the newer version failed.

  - fixes #322 

Comparison of behavior between K8S versions in my test KIND environments is available in [15_16_behavior.txt](https://github.com/Kong/charts/files/6208323/15_16_behavior.txt)

#### Checklist

### DIFF
--- a/charts/kong/templates/custom-resource-definitions.yaml
+++ b/charts/kong/templates/custom-resource-definitions.yaml
@@ -22,6 +22,7 @@
   {{- if .Capabilities.APIVersions.Has "apiextensions.k8s.io/v1/CustomResourceDefinition" -}}
     {{- $kongPluginCRD = (lookup "apiextensions.k8s.io/v1" "CustomResourceDefinition" "" "kongplugins.configuration.konghq.com") -}}
   {{- else -}}
+    {{/* TODO: remove the v1beta1 path when we no longer support k8s <1.16 */}}
     {{- $kongPluginCRD = (lookup "apiextensions.k8s.io/v1beta1" "CustomResourceDefinition" "" "kongplugins.configuration.konghq.com") -}}
   {{- end -}}
   {{- if $kongPluginCRD -}}
--- a/charts/kong/templates/custom-resource-definitions.yaml
+++ b/charts/kong/templates/custom-resource-definitions.yaml
@@ -18,7 +18,12 @@
        so this clause pretends the default didn't change if you have an existing release that relied
        on it
   */}}
-  {{- $kongPluginCRD := (lookup "apiextensions.k8s.io/v1" "CustomResourceDefinition" "" "kongplugins.configuration.konghq.com") -}}
+  {{- $kongPluginCRD := false -}}
+  {{- if .Capabilities.APIVersions.Has "apiextensions.k8s.io/v1/CustomResourceDefinition" -}}
+    {{- $kongPluginCRD = (lookup "apiextensions.k8s.io/v1" "CustomResourceDefinition" "" "kongplugins.configuration.konghq.com") -}}
+  {{- else -}}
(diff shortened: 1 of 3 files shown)


charts/kong/templates/custom-resource-definitions.yaml
```diff
@@ -18,7 +18,12 @@
        so this clause pretends the default didn't change if you have an existing release that relied
        on it
   */}}
-  {{- $kongPluginCRD := (lookup "apiextensions.k8s.io/v1" "CustomResourceDefinition" "" "kongplugins.configuration.konghq.com") -}}
+  {{- $kongPluginCRD := false -}}
+  {{- if .Capabilities.APIVersions.Has "apiextensions.k8s.io/v1/CustomResourceDefinition" -}}
+    {{- $kongPluginCRD = (lookup "apiextensions.k8s.io/v1" "CustomResourceDefinition" "" "kongplugins.configuration.konghq.com") -}}
+  {{- else -}}
+    {{- $kongPluginCRD = (lookup "apiextensions.k8s.io/v1beta1" "CustomResourceDefinition" "" "kongplugins.configuration.konghq.com") -}}
```
mflendrich, line 26:
```suggestion
    {{/* TODO: remove the v1beta1 path when we no longer support k8s <1.16 */}}
    {{- $kongPluginCRD = (lookup "apiextensions.k8s.io/v1beta1" "CustomResourceDefinition" "" "kongplugins.configuration.konghq.com") -}}
```

mflendrich (dismissed):
Since 1.22 will come with many breaking API removals (see Kong/kubernetes-ingress-controller#801), it has been assumed so far that the version of KIC that will be the first to support 1.22, will be the first to no longer support <1.16.

Therefore I find it unlikely that we'll run into a situation where we need to support both APIs at the same time. Instead, both KIC and the chart will probably need to make a breaking change cutting off <1.16 support around the time 1.22 becomes available (in a few months).

But this PR appears to make this breaking change smoother for the maintainers, though - which makes it worthwhile. One small comment attached (put a TODO to remove debt after we drop <1.16 which will almost certainly happen in a few months).

mflendrich (approved):

PR #324, opened 2021-03-29
[kong] feat: add possibility to use Container Storage Interface (e.g. mount Secrets from Azure Key Vault)

nicolai92:

In enterprise setups, you may not want to create secrets in Kubernetes, but store them elsewhere, e.g. in Azure Key Vault. The secrets can then be mounted to the respective container by making use of the [Container Storage Interface (CSI)](https://kubernetes-csi.github.io/docs/introduction.html). For this to work, a volume has to be created and mounted.

This PR adds the possibility to create a CSI volume and mount it to the respective container, so that secrets and certificates from e.g. Azure Key Vault can be used.

#### Checklist
- [ x] New or modified sections of values.yaml are documented in the README.md
- [ x] Title of the PR and commit headers start with chart name (e.g. `[kong]`)

### DIFF
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -19,29 +19,40 @@ $ helm install kong/kong --generate-name
 
 ## Table of contents
 
+- [Kong for Kubernetes](#kong-for-kubernetes)
+- [TL;DR;](#tldr)
+- [Table of contents](#table-of-contents)
 - [Prerequisites](#prerequisites)
 - [Install](#install)
 - [Uninstall](#uninstall)
-- [Kong Enterprise](#kong-enterprise)
 - [FAQs](#faqs)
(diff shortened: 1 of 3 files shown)


rainest:
The user-defined mounts added in https://github.com/Kong/charts/pull/317 should be able to handle this. @nicolai92 do you have any issue mounting your volumes using them?

nicolai92:
@rainest 

That‘s exactly what I was looking for, but couldn’t find it in the README. Although I haven’t tested the proposed solution yet, I think it should work and this PR can be closed then.

rainest:
It's merged but not yet released, so it will be available in an upcoming release. Currently we're in the process of rolling out 2.0, whose feature set is frozen, so that addition should go out in 2.1.

PR #327, opened 2021-03-30
[repo] keep master in sync with main

rainest:
This repo didn't get a sync job when we did the master/main switch. While there shouldn't be any actual use of master still, it may persist in user bookmarks or similar, so we may as well keep it updated.

Copied from the controller repo workflow: https://github.com/Kong/kubernetes-ingress-controller/blob/main/.github/workflows/master-mirrors-main.yaml

### DIFF
--- a/.github/workflows/master-mirrors-main.yaml
+++ b/.github/workflows/master-mirrors-main.yaml
@@ -1,0 +1,19 @@
+name: 'sync master with main'
+
+on:
+  push:
+    branches:
+      - 'main'
+
+jobs:
+  push-master:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+    - name: Push HEAD to master
+      # Deliberately not force-pushing: designed to fail if this would mean rewriting history
+      run: |
+        git push origin HEAD:master


.github/workflows/master-mirrors-main.yaml
```diff
@@ -0,0 +1,19 @@
...
+jobs:
+  push-master:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+    - name: Push HEAD to master
+      # Deliberately not force-pushing: designed to fail if this would mean rewriting history
```
shaneutt, line 17: question, non-blocking: how difficult would it be to add a Slack or email notification for @Kong/team-k8s if this ever failed?

shaneutt (approved):
LGTM, just one potential idea about notifying us if this ever fails, though we would basically never expect it to.

mflendrich (commented):
Or maybe let's explicitly drop `master` altogether with 2.0?

.github/workflows/master-mirrors-main.yaml
rainest: I believe we get this by default more or less, though it doesn't go to the entire team. GitHub action failures natively send an email alert to whomever took the action that initiated when they fail, e.g. I receive an email if the existing action to release charts on main push fails after I merge a PR to main. That might be a per-account preference, not sure.

rainest:
Ambivalent about removing the master branch entirely. Preserving it and keeping it in sync with main was the approach we took with the KIC and decK repos, so my thinking was to make charts follow the same pattern.

PR #343, opened 2021-04-13
[kong] fix appVersion type and CI issues

rainest:
- Forces YAML to treat the appVersion as a string.
- Fixes an issue with the lint arguments in CI.
- Sets target branches in CI.

Recent versions of Helm introduced a lint check for this. Doesn't appear that it causes issues beyond that (the chart still installs), but may as well fix it. Prior to this change:

```
$ helm version
version.BuildInfo{Version:"v3.5.3", GitCommit:"041ce5a2c17a58be0fcd5f5e16fb3e7e95fea622", GitTreeState:"dirty", GoVersion:"go1.16.2"}
13:33:43-0700 yagody $ helm lint
==> Linting .
[ERROR] Chart.yaml: appVersion should be of type string but it's of type float64
```
Reported elsewhere: https://github.com/Kong/kong-operator/pull/58

Also wanted to create a small PR to check whether an unrelated linter failure was due to weirdness handling changes in a PR base or something else after both https://github.com/Kong/charts/actions/runs/742365097 (a PR accidentally targeted at main initially) AND https://github.com/Kong/charts/actions/runs/746000964 (merger to next) failed. Turns out that was failing even on the branch push and PR runs even when the PR target was set to next.

This happened because the argument was never correct to begin with, which didn't matter until #327 because all comparisons were against the (previously very outdated) master branch. After adding that sync, comparisons were against the actual current version and began to fail.

I don't think `ct lint` uses the target branch for anything other than the version increment (its other checks are against the branch itself, not the target), but figured it was prudent to indicate the branches we actually care about.

#### Checklist

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -35,7 +35,7 @@ jobs:
           fi
 
       - name: Run chart-testing (lint)
-        run: ct lint --chart-repos bitnami=https://charts.bitnami.com/bitnami --remote origin
+        run: ct lint --chart-repos bitnami=https://charts.bitnami.com/bitnami --remote origin --target-branch main
 
       - name: Create kind cluster
         uses: helm/kind-action@v1.0.0
--- a/.github/workflows/non-main.yaml
+++ b/.github/workflows/non-main.yaml
@@ -39,7 +39,7 @@ jobs:
           fi
 
       - name: Run chart-testing (lint)
-        run: ct lint --chart-repos bitnami=https://charts.bitnami.com/bitnami --check-version-increment=false --remote origin
+        run: ct lint --chart-repos bitnami=https://charts.bitnami.com/bitnami --check-version-increment=false --remote origin --target-branch next
 
       - name: Create kind cluster
         uses: helm/kind-action@v1.0.0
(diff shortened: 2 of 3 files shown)


shaneutt (approved):
:+1:

mflendrich (commented):
I've read up on `--target-branch` and, IIUC, this setting is used by certain checks (which - please correct me if I'm wrong - won't run in the workflow changed by this PR) as a _base_ that we're comparing the chart under test against.

If we set the `--target-branch` to the `main`/`next` Git ref respectively, `ct` will assume that it needs to test the diff between the commit it's been run on, and that tip, which is either:
- an empty diff, or
- if `ct` happens to be running when the tip of `main`/`next` has already changed (e.g. due to two quick pushes one after another, or a re-run historical workflow): a completely unrelated newer state of the `main`/`next` branch, which will trick `ct` into thinking that it's in fact testing the _reverse_ of the change newly made to `main`/`next`.

If my understanding is valid, then the addition of `--target-branch` in this form will mislead `ct` into thinking that it shall test a change made between the commit it's being run on, and the current state behind the ref passed to it. This appears as incorrect to me.

Therefore (again - if I'm not mistaken - you probably understand this change better than I do), I think that we shouldn't be adding `--target-branch` in its current form.

rainest:
@mflendrich the target branch is _always_ set though--not setting it just falls back to the default, `master`, which is wrong-ish for PRs to `main` (though `master` is technically unused, it's kept up to date with `main`) and definitely wrong for PRs to `next`.

PR #344, opened 2021-04-14
[repo] add CT arguments to install also

rainest:
Copy the common arguments used in lint to the ct install command also. They shouldn't differ.

Main issue is the lack of extra repo: https://github.com/Kong/charts/runs/2345521060?check_suite_focus=true

This weirdly doesn't always matter. https://github.com/Kong/charts/runs/2337522014 is the same workflow, but _doesn't_ care that we didn't indicate what bitnami was during install. Not sure why, but not too interested in investigating that in depth, since there's no good reason we shouldn't use the same set of common args for each of the ct steps.

#### Checklist

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -42,7 +42,7 @@ jobs:
         if: steps.list-changed.outputs.changed == 'true'
 
       - name: Run chart-testing (install)
-        run: ct install
+        run: ct install --chart-repos bitnami=https://charts.bitnami.com/bitnami --remote origin --target-branch main
   release:
     needs: lint-test
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yaml
```diff
@@ -42,7 +42,7 @@ jobs:
         if: steps.list-changed.outputs.changed == 'true'
 
       - name: Run chart-testing (install)
-        run: ct install
+        run: ct install --chart-repos bitnami=https://charts.bitnami.com/bitnami --remote origin --target-branch main
```
mflendrich: If the purpose of this PR is to guarantee that arguments to all `ct` calls match, then we should reuse instead of copying.

For example, by creating a `CT_ARGS` env variable in the job.

mflendrich (approved):

PR #345, opened 2021-04-14
docs: add a note to serviceMonitor

### LINKED ISSUE #311: ServiceMonitor does not select for kong CP
Hi,

serviceMonitor is generated for Kong CP however, there is missing enabled-metrics in kong admin service.

serviceMonitor.yaml:
```javascript 
  selector:
    matchLabels:
      enable-metrics: "true"
      {{- include "kong.metaLabels" . | nindent 6 }}
```

service-kong-admin.yaml:
```javascript
  labels:
    enable-metrics: "true" -----> missing part
```

Helm Chart: 1.15.0

shaneutt:
Sometimes `values.yaml` fields can be slightly misleading, in
this case setting `serviceMonitor.enabled=true` doesn't inherently
place the label for all components (such as the admin api) and it
may not be intuitive that several other configurations must be updated
in order to achieve that. As such this commit adds some helper text
and a link to the README to provide some hints, and more context.

Resolves #311

### DIFF
--- a/charts/kong/values.yaml
+++ b/charts/kong/values.yaml
@@ -573,11 +573,8 @@ securityContext: {}
 
 serviceMonitor:
   # Specifies whether ServiceMonitor for Prometheus operator should be created
-  # NOTE: setting `enabled: true` here enables the ServiceMonitor resource, but
-  #       by itself doesn't enable the label for all components that can be monitored.
-  #       For instance, if you want to turn metrics on for the Admin API you must add
-  #       a `metrics-enabled: true` label in the admin.labels above.
-  #       For more context see: https://github.com/Kong/charts/blob/main/charts/kong/README.md#prometheus-operator-integration
+  # If you wish to gather metrics from a Kong instance with the proxy disabled (such as a hybrid control plane), see:
+  # https://github.com/Kong/charts/blob/main/charts/kong/README.md#prometheus-operator-integration
   enabled: false
   # interval: 10s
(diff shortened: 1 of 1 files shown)


charts/kong/values.yaml
```diff
@@ -573,6 +573,11 @@ securityContext: {}
 
 serviceMonitor:
   # Specifies whether ServiceMonitor for Prometheus operator should be created
+  # NOTE: setting `enabled: true` here enables the ServiceMonitor resource, but
+  #       by itself doesn't enable the label for all components that can be monitored.
+  #       For instance, if you want to turn metrics on for the Admin API you must add
+  #       a `metrics-enabled: true` label in the admin.labels above.
+  #       For more context see: https://github.com/Kong/charts/blob/main/charts/kong/README.md#prometheus-operator-integration
```
```suggestion
  # If you wish to gather metrics from a Kong instance with the proxy disabled (such as a hybrid control plane), see:
  # https://github.com/Kong/charts/blob/main/charts/kong/README.md#prometheus-operator-integration
```

rainest (changes requested):
I'll still try and push for brevity here :slightly_smiling_face: 

We also _don't_ want to imply that there are admin API metrics--there aren't, and the label really should be set on one service only. A DeploymentMonitor would be more appropriate for how our metrics actaully work, but the Prometheus operator doesn't offer that.

rainest (approved):

PR #347, opened 2021-04-19
[kong] add support for DaemonSets

rainest:
Add a toggle to use a DaemonSet instead of a Deployment.

Carried over from https://github.com/Kong/charts/pull/269 with some changes to docs and unintended changes.

#### Checklist

### DIFF
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -36,6 +36,7 @@ $ helm install kong/kong --generate-name
   - [Sidecar containers](#sidecar-containers)
   - [User Defined Volumes](#user-defined-volumes)
   - [User Defined Volume Mounts](#user-defined-volume-mounts)
+  - [Using a DaemonSet](#using-a-daemonset)
   - [Example configurations](#example-configurations)
 - [Configuration](#configuration)
   - [Kong Parameters](#kong-parameters)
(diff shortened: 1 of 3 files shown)


mflendrich (approved):
I'm torn.

When I first saw this PR, I was surprised by how easy (on the code) introduction of DaemonSets was in this PR. Then I looked at the API spec for `DeploymentSpec` and `DaemonSetSpec` in the k8s api to discover that they're similar, but not the same type - in fact, they share a subset of fields, but the types are not identical. This effectively means that this "unified" template be restricted to the set product of Deployment and DaemonSet fields by making this change, or need to introduce some terrible branching depending on which deployment method was chosen.

To me the "prudent" solution for this (in the strongly typed k8s world) would be to define two conditionally enabled templates (one for a deployment, one for a daemonset) with 99%-100% of the spec being generated by more specific templates, thus avoiding the hack of treating schemas of distinct types as identical.

But maybe that would add complexity to an already bloated chart and we'll never need it :shrug:. Time will show, I guess.

Use your best judgment.

rainest:
There isn't a whole great deal of difference since most of the actual content is in the Pod template, which is the same object type in both. Split templates could move that into _helpers.tpl, which grows ever larger, and has become a bit cumbersome to deal with.

The single template branching approach is effectively what we're doing with the `replicas` gate, and is probably fine given that there aren't many differences between the two controllers. As for the rest:

- `paused` doesn't look like you'd manage it with the chart, since you'd instead manage it with `kubectl rollout pause`/`kubectl rollout resume`.
- `progressDeadlineSeconds` we don't use yet, but would need to handle it the same as `replicas`
- `strategy` versus `updateStrategy` is odd since it's fully user-defined and only rendered if set. We could overload the values setting, populate the correct name depending on which controller you're using, and still leave setting the correct contents up to the user.

PR #362, opened 2021-05-12
Update README.md

svenwal:
The README talks about Enterprise running without license. This is misleading as only a small subset is available without license


mflendrich (changes requested):
Looks good, but please target `next` rather than `main`, because of how our releasing logic works.

PR #366, opened 2021-05-29
[kong] add hostAliases for deployment

lumincui:
expose hostAliases to values.yaml to be able to inject host aliases
into containers, so that we can resolve additional domain name which
can't be looked-up directly from dns server.

What:
expose hostAliases to values.yaml to be able to inject host aliases into containers.

Why:
In some case we need resolve custom domain which can't be looked-up from dns server

#### Checklist


lumincui:
@rainest would you please help to review this pr?

Issue #371, opened 2021-06-11
"wait-for-db" init container fails on unavailable custom plugins


Tieske:
We're using a custom image that has custom plugins installed and in use.

When doing an upgrade the init container that wait for the postgres database to be available fails with;
```
2021/06/11 09:20:17 [error] 5276#0: init_by_lua error: /usr/local/share/lua/5.1/kong/init.lua:582: error building initial plugins: helloplugin plugin is in use but not enabled                                  ││ stack traceback:                                                                                                                                                                                                 ││     [C]: in function 'error'                                                                                                                                                                                     │
│     /usr/local/share/lua/5.1/kong/init.lua:582: in function 'init'                                                                                                                                               ││     init_by_lua:3: in main chunk                                                                                                                                                                                 ││   Run with --v (verbose) or --vv (debug) for more details                                                                                                                                                        ││ waiting for db                                                                                                                                                                                                   │
│ 2021/06/11 09:20:19 [warn] 5287#0: *2 [lua] plugins.lua:231: load_plugin(): plugin 'route-by-header' has been deprecated, context: ngx.timer
```

Quick workaround is adding an environment variable `KONG_PLUGINS=bundled,helloplugin`, but that should not be necessary since the docker image has start up scripts that do this already.

The reason is in this line: https://github.com/Kong/charts/blob/bd8e8600f708a956e4041ec84d6a0814e1fbcfd2/charts/kong/templates/_helpers.tpl#L507 where it does the regular `kong start` instead of using the docker entrypoint scripts located at `/docker-entrypoint.sh` in the docker container.

For reference, the entrypoint scripts: https://github.com/Kong/docker-kong/blob/master/alpine/docker-entrypoint.sh
And if customized with custom plugins this: https://github.com/Kong/docker-kong/blob/master/customize/packer.lua#L376-L411

fyi @mflendrich @lokesh-chechani

rainest:
How are you loading the custom plugin elsewhere in values.yaml? It sounds like this isn't using the [dedicated plugin loader](https://github.com/Kong/charts/blob/kong-2.1.0/charts/kong/values.yaml#L251-L270) (which updates `KONG_PLUGINS`) and instead have manual mounts.

`wait-for-db` intentionally doesn't use the entrypoint script because we want the container to exit, whereas the entrypoint script will run indefinitely. If you're not using the dedicated plugin loader, manually setting `KONG_PLUGINS` is probably the correct approach.

[...]

PR #377, opened 2021-06-22
[kong] use args instead of command

rainest:
Uses `args` instead of `command` to invoke non-standard commands for Kong containers. This allows the standard image entrypoint to run [rather than replacing it](https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes).

Inject the no daemonize environment variable directly into the wait-for-db command. There's currently a limitation in the entrypoint that overrides the container environment. https://github.com/Kong/docker-kong/pull/474 intends to fix it but we'll need this for backwards compatibility.

See the commit message for a bit more detail on why we need this.

  - fixes https://github.com/Kong/charts/issues/371

#### Checklist

### DIFF
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -508,8 +508,9 @@ The name of the service used for the ingress controller's validation webhook
   env:
   {{- include "kong.env" . | nindent 2 }}
 {{/* TODO: the rm command here is a workaround for https://github.com/Kong/charts/issues/295
-     It should be removed once that's fixed */}}
-  command: [ "/bin/sh", "-c", "until kong start; do echo 'waiting for db'; sleep 1; done; kong stop; rm -fv {{ $sockFile | squote }}"]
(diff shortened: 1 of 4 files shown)


Tieske:
fyi: PR https://github.com/Kong/docker-kong/pull/474 has been merged

Tieske:
@rainest will this work backward compatible? The fix in https://github.com/Kong/docker-kong/pull/474 shows that old images will ALWAYS set the daemonize option to off. So even if you pass the env var in, the entrypoint will override it again and set it to off again.

Not sure how to fix this....

rainest:
https://github.com/Kong/charts/pull/377/files#diff-3e8a71242505abb72903a338eef1cdea2cd2c674f877c59378eff807951a1fa8R513

The export there will run after the one in the entrypoint script. CI would fail otherwise (since the initcontainer would never complete).

shaneutt (approved):
LGTM

Once we're happy with this in testing, good to go :+1:

PR #379, opened 2021-06-22
[kong] add support for UDP streams

### LINKED ISSUE #329: UDP Stream Listener templating hardcoded to TCP proto
Using the chart with the following flags:

```shell
    --set proxy.stream[0].containerPort=9999
    --set proxy.stream[0].servicePort=9999
    --set proxy.stream[0].parameters[0]=udp
    --set proxy.stream[0].parameters[1]=reuseport
```

Results in the following `ContainerPort` for the Kong proxy pod:

```yaml
- name: stream-9999
  containerPort: 9999
  protocol: TCP
```

And the `ServicePort`:

```yaml
- name: stream-9999
  port: 9999
  protocol: TCP
  targetPort: 9999
```

This appears to be hardcoded: https://github.com/Kong/charts/blob/main/charts/kong/templates/deployment.yaml#L118

Due to constraints on `LoadBalancer` type `Service`s in the Kubernetes API which disallows mixing protocols for `ServicePorts`, it looks like we'll need to create a separate `Service` when a `UDP` port is present.

rainest:
- Adds a new UDP proxy Service. This needs to be separate from the standard proxy Service because of the [limitations on multi-protocol LoadBalancers](https://kubernetes.io/docs/concepts/services-networking/service/#load-balancers-with-mixed-protocol-types).
- Updates the stream listener generator and stream Service template handler to be protocol-aware.
- Adds protocol configuration to the TCP stream listen examples under the standard proxy.
- Includes UDP stream ports in the Deployment template.

  - fixes #329 

TBD

#### Checklist

### DIFF
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -160,13 +160,13 @@ spec:
   {{- end }}
   {{- if (hasKey . "stream") }}
   {{- range .stream }}
-  - name: stream-{{ if (eq .protocol "UDP") }}udp-{{ end }}{{ .containerPort }}
+  - name: stream-{{ if (eq (default .protocol "TCP") "UDP") }}udp-{{ end }}{{ .containerPort }}
     port: {{ .servicePort }}
     targetPort: {{ .containerPort }}
     {{- if (and (or (eq $.type "LoadBalancer") (eq $.type "NodePort")) (not (empty .nodePort))) }}
     nodePort: {{ .nodePort }}
     {{- end }}
-    protocol: {{ .protocol }}
+    protocol: {{ .protocol | default "TCP" }}
   {{- end }}
   {{- end }}
   {{- if .externalTrafficPolicy }}
@@ -255,10 +255,11 @@ Create KONG_STREAM_LISTEN string
     {{- $listenConfig := dict -}}
     {{- $listenConfig := merge $listenConfig . -}}
     {{- $_ := set $listenConfig "address" "0.0.0.0" -}}
-    {{- if (eq .protocol "UDP") -}}
-      {{/* You set NGINX stream listens to UDP using a parameter because reasons. Since our configuration
-           is dual-purpose for both the Service and listen string, we forcibly inject this parameter if
-           that's the Service protocol. */}}
+    {{/* You set NGINX stream listens to UDP using a parameter because reasons. Since our configuration
+         is dual-purpose for both the Service and listen string, we forcibly inject this parameter if
+         that's the Service protocol. The default handles configs that predate the addition of the
+         protocol field, where we only supported TCP. */}}
+    {{- if (eq (default .protocol "TCP") "UDP") -}}
       {{- $_ := set $listenConfig "parameters" (append .parameters "udp") -}}
     {{- end -}}
     {{- $unifiedListen = append $unifiedListen (include "kong.singleListen" $listenConfig ) -}}
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -160,7 +160,7 @@ spec:
   {{- end }}
   {{- if (hasKey . "stream") }}
   {{- range .stream }}
-  - name: stream-{{ .containerPort }}
+  - name: stream-{{ if (eq .protocol "UDP") }}udp-{{ end }}{{ .containerPort }}
     port: {{ .servicePort }}
     targetPort: {{ .containerPort }}
     {{- if (and (or (eq $.type "LoadBalancer") (eq $.type "NodePort")) (not (empty .nodePort))) }}
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -255,10 +255,10 @@ Create KONG_STREAM_LISTEN string
     {{- $listenConfig := dict -}}
     {{- $listenConfig := merge $listenConfig . -}}
     {{- $_ := set $listenConfig "address" "0.0.0.0" -}}
-    {{/* You set NGINX stream listens to UDP using a parameter because reasons. Since our configuration
-         is dual-purpose for both the Service and listen string, we forcibly inject this parameter if
-         that's the Service protocol. The default handles configs that predate the addition of the
-         protocol field, where we only supported TCP. */}}
+    {{/* You set NGINX stream listens to UDP using a parameter due to historical reasons.
+         Our configuration is dual-purpose, for both the Service and listen string, so we
+         forcibly inject this parameter if that's the Service protocol. The default handles
+         configs that predate the addition of the protocol field, where we only supported TCP. */}}
     {{- if (eq (default .protocol "TCP") "UDP") -}}
       {{- $_ := set $listenConfig "parameters" (append .parameters "udp") -}}
     {{- end -}}
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -166,7 +166,7 @@ spec:
     {{- if (and (or (eq $.type "LoadBalancer") (eq $.type "NodePort")) (not (empty .nodePort))) }}
     nodePort: {{ .nodePort }}
     {{- end }}
-    protocol: TCP
+    protocol: {{ .protocol }}
   {{- end }}
   {{- end }}
   {{- if .externalTrafficPolicy }}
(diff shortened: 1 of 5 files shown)


rainest:
Draft pending:
- Documentation
- Default handling for backwards compatibilty with existing stream configurations that omit protocol (which is why CI fails currently).
- Mulling over design.

The initial proposal for this explicitly declares the protocol in a stream listen configuration. There are no restrictions on which protocols you use, and this does allow you to create configurations that do not work (including both UDP and TCP streams in a single LoadBalancer Service).

My rationale for exposing potentially broken configuration is:
- Current support _does_ allow you to mix protocols if you use a non-LoadBalancer Service. While this is much less common for the proxy, it is not unheard of.
- Exposing the protocol allows for forwards compatibility with multi-protocol LoadBalancers once they become available.
- The helper templates aren't aware of which service they're handling, and so lack a good way to determine which protocol to set without an explicit declaration. Although we could inject the protocol higher up in the template, such automagic configuration is ideally avoided.

So I think I'll probably stick with that over something that instead adds the protocol for you based on which service you're configuring, but do want to sleep on it.

charts/kong/templates/_helpers.tpl
```diff
@@ -607,7 +613,24 @@ the template that it itself is using form the above sections.
...
-{{- $_ := set $autoEnv "KONG_STREAM_LISTEN" (include "kong.streamListen" .Values.proxy) -}}
+{{- $streamStrings := list -}}
+{{- if .Values.proxy.enabled -}}
+  {{- $tcpStreamString := (include "kong.streamListen" .Values.proxy) -}}
+  {{- if (not (eq $tcpStreamString "")) -}}
+    {{- $streamStrings = (append $streamStrings $tcpStreamString) -}}
+  {{- end -}}
+{{- end -}}
+{{- if .Values.udpProxy.enabled -}}
+  {{- $udpStreamString := (include "kong.streamListen" .Values.udpProxy) -}}
```
ccfishk, line 625: can we use const, or env replace "kong.streamListen" here

charts/kong/templates/service-kong-udp-proxy.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if .Values.deployment.kong.enabled }}
+{{- if and .Values.udpProxy.enabled -}}
+{{- $serviceConfig := dict -}}
+{{- $serviceConfig := merge $serviceConfig .Values.udpProxy -}}
+{{- $_ := set $serviceConfig "fullName" (include "kong.fullname" .) -}}
+{{- $_ := set $serviceConfig "namespace" (include "kong.namespace" .) -}}
+{{- $_ := set $serviceConfig "metaLabels" (include "kong.metaLabels" .) -}}
+{{- $_ := set $serviceConfig "selectorLabels" (include "kong.selectorLabels" .) -}}
+{{- $_ := set $serviceConfig "serviceName" "udp-proxy" -}}
```
ccfishk, line 9: Should not serviceName include "kong-ingress" ? should they have overlap ?

charts/kong/values.yaml
```diff
@@ -256,6 +261,40 @@ proxy:
...
+  labels: {}
+  # Optionally specify a static load balancer IP.
+  # loadBalancerIP:
+
+  # Define stream (UDP) listen
+  # To enable, remove "{}", uncomment the section below, and select your desired
+  # ports and parameters. Listens are dynamically named after their servicePort,
+  # e.g. "stream-9000" for the below.
+  stream: {}
+    #   # Set the container (internal) and service (external) ports for this listen.
```
ccfishk, line 286: Do we support using the same ports as TCP does ? 
The collision port we faced during TCP/UDP ingress tests are not a blocker here ?

ccfishk (commented):
Add few comments

charts/kong/templates/_helpers.tpl
rainest: Not sure I follow--can you elaborate? How would you want this to function differently?

charts/kong/templates/service-kong-udp-proxy.yaml
rainest: We prefix the generic name (`udp-proxy` here) with the release name in the helper template that we call on everything in `$serviceConfig` after: https://github.com/Kong/charts/blob/kong-2.1.0/charts/kong/templates/_helpers.tpl#L103

charts/kong/templates/_helpers.tpl
ccfishk: Can you replace "kong.streamListen"  using Env "KONG_STREAM_LISTEN

charts/kong/templates/deployment.yaml
```diff
@@ -134,7 +134,15 @@ spec:
...
           hostPort: {{ .hostPort }}
           {{- end}}
-          protocol: TCP
+          protocol: {{ .protocol }}
+        {{- end }}
+        {{- range .Values.udpProxy.stream }}
+        - name: stream-udp-{{ .containerPort }}
+          containerPort: {{ .containerPort }}
+          {{- if .hostPort }}
+          hostPort: {{ .hostPort }}
```
ccfishk, line 143: targetPort is missed within Service Definition.

ccfishk (commented):
Add few comments

charts/kong/templates/_helpers.tpl
rainest: Still not sure I understand. `kong.streamListen` is a function this uses to generate a set of stream listen configurations (a comma-separated list of `IP:port arguments` strings) from the service configuration in values.yaml.

If you're asking if it can be overridden with user configuration, it can be--anything that we populate in `$autoEnv` can be overridden by setting it in the `env` section of values.yaml--though normally this wouldn't be, as it's pretty closely tied to all the rest of the auto-generated service configuration.

charts/kong/templates/deployment.yaml
rainest: This is the Deployment container template; it only includes the target port.

We use target and container port interchangeably, since they're similar things with different names depending on whether you're building a Service or Deployment. We populate the Service targetPort value at https://github.com/Kong/charts/blob/bd8e8600f708a956e4041ec84d6a0814e1fbcfd2/charts/kong/templates/_helpers.tpl#L140

ccfishk: understand. the changes lgtm.

charts/kong/templates/_helpers.tpl
```diff
@@ -255,12 +255,19 @@ Create KONG_STREAM_LISTEN string
     {{- $listenConfig := dict -}}
     {{- $listenConfig := merge $listenConfig . -}}
     {{- $_ := set $listenConfig "address" "0.0.0.0" -}}
+    {{/* You set NGINX stream listens to UDP using a parameter because reasons. Since our configuration
+         is dual-purpose for both the Service and listen string, we forcibly inject this parameter if
+         that's the Service protocol. The default handles configs that predate the addition of the
+         protocol field, where we only supported TCP. */}}
```
shaneutt: I'm worried that the language here would potentially confuse some readers, how would you feel about this instead?
```suggestion
    {{/* You set NGINX stream listens to UDP using a parameter due to historical reasons.
         our configuration is dual-purpose for both the Service and listen string, we forcibly inject
         this parameter ifthat's the Service protocol. The default handles configs that predate the
         addition of theprotocol field, where we only supported TCP. */}}
```

charts/kong/templates/_helpers.tpl
```diff
@@ -255,12 +255,19 @@ Create KONG_STREAM_LISTEN string
...
+    {{- if (eq (default .protocol "TCP") "UDP") -}}
+      {{- $_ := set $listenConfig "parameters" (append .parameters "udp") -}}
+    {{- end -}}
     {{- $unifiedListen = append $unifiedListen (include "kong.singleListen" $listenConfig ) -}}
   {{- end -}}
 
   {{- $listenString := ($unifiedListen | join ", ") -}}
   {{- if eq (len $listenString) 0 -}}
-    {{- $listenString = "off" -}}
+    {{- $listenString = "" -}}
```
shaneutt, line 270: Q: what's the context here, this never worked?

charts/kong/values.yaml
```diff
@@ -226,12 +226,17 @@ proxy:
   # To enable, remove "{}", uncomment the section below, and select your desired
   # ports and parameters. Listens are dynamically named after their servicePort,
   # e.g. "stream-9000" for the below.
+  # Note: although you can select the protocol here, you cannot set UDP if you
+  # use a LoadBalancer Service due to limitations in current Kubernetes versions.
+  # To proxy both TCP and UDP with LoadBalancers, you must enable the udpProxy Service
+  # in the next section and place all UDP stream listen configuration under it.
```
shaneutt, line 232: Good comments, very informative and in a location I personally appreciate finding them (in the config file itself) :+1:

shaneutt (approved):
LGTM, but do see my comments for your considerations (non blocking)

charts/kong/templates/_helpers.tpl
rainest: Moved to https://github.com/Kong/charts/pull/379/files#diff-3e8a71242505abb72903a338eef1cdea2cd2c674f877c59378eff807951a1fa8R631-R633

The original version of this handled disabling the listen in the context of a single type of listen (TCP), and the updated version needs to combine multiple `kong.streamListen` results into a single stream listen config. If all are empty, the Kong configuration should be `off`.

Returning the empty string seemed a bit easier to work with elsewhere (in the autoEnv builder). Templating doesn't allow for the most elegant implementations--this is yet another helper that's adding to my increasing list of things that'd be better-handled in Go implementation that provides luxuries like actual types and typed function returns.

charts/kong/templates/_helpers.tpl
rainest: b2522aa6f as Github edit did unpleasant things to spacing.

jrsmroz:
Well, technically it's all correct but the naming is bit off. TCP is stream oriented protocol, so naming it stream is fine. UDP is packet oriented protocol, without guarantee that packet is not lost or order of delivery, naming it a stream is misleading a lot.

PR #382, opened 2021-06-24
[kong] Remove maxUnavailable PDB default

rainest:
Helm's handling of default values means you cannot unset this from your values.yaml. The default values.yaml will repopulate empty values. This setting is mutually exclusive with minUnavailable, so the default prevents you from using minUnavailable at all.

#### Checklist

### DIFF
--- a/charts/kong/values.yaml
+++ b/charts/kong/values.yaml
@@ -560,6 +560,9 @@ autoscaling:
 # Kong Pod Disruption Budget
 podDisruptionBudget:
   enabled: false
+  # Uncomment only one of the following when enabled is set to true
+  # maxUnavailable: "50%"
+  # minUnavailable: "50%"
 
 podSecurityPolicy:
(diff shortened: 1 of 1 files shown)


shaneutt (approved):
LGTM, let's just make sure before we merge we do whatever manual testing might be worth doing to double tap it? :+1:

rainest:
Not much you can really manually test to verify the change--the actual value you use isn't any different, it's just no longer a default value. If you place the original default in your values.yaml, the end result is the same.

This is a somewhat annoying breaking change, because if you had previously simply included `podDisruptionBudget.enabled: true` in your values, you now need to also specify either the min or max, else your PDB spec will be invalid. Adding a fallback "if neither min nor max are set, set max with a default value" inside the template seems overkill though, given that k8s just reject the invalid resource on upgrade if you don't add it. c'est la templatevie :disappointed:

PR #385, opened 2021-06-25
[kong] Admission controller - Bring your own certificates (#381)

### LINKED ISSUE #381: Admission controller - Bring your own certificates
Hi,

In the context of using cert-manager it should be feasible to provide the admission webhook certificate to the Helm chart.

I was thinking of adding properties to the definition like this, in the values.yaml

```
    admissionWebhook:
    enabled: false
    failurePolicy: Fail
    port: 8080
    certificate:
      provided: false
    # Specifiy the secretName when the certificate is provided via a TLS secret
    # secretName: "" 
    # Specifiy the CA bundle of the provided certificate.  
    # This is a PEM encoded CA bundle which will be used to validate the webhook certificate. If unspecified, system trust roots on the apiserver are used.
    # ca: 
    #   bundle: 
    #     | Add the CA bundle content here.
```

Then within the admission-webhook.yaml and in the _helpers.tpl some checks area performed to apply to right setup.

I will submit a PR for  that.

ludovic-pourrat:
Add the support of providing your own certificate to the admission controller.

  - fixes #381

#### Checklist


charts/kong/templates/_helpers.tpl
```diff
@@ -409,7 +409,11 @@ The name of the service used for the ingress controller's validation webhook
 {{- if .Values.ingressController.admissionWebhook.enabled }}
 - name: webhook-cert
   secret:
+    {{- if not .Values.ingressController.admissionWebhook.certificate.provided }}
     secretName: {{ template "kong.fullname" . }}-validation-webhook-keypair
+    {{- else }}
+    secretName: {{ .Values.ingressController.admissionWebhook.certificate.secretName }}
+    {{- end }}
```
shaneutt: nit: it's a small thing, but unless there's a reason to check the negative, I feel like we should just make this a base if:
```suggestion
    {{- if .Values.ingressController.admissionWebhook.certificate.provided }}
    secretName: {{ .Values.ingressController.admissionWebhook.certificate.secretName }}
    {{- else }}
    secretName: {{ template "kong.fullname" . }}-validation-webhook-keypair
    {{- end }}  
```

shaneutt (commented):
Thanks for this contribution! :bow: 

@rainest given your expertise I would like if you could review this as well, but to me this seems functional however I'm wondering if we want to name the values differently?

charts/kong/templates/_helpers.tpl
ludovic-pourrat: Agreed !

rainest (changes requested):
Overriding the default generated variable values in the template works around the issues I'd laid out in https://github.com/Kong/charts/issues/252#issuecomment-772976255 (it's always at the default location, so the controller doesn't need to worry about pulling them out of a ConfigMap and loading them into the environment instead), so :partying_face: 

Please change `admissionWebhook.certificate.ca.bundle` to `admissionWebhook.certificate.caBundle`. That shouldn't be a nested object unless we need to include other CA bundle parameters, and I don't think we ever will.

Along with that, I believe you need something like this:

```
  clientConfig:
    {{- if not .Values.ingressController.admissionWebhook.certificate.provided }}
    caBundle: {{ b64enc $caCert }}
    {{- else }}
    {{- if .Values.ingressController.admissionWebhook.certificate.caBundle }}
    caBundle: {{ b64enc .Values.ingressController.admissionWebhook.certificate.caBundle }}
    {{- end }}
    {{- end }}
```
Otherwise, it will always try to set a `caBundle` field (even if the value is nil) when you set a provided certificate, which you don't want if you issued your certificate from the cluster CA. Do you agree?

ludovic-pourrat:
@rainest, yes I agree to your feedback, I will change accordingly. Thanks !

rainest (approved):

PR #387, opened 2021-06-28
[kong] Custom Container SecurityContext fields

### LINKED ISSUE #373: Custom Container SecurityContext fields
Hi,

I found that there is no place to include a security context in init-containers (wait-for-db) and containers. recently, our security scan tools pointed that such securityContext is mandatory:

```yaml
securityContext:
  readOnlyRootFilesystem: false
```

Currently, kong-helm supported securityContext for kong pods. for this request, we can use containers.securityContext? let me know if what's the suitable name for this.

ttony:
  - fixes #373 

#### Checklist

### DIFF
--- a/charts/kong/templates/migrations.yaml
+++ b/charts/kong/templates/migrations.yaml
@@ -55,6 +55,8 @@ spec:
       - name: {{ template "kong.name" . }}-migrations
         image: {{ include "kong.getRepoTag" .Values.image }}
         imagePullPolicy: {{ .Values.image.pullPolicy }}
+        securityContext:
+        {{ toYaml .Values.containerSecurityContext | nindent 10 }} 
         env:
         {{- include "kong.no_daemon_env" . | nindent 8 }}
         args: [ "/bin/sh", "-c", "kong migrations bootstrap" ]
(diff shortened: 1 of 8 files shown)


charts/kong/templates/migrations.yaml
```diff
@@ -55,6 +55,8 @@ spec:
       - name: {{ template "kong.name" . }}-migrations
         image: {{ include "kong.getRepoTag" .Values.image }}
         imagePullPolicy: {{ .Values.image.pullPolicy }}
+        securityContext:
+        {{ toYaml .Values.containerSecurityContext | nindent 10 }}
```
shaneutt, line 59: It occurred to me that we should be doing conditional value setting here, basically don't set this field as empty if the `values.yaml` is empty... @rainest?

rainest: `toYaml` (which is [basically `yaml.Marshal()`](https://github.com/helm/helm/blob/a499b4b179307c267bdf3ec49b880e3dbd2a5591/pkg/engine/funcs.go#L79-L90) apparently handles nil inputs fine, and returns an empty object, which is valid. Actual template output is:

```
      containers:
      - name: ingress-controller
        securityContext:
      
          {}  
```
Which is a bit ugly, but valid. :shrug:

rainest (approved):

PR #389, opened 2021-06-29
[kong] CKV_K8S_38: Ensure that Service Account Tokens are only mounted where necessary

### LINKED ISSUE #388: CKV_K8S_38: Ensure that Service Account Tokens are only mounted where necessary
Hi Kong Team,

I would like to address this security best practice for kubernetes  -- CKV_K8S_38. To address this, we need to add automountServiceAccountToken: false. 

by default, we set to false in values.yaml as:

```yaml
automountServiceAccountToken: false
```

ttony:
  - fixes #388 

#### Checklist

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -40,7 +40,7 @@ jobs:
         run: ct lint $COMMON_CT_ARGS
 
       - name: Create kind cluster
-        uses: helm/kind-action@v1.0.0
+        uses: helm/kind-action@v1.1.0
         if: steps.list-changed.outputs.changed == 'true'
 
       - name: Run chart-testing (install)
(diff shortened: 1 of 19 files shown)


ttony:
@rainest, current test case required service account token to be mounted. So, should we set the default value as true rather than false? or should be change the test case to explicitly enabled mount service account token?

rainest:
This shouldn't need an independent toggle in values.yaml. The token must be mounted on the Deployment's Pods when the controller is enabled, so you can set the automount setting to the value of `.Values.ingressController.enabled`. You can always set it to `false` on the migrations Jobs.

ttony:
thanks for the feedback

rainest (approved):

PR #397, opened 2021-07-02
[kong] initContainer for job

### LINKED ISSUE #390: initContainer for job
Hi Kong Team,

To include initContainer for job type which feature has been implemented in deployment. Also, initContainer's yaml can contain other helm's values as a reference.

ttony:
  - fixes #390 

#### Checklist

### DIFF
--- a/.github/workflows/main-pr.yaml
+++ b/.github/workflows/main-pr.yaml
@@ -1,17 +1,14 @@
-name: Lint and Test Charts
+name: Release Charts
 
 on:
-  push:
-    branches:
-    - 'next'
   pull_request:
     branches:
-    - 'next'
+      - main
 
 jobs:
   lint-test:
     env:
-      COMMON_CT_ARGS: "--chart-repos bitnami=https://charts.bitnami.com/bitnami --remote origin --target-branch next"
+      COMMON_CT_ARGS: "--chart-repos bitnami=https://charts.bitnami.com/bitnami --remote origin --target-branch main"
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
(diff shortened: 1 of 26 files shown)


rainest (changes requested):
The history here is a bit confusing--it looks like you're working from a `next` branch on your fork that contains all your changes, not just those related to this change. Can you instead create a feature branch on your fork that contains only the commits related to the initContainer change and PR from that? They don't actually make a difference here (since, for example, https://github.com/Kong/charts/pull/389 was already merged and includes those changes), but using a branch specific to the PR will make review easier.

As far as the changes themselves, what's the reasoning for the change from

```
{{- toYaml .Values.deployment.userDefinedVolumes }}
```
to
```
{{- tpl (toYaml .Values.deployment.userDefinedVolumes) . }}
```
?

That looks like it should remain just `toYaml` (so the changeset adds the initContainers to Jobs, but doesn't change how the template renders the initContainers). Wrapping it in [`tpl`](https://helm.sh/docs/howto/charts_tips_and_tricks/#using-the-tpl-function) suggests that it'd take the content from values.yaml, run it through the YAML parser, and then use the result of that as a template string, so that you could do something bizarre like place `{{- .Values.someKey -}}` _inside_ the initContainer blob and have that then render content from other parts of values.yaml, or loop over something, or whatever.

That seems rather odd without some corresponding change to values.yaml, and also seems like a level of template inception that'd be quite hard to use in practice.

rainest:
Hey @ttony, pinging on this again, had you had a chance to review the previous comment?

ttony:
hi @rainest, sorry for late reply. has been busy with work recently.

1. regarding the commit, i will use feature branch later. not sure why syncing the next branch to my fork branch causing the issue. 
2. for the tpl parts, the reason i am using tpl is to reduce amount of yaml which i can copy and paste. for example, we have few different kube cluster setup within the org -- EKS and Openshfit. each of the clusters contains dev, staging and prod env setup which required separate docker registry to work with. 

with the setup above, example:

common open shift values:
```yaml
kong:
  secretVolumes:
  - ocp-kong-secret
```

dev Openshift for example:
```yaml
global:
   docker_registry: dev.docker.company.registry.com
```

staging:
```yaml
global:
   docker_registry: staging.docker.company.registry.com
```

common EKS values:
```yaml
kong:
  secretVolumes:
  - eks-kong-ssm
```

common for both platform eks and openshift values.yaml:

```yaml
deployment:
  initContainers:
  - name: secret-to-kong-conf
     image: {{ .Values.global.docker_registry }}/agp/kube-base-image:1.0.0
     ....
     ....
     volumeMounts:
     - name: "{{ .Values.secretVolumes | first }}" 
        mountPath: "/etc/kong" 
```

Wihout the tpl magical function, i guess i have to copy and paste all these big chunk of yaml into individual values.yaml  which difficult to maintain in the future as so many repeatable fields.

I  hope that helps to clear out why the reason.

Cheers,

-Y

ttony:
I have re-created a new PR #408

PR #401, opened 2021-07-06
[repo] update workflow branch criteria

### LINKED ISSUE #400: CI: don't allow main merges (without release updates)
While we do have an issue template that tries to avoid main merges, it would be more ideal to have CI that will disallow main merges unless they have completed a series of tasks, including updating the chart version.

Otherwise you can get failures like this: https://github.com/Kong/charts/runs/2999959521?check_suite_focus=true#step:7:14

rainest:
- Run the main workflow on PRs to main.
- Exclude PRs to main from the non-main workflow.

Ensures that the default lint behavior (do not skip the version increment check) runs on PRs to main.

Fix #400 

#### Checklist

### DIFF
--- a/.github/workflows/main-pr.yaml
+++ b/.github/workflows/main-pr.yaml
@@ -1,0 +1,47 @@
+name: Release Charts
+
+on:
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  lint-test:
+    env:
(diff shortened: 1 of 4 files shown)


shaneutt (approved):
Makes sense to me I'm happy when CI is happy :+1:

PR #406, opened 2021-07-14
chore: de-duplicate chart test CI runs

shaneutt:
At the time of writing, the Chart Test CI is run
twice for every PR: once for the PR, and once for
the branch.

This patch makes it so that the chart tests ONLY
run once for each PR, and once for when the PR
is finally merged into next.


rainest (approved):
LGTM, waiting for the new CI run for finish.

Issue #391, opened 2021-06-29
Test that upgrades using the Helm chart work


mflendrich:
Similar to https://github.com/Kong/kubernetes-ingress-controller/issues/1464 but is concerned about the Helm chart path as opposed to the Kustomize path.

Acceptance criteria:
- [x] for Helm chart versions X and Y, the test makes a version X release and then upgrades the release to HEAD and runs a trivial smoke test
- [x] set up CI which automates deployment and upgrade testing all v2 tags from now until release for each chart change
- [x] Helm Chart can deploy v1 CRDs
- [x] test upgrades of CRDs

PR #407, opened 2021-07-15
[kong] add an image compat test for pre-releases

shaneutt:

This patch adds a test intended to help catch regressions
that would make the upcoming KIC 2.0 releases incompatible
with the chart (or visa versa) by ensuring a minimal upgrade
test of the image is performed and the helm tests pass.

Partially resolves https://github.com/Kong/charts/issues/391

#### Checklist

### DIFF
--- a/scripts/test-kic-2.x-upgrade.sh
+++ b/scripts/test-kic-2.x-upgrade.sh
@@ -26,13 +26,6 @@ fi
 set -euo pipefail
 
 # ------------------------------------------------------------------------------
-# Kubernetes Cluster Setup
-# ------------------------------------------------------------------------------
-
-kubectl cluster-info --context kind-${TEST_ENV_NAME} 1>/dev/null
-KUBERNETES_VERSION="$(kubectl version -o json | jq -r '.serverVersion.gitVersion')"
-
-# ------------------------------------------------------------------------------
 # Determine the latest KIC 2.x Pre-Release Version
 # ------------------------------------------------------------------------------
 
@@ -48,47 +41,9 @@ fi
 echo "INFO: the latest release of ${BASE} v2.x was determined to be ${LATEST_PRERELEASE}"
 
 # ------------------------------------------------------------------------------
-# Setup Chart Cleanup - Kubernetes Ingress Controller
+# Run Upgrade Tests
 # ------------------------------------------------------------------------------
 
-RELEASE_NAME="chart-tests-prerelease-compat"
-RELEASE_NAMESPACE="$(uuidgen)"
+echo "INFO: running upgrade test to ${LATEST_PRERELEASE}"
 
-function cleanup() {
-    echo "INFO: cleaning up helm release ${RELEASE_NAME}"
-    helm delete --namespace ${RELEASE_NAMESPACE} ${RELEASE_NAME}
-    kubectl delete namespace ${RELEASE_NAMESPACE}
-    kubectl delete clusterrole ${RELEASE_NAME}-kong
-    exit 1
-}
-
-trap cleanup SIGINT SIGKILL
-
-# ------------------------------------------------------------------------------
-# Deploy Chart - Kubernetes Ingress Controller
-# ------------------------------------------------------------------------------
-
-cd charts/kong/
-echo "INFO: installing chart as release ${RELEASE_NAME} to namespace ${RELEASE_NAMESPACE}"
-helm install --create-namespace --namespace ${RELEASE_NAMESPACE} ${RELEASE_NAME} ./
-
-# ------------------------------------------------------------------------------
-# Test Chart - Kubernetes Ingress Controller
-# ------------------------------------------------------------------------------
-
-echo "INFO: running helm tests for all charts on Kubernetes ${KUBERNETES_VERSION}"
-helm test --namespace ${RELEASE_NAMESPACE} ${RELEASE_NAME}
-
-# ------------------------------------------------------------------------------
-# Upgrade Chart Image To v2.x
-# ------------------------------------------------------------------------------
-
-echo "INFO: upgrading the helm chart to latest pre-release image ${LATEST_PRERELEASE}"
-helm upgrade --namespace ${RELEASE_NAMESPACE} --set ingressController.image.tag=${LATEST_PRERELEASE} ${RELEASE_NAME} ./
-
-# ------------------------------------------------------------------------------
-# Test Upgraded Chart
-# ------------------------------------------------------------------------------
-
-echo "INFO: running helm tests for all charts on Kubernetes ${KUBERNETES_VERSION}"
-helm test --namespace ${RELEASE_NAMESPACE} ${RELEASE_NAME}
+RELEASE_NAME="chart-test-v2-upgrade" TEST_ENV_NAME=${TEST_ENV_NAME} TAG=${LATEST_PRERELEASE} ./scripts/test-upgrade.sh
--- a/scripts/test-kic-2.x-upgrade.sh
+++ b/scripts/test-kic-2.x-upgrade.sh
@@ -1,14 +1,13 @@
 #!/usr/bin/env bash
 
 # ------------------------------------------------------------------------------
-# test-run.sh
+# test-kick-2.x-upgrade.sh
 #
-# This script performs a `helm install` of all charts in the repository and runs
-# the chart tests for each of them.
-#
-# Note: This script assumes you've created an environment with the adjacent
-#       script "test-env.sh" and have a running Kubernetes cluster configured in
-#       your local environment.
+# This script is temporary: in the timeframe between 1.x and 2.x KIC
+# releases this script was made to validate that upgrading to the
+# latest pre-release image was functional, once 2.0 releases fully
+# this script will be removed because all tests thereafter will be
+# running on that version and it's lineage.
 # ------------------------------------------------------------------------------
 
 # ------------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


scripts/test-kic-2.x-upgrade.sh
```diff
@@ -0,0 +1,94 @@
...
+# ------------------------------------------------------------------------------
+
+# ------------------------------------------------------------------------------
+# Environment Variables
+# ------------------------------------------------------------------------------
+
+if [[ -z $TEST_ENV_NAME ]]
+then
+    TEST_ENV_NAME="kong-charts-tests"
+fi
```
mflendrich:
```suggestion
TEST_ENV_NAME="${TEST_ENV_NAME:-kong-charts-tests}"
```

scripts/test-kic-2.x-upgrade.sh
```diff
@@ -0,0 +1,94 @@
...
+# Shell Configuration
+# ------------------------------------------------------------------------------
+
+set -euo pipefail
+
+# ------------------------------------------------------------------------------
+# Kubernetes Cluster Setup
+# ------------------------------------------------------------------------------
+
+kubectl cluster-info --context kind-${TEST_ENV_NAME} 1>/dev/null
```
mflendrich: Please run shellcheck on this file and fix all errors. Here it should complain about the variable expansion not being quoted.

scripts/test-kic-2.x-upgrade.sh
```diff
@@ -0,0 +1,94 @@
...
+KUBERNETES_VERSION="$(kubectl version -o json | jq -r '.serverVersion.gitVersion')"
```
mflendrich: nit: use `-o jsonpath=...`

scripts/test-kic-2.x-upgrade.sh
```diff
@@ -0,0 +1,94 @@
...
+
+# ------------------------------------------------------------------------------
+# Determine the latest KIC 2.x Pre-Release Version
+# ------------------------------------------------------------------------------
+
+BASE="kong/kubernetes-ingress-controller"
+LATEST_PRERELEASE=$(curl -s https://api.github.com/repos/${BASE}/releases | \
```
mflendrich, line 30: Could we just run against the `next-railgun` tag instead of trying to guess that latest pre-release here?

That will give us earlier feedback, without the need to wait for the pre-release to happen.

This suggestion boils down to redefining this test as "please test the upgrade from version X to version Y" where the default X is what's set in the chart, and the default Y is `railgun-next`.

scripts/test-kic-2.x-upgrade.sh
```diff
@@ -0,0 +1,94 @@
...
+# ------------------------------------------------------------------------------
+
+RELEASE_NAME="chart-tests-prerelease-compat"
+RELEASE_NAMESPACE="$(uuidgen)"
+
+function cleanup() {
+    echo "INFO: cleaning up helm release ${RELEASE_NAME}"
+    helm delete --namespace ${RELEASE_NAMESPACE} ${RELEASE_NAME}
+    kubectl delete namespace ${RELEASE_NAMESPACE}
+    kubectl delete clusterrole ${RELEASE_NAME}-kong
```
mflendrich: This should be a helm post-delete hook instead, no?

rainest: I don't think so--could you clarify why you think it should? Helm hooks apply to things the chart does when you actually install it, not to stuff specific to test envs.

I think the clusterrole delete here is probably superfluous--`helm delete` should normally purge that since it's a chart-owned resource. The only things you should need to clean up independent of Helm are the namespace and CRDs (which Helm will create, but not delete).

scripts/test-kic-2.x-upgrade.sh
rainest: :+1: and we may also want to consider moving this to a cron trigger rather than a push trigger: KIC will get continuous pre-release (in the "it's committed to next" sense, not in the "it's a released alpha/beta image" sense) changes independent of the chart, and testing the chart against next will call out when we make changes that the chart must account for. We probably want to observe those as they happen in KIC, rather than waiting for a (probably unrelated) push to the chart to trigger it.

For specific KIC versions, we can probably test upgrades via `ct install` tests, which support a `--upgrade` mode to compare a previous chart version to a proposed version. That proposed version can include a values.yaml with a version bump for images, and testing that upgrade will thus test an image upgrade.

`ct` version upgrade testing is static (it is always the exact image versions you have in the previous and proposed values.yaml), but that's probably what we want normally. The KIC 1.x -> 2.x transition is unique and we probably want to test that migration as a one-off thing outside, whereas CI will be concerned with more routine "default minor version upgrade succeeds" checks that we'll have ongoing going forward.

rainest (commented):
Commentary and planning review: there's probably enough historical cruft and inbound changes here to warrant a sync discussion, so I've added some comments, will mull chart test things over the weekend, and have tentatively planned time to discuss on 7/21.

rainest:
Point that came up in today's sync meeting: as part of the 1.x -> 2.x transition, we now use generated CRDs from kubebuilder which have differences from the original handwritten CRDs. Helm [will not apply CRD updates](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-1-let-helm-do-it-for-you) after the initial install; you must apply these manually. To complete the upgrade test, the script must also apply CRDs outside of [its `helm upgrade` invocation](https://github.com/Kong/charts/pull/407/files#diff-66ae022f28c059f599baf04d70d8b76169a095d426ed12b4137f0aadec197ac2R87).

That should presumably happen before `helm upgrade`, though I'm not 100% sure--there's something of a chicken/egg problem here where installing new CRDs with 1.x still running may mean it can no longer parse them properly.

This does currently pass because 2.x is compatible with the 1.x CRDs so we can merge this as-is (with Michal's change requests) and add CRD migration tests to this once (in a follow-up PR) https://github.com/Kong/kubernetes-ingress-controller/issues/1133 is done.

rainest (approved):

scripts/test-kic-2.x-upgrade.sh
shaneutt: resolved across all scripts as per 4417aa3d2862c1641395cbffeb8133504ea1e8bf

scripts/test-upgrade.sh
```diff
@@ -0,0 +1,76 @@
...
+# ------------------------------------------------------------------------------
+
+# ------------------------------------------------------------------------------
+# Environment Variables
+# ------------------------------------------------------------------------------
+
+TEST_ENV_NAME="${TEST_ENV_NAME:-kong-charts-tests}"
+TAG="${TAG:-next-railgun}"
+RELEASE_NAME="${RELEASE_NAME:-chart-tests-upgrade-compat}"
+RELEASE_NAMESPACE="${RELEASE_NAMESPACE:-$(uuidgen)}"
```
ccfishk, line 20: do we even need release_namespace ?

scripts/test-upgrade.sh
```diff
@@ -0,0 +1,76 @@
...
+KUBERNETES_VERSION="$(kubectl version -o json | jq -r '.serverVersion.gitVersion')"
+
+# ------------------------------------------------------------------------------
+# Setup Chart Cleanup - Kubernetes Ingress Controller
+# ------------------------------------------------------------------------------
+
+function cleanup() {
+    echo "INFO: cleaning up helm release ${RELEASE_NAME}"
+    helm delete --namespace ${RELEASE_NAMESPACE} ${RELEASE_NAME}
+    kubectl delete namespace ${RELEASE_NAMESPACE}
```
ccfishk: Long-running cluster takes forever removing namespace, you might want to ensure namespace is cleaned before next step.

scripts/test-upgrade.sh
```diff
@@ -0,0 +1,76 @@
...
+    kubectl delete clusterrole ${RELEASE_NAME}-kong
```
ccfishk: A more professional way is putting the resource in cleanup.yaml and run tool "kubectl delete -f cleanup.yaml"

scripts/test-upgrade.sh
```diff
@@ -0,0 +1,76 @@
...
+cd charts/kong/
+echo "INFO: installing chart as release ${RELEASE_NAME} to namespace ${RELEASE_NAMESPACE}"
+helm install --create-namespace --namespace ${RELEASE_NAMESPACE} ${RELEASE_NAME} ./
+
+# ------------------------------------------------------------------------------
+# Test Chart - Kubernetes Ingress Controller
+# ------------------------------------------------------------------------------
+
+echo "INFO: running helm tests for all charts on Kubernetes ${KUBERNETES_VERSION}"
+helm test --namespace ${RELEASE_NAMESPACE} ${RELEASE_NAME}
```
ccfishk: No success/failure checkpoint ?

scripts/test-kic-2.x-upgrade.sh
shaneutt: 0caae597af65de6708c83a27f8b5a2c0a24687f0 resolves this, but I did you one better and make CI checks for this in 3409c9ddc068d7c3ccc956d15a00991287c4a862 :+1:

scripts/test-kic-2.x-upgrade.sh
shaneutt: added in ab26afcebebe82f212918915b0814616800d408b :+1:

scripts/test-kic-2.x-upgrade.sh
shaneutt: These all run as cleanup if the script is interrupted, so if they were in a hook they would not have the same guarantees to execute.

scripts/test-upgrade.sh
shaneutt: I do ensure namespace is cleaned up by running the `helm delete` above.

scripts/test-upgrade.sh
shaneutt: I don't understand how this is better, or "more professional". We don't create this resource so we only know of its existence by name. Perhaps I should add some further comments to the cleanup?

scripts/test-upgrade.sh
shaneutt: the success/failure is already determined via exit code and the enforcement of `-o pipefail` above (any failure of a subcommand will fail the entire thing).

scripts/test-kic-2.x-upgrade.sh
shaneutt: unfortunately the `version` subcommand doesn't support the same inspection flags as the `get` subcommand does, or yes I would have added this (or in my case, `-o gotemplate=='<TEMPLATE'` of which I am most comfortable with).

scripts/test-upgrade.sh
ccfishk: ? if so, why use duplicate command.

scripts/test-upgrade.sh
ccfishk: From what you implemented, yo do know which resources you need cleanup -- put them in a central place and execute one command.

scripts/test-upgrade.sh
ccfishk: sorry, this is a test file, generically check each test step. Does it make sense ?

scripts/test-kic-2.x-upgrade.sh
mflendrich: Ah, my bad.

scripts/test-kic-2.x-upgrade.sh
mflendrich: @rainest my thinking has been that, if the chart release applies the clusterrole, then it should be responsible for cleaning it up. Passing this responsibility over to the caller (here: the test script) is a sign of a leaky abstraction.

What am I missing?

scripts/test-kic-2.x-upgrade.sh
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
+
+# ------------------------------------------------------------------------------
+# test-kick-2.x-upgrade.sh
```
mflendrich:
```suggestion
# test-kic-2.x-upgrade.sh
```

mflendrich (approved):
one typo + lgtm

PR #408, opened 2021-07-20
[kong] initContainer for job #390

### LINKED ISSUE #390: initContainer for job

ttony:
  - fixes #390 

#### Checklist

### DIFF
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -369,7 +369,7 @@ The name of the service used for the ingress controller's validation webhook
 
 {{- define "kong.userDefinedVolumes" -}}
 {{- if .Values.deployment.userDefinedVolumes }}
-{{- tpl (toYaml .Values.deployment.userDefinedVolumes) . }}
+{{- toYaml .Values.deployment.userDefinedVolumes }}
 {{- end }}
 {{- end -}}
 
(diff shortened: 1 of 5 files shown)


rainest (changes requested):
Gotcha re the [`tpl` use description in the previous PR](https://github.com/Kong/charts/pull/397#issuecomment-883036453). I'd still request removing that and using plain `toYaml` only. The meta-templating enabled by using content elsewhere in values.yaml as an injected user template feels overcomplicated, so absent demand from additional users I'd like to avoid it.

As an alternative I'd suggest pre-processing values.yamls with another templating engine to fill in those blocks prior to passing it to Helm. That allows for the same end result and, as a general strategy, is more flexible, as it doesn't rely on specific parts of the chart template using `tpl`.

https://github.com/noqcks/gucci can be used as a standalone template renderer that supports the same capabilities as Helm's standard template engine (since it's Golang templating with the sprig library included).

ttony:
for tpl related, i will removed it soon. aside of this, we can ask for vote on the community about tpl feature which supported in helm. as it is very handy and bundled function as part of helm instead of rely of another different tools.

rainest (approved):

rainest:
> aside of this, we can ask for vote on the community about tpl feature

Yeah, feel free to create an issue outlining the proposed change, and we'll await input from other users looking for something similar and/or note down when it'd be useful for handling some other need.

PR #415, opened 2021-07-22
[kong] v1 CRDs and tests for CRDs

shaneutt:

- Upgrades CRDs from deprecated v1beta1 to v1 CRD API
- Adds tests for CRDs, including testing against legacy CRDs, and testing upgrading from legacy to V1 CRDs.

Partially resolves https://github.com/Kong/charts/issues/391

#### Checklist


charts/kong/crds/custom-resource-definitions.yaml
```diff
@@ -1,425 +1,991 @@
+# generated using: kubectl kustomize github.com/kong/kubernetes-ingress-controller/railgun/config/crd?ref=main
```
mflendrich, line 1: This is a big fat breaking change that must have a huge warning in the release notes. We're effectively dropping support for k8s versions too old for v1.

scripts/test-crds.sh
```diff
@@ -0,0 +1,130 @@
+#!/usr/bin/env bash
+
+# ------------------------------------------------------------------------------
+# test-crds.sh
+#
+# This script tests that our legacy v1beta1 versions of our CRDs upgrade cleanly
+# to their v1 versions, and that both the legacy and the v1 versions of the CRDs
+# both function with our controllers.
+#
+# Test Workflow
```
mflendrich: (non-blocking but this concerns me)

imo elaborate descriptions "how the code works" at the top of a file like this don't age well. Do you think that we could achieve a similar degree of documentation by means of self-documenting code or local comments?

scripts/test-crds.sh
```diff
@@ -0,0 +1,130 @@
...
+#   2. deploy the v1 CRDs
+#   3. verify the CRDs updates
+#   4. redeploy the chart
+#   5. run helm tests (via v1 CRDs)
+#
+# NOTE: these tests are sensitive to which Kubernetes version they are run on.
+#       they will only run on Kubernetes clusters between v1.16.x and 1.21.x,
+#       as v1.15.x is too old for the new CRDs and v1.22.x is to new for the
+#       old CRDs. At some point in the future (potentially when <v1.22.x is no
+#       longer supported by the chart) we can remove these tests.
```
mflendrich: please be more specific: "this test requires k8s between 1.16 and 1.21 inclusively because it requires support for both v1beta1 and v1 CRD apis"

scripts/test-crds.sh
```diff
@@ -0,0 +1,130 @@
...
+# ------------------------------------------------------------------------------
+
+# ------------------------------------------------------------------------------
```
mflendrich, line 17: not blocking - opinionated code style rant

I'm personally not a fan of dividing code into chunks by comment blocks. I'd say: modularize the code "properly" using functions, and make comments comments and not concern "separators"

scripts/test-crds.sh
```diff
@@ -0,0 +1,130 @@
...
+
+RELEASE_NAME="chart-tests"
+RELEASE_NAMESPACE="$(uuidgen)"
+
+function cleanup() {
+    echo "INFO: cleaning up helm release ${RELEASE_NAME}"
+    helm delete --namespace "${RELEASE_NAMESPACE}" "${RELEASE_NAME}"
+    kubectl delete namespace "${RELEASE_NAMESPACE}"
+    kubectl delete clusterrole "${RELEASE_NAME}"-kong
+    exit 1
```
mflendrich: Why does this exit 1 need to be here?
```suggestion
```

scripts/test-crds.sh
```diff
@@ -0,0 +1,130 @@
...
+}
+
+trap cleanup SIGINT
```
mflendrich: why not trap on exit, instead of specific signals?

mflendrich (changes requested):
requested changes because of the action required

charts/kong/crds/custom-resource-definitions.yaml
shaneutt: I discussed this with @rainest before doing it, and yes it is breaking (for old kubernetes clusters).

Had to break the mold here a bit since we normally don't write the CHANGELOG in this repo until release, but dec2b4866fbc5d531b550a60bce1015a8b52a170 adds the required notices.

scripts/test-crds.sh
shaneutt: There's a lot of "INFO" output and commented sections which I feel do a good job of explaining the code. This was just meant as a legend so you don't have the scan the whole thing, but I can just as easily pull it and people can read the code for themselves.

scripts/test-crds.sh
shaneutt:
```suggestion
# NOTE: these tests are sensitive to which Kubernetes version they are run on.
#       they will only run on Kubernetes clusters between v1.16.x and 1.21.x
#       (inclusively) as v1.15.x is too old for the new CRDs and v1.22.x is to new
#       for the old CRDs. At some point in the future (potentially when <v1.22.x is
#       no longer supported by the chart) we can remove these tests.
```

scripts/test-crds.sh
shaneutt: I am a fan of these, but I also created a follow up issue to turn some of the repeat functionality into sourceable functions:

https://github.com/Kong/charts/issues/416

Since you consider this non-blocking, I'm going to resolve this as it matches the style of what exists around it already. If you still feel strongly about the style going forward we should put together an issue to add style guide documentation and perhaps linters to enforce them.

scripts/test-crds.sh
shaneutt: The documentation for `trap` states that it will produce an exit status based on the exit status of the function:

```
       builtin shell-builtin [arguments]
              Execute  the specified shell builtin, passing it arguments, and return its
              exit status.  This is useful when defining a function whose  name  is  the
              same as a shell builtin, retaining the functionality of the builtin within
              the function.  The cd builtin is commonly redefined this way.  The  return
              status is false if shell-builtin is not a shell builtin command.
```

Thus, having this exit ensures we have a specific error status that will be returned by trap when cleanup is performed due to an interrupt.

scripts/test-crds.sh
shaneutt: That would run cleanup under normal script exit. We could make that work, but really this trap is _only_ here to cleanup in the event that something fails mid-script.

scripts/test-crds.sh
shaneutt: removed in 0eac9952ed92ff1fb26bec4609bdf183e595fc6b

scripts/test-crds.sh
shaneutt: Pursuant to another comment, these cleanups are entirely removed now as of f3968d43f00feae14859224d9f9be514d4348dd0

scripts/test-crds.sh
shaneutt: These cleanups were helpful for debugging/development of the scripts but ultimately there's a script entirely for cleanup so these are now removed in f3968d43f00feae14859224d9f9be514d4348dd0.

scripts/test-crds.sh
```diff
@@ -0,0 +1,92 @@
...
+
+# ------------------------------------------------------------------------------
+# Configuration
+# ------------------------------------------------------------------------------
+
+RELEASE_NAME="${RELEASE_NAME:-chart-tests-crds}"
+RELEASE_NAMESPACE="${RELEASE_NAMESPACE:-$(uuidgen)}"
+TEST_ENV_NAME="${TEST_ENV_NAME:-kong-charts-tests}"
+
+set -euo pipefail
```
mflendrich: this should probably be above anything else.

`uuidgen` on line 22 can fail silently otherwise

charts/kong/CHANGELOG.md
```diff
@@ -1,5 +1,25 @@
...
+* Upgraded CRDs to V1 from the previous deprecated v1beta1.
+  [#391](https://github.com/kong/charts/issues/391)
+  ACTION REQUIRED: This is a breaking change as it makes
+  this chart incompatible with Kubernetes clusters older
+  than v1.16.x. Upgrade your cluster to a version greater
+  than or equal to v1.16 before installing.
+  Note that technically it will remain possible to deploy
+  on older clusters by managing the CRDs manually ahead of
+  time (e.g. intentionally deploying the legacy CRDs) but
+  these configurations will be considered unsupported.
```
mflendrich, line 16:
```suggestion
```

this is a challenging commitment imo - we can't afford to accept the commitment to support unsupported versions "if you just do X and Y"

scripts/test-crds.sh
```diff
@@ -0,0 +1,92 @@
...
+
+kubectl cluster-info --context kind-"${TEST_ENV_NAME}" 1>/dev/null
```
mflendrich: what is the purpose of this command if we drop the output?

scripts/test-crds.sh
```diff
@@ -0,0 +1,92 @@
...
+KUBERNETES_VERSION="$(kubectl version -o json | jq -r '.serverVersion.gitVersion')"
```
mflendrich: shouldn't we be using the context name in all kubectl commands?

suggestion replace kubectl invocations with `$KUBECTL` defined as:
`KUBECTL="kubectl --context kind-..."`

scripts/test-crds.sh
```diff
@@ -0,0 +1,92 @@
...
+
+echo "INFO: running CRD tests on Kubernetes ${KUBERNETES_VERSION}"
+
+# ------------------------------------------------------------------------------
+# Cleanup Existing CRDs and Deploy Legacy CRDs
+# ------------------------------------------------------------------------------
+
+echo "INFO: cleaning up existing CRDs from the test cluster"
+for CRD in $(kubectl get crds -o=go-template='{{range .items}}{{.metadata.name}}{{"\n"}}{{end}}' |grep 'konghq.com')
```
mflendrich:
```suggestion
for CRD in $(kubectl get crds -o=go-template='{{range .items}}{{.metadata.name}}{{"\n"}}{{end}}' |grep '\.konghq\.com$')
```

so that we don't match

```
something-special.for.konghq.com.acme.org
konghqxcomx.potato.com
```

scripts/test-crds.sh
```diff
@@ -0,0 +1,92 @@
...
+do
+    kubectl delete crd "${CRD}"
```
mflendrich: nit (optional): bind the output of the `get crds` above to a variable, and pass them all to a single `kubectl delete crd <list of names>`

benefits: single finalization barrier and fewer forks, achieving the same result

scripts/test-crds.sh
```diff
@@ -0,0 +1,92 @@
...
+done
+
+LAST_CRD_UPDATE="d2442de89cc036843c5f552c7daa6b023405f365"
```
mflendrich: what is this magic number? does it refer to a specific commit? under what circumstances can this value change? if it should change, how to update it?

answers to these questions shouldn't go to a reply to this comment, but to code comments above this line

maybe we should be using a git tag or a git branch instead of a raw commit hash?

scripts/test-crds.sh
```diff
@@ -0,0 +1,92 @@
...
+kubectl apply -f "https://raw.githubusercontent.com/Kong/kubernetes-ingress-controller/${LAST_CRD_UPDATE}/deploy/manifests/base/custom-types.yaml"
+
+# ------------------------------------------------------------------------------
+# Deploy Chart - Kubernetes Ingress Controller
+# ------------------------------------------------------------------------------
+
+cd charts/kong/
```
mflendrich: implicit assumption that this script will be run with CWD set to the root of the repo. 

1. Please consider using this pattern to ensure the CWD at the beginning of the script: https://stackoverflow.com/a/59916
2. Instead of `cd`, consider replacing the `./` in the `helm install` invocation with a `./charts/kong/` to avoid unnecessarily altering the global state of the script with a `cd`.

scripts/test-run.sh
```diff
@@ -12,16 +12,14 @@
...
+RELEASE_NAME="${RELEASE_NAME:-chart-tests}"
+RELEASE_NAMESPACE="${RELEASE_NAMESPACE:-$(uuidgen)}"
 TEST_ENV_NAME="${TEST_ENV_NAME:-kong-charts-tests}"
 TAG="${TAG:-default}"
 
-# ------------------------------------------------------------------------------
-# Shell Configuration
-# ------------------------------------------------------------------------------
-
 set -euo pipefail
```
mflendrich, line 18: ditto - this needs to happen before any other command invocations (`uuidgen` in this case)

scripts/test-upgrade.sh
```diff
@@ -11,41 +11,19 @@
...
 RELEASE_NAME="${RELEASE_NAME:-chart-tests-upgrade-compat}"
 RELEASE_NAMESPACE="${RELEASE_NAMESPACE:-$(uuidgen)}"
-
-# ------------------------------------------------------------------------------
-# Shell Configuration
-# ------------------------------------------------------------------------------
+TEST_ENV_NAME="${TEST_ENV_NAME:-kong-charts-tests}"
+TAG="${TAG:-next-railgun}"
 
 set -euo pipefail
```
mflendrich: ditto

scripts/test-crds.sh
shaneutt: ah good catch eeecd15f23143944bb98ee162b8ec4cf53a6f6e6

scripts/test-crds.sh
shaneutt: It changes the current context, and even though we don't care about the output `pipefail` will cascade any failure to the test.

shaneutt: It's a leftbehind, it was just testing that the context exists, but given your below comments we can drop it if we use the context directly.

scripts/test-crds.sh
shaneutt: Yep we can do that.

scripts/test-crds.sh
shaneutt: Yep sounds good

scripts/test-crds.sh
shaneutt: Sounds good, I've opted for tag 1.3.1 instead

scripts/test-run.sh
shaneutt: :+1: eeecd15f23143944bb98ee162b8ec4cf53a6f6e6

scripts/test-upgrade.sh
shaneutt: :+1: eeecd15f23143944bb98ee162b8ec4cf53a6f6e6

scripts/test-crds.sh
shaneutt: :+1: 4fdf72c8f6880d9704a243dcb2680a4c78c09db5

shaneutt: resolved by 4fdf72c8f6880d9704a243dcb2680a4c78c09db5

scripts/test-crds.sh
shaneutt: resolved by 4fdf72c8f6880d9704a243dcb2680a4c78c09db5

scripts/test-crds.sh
shaneutt: 1d7bafe13a2ec3ead8b3fcf1d7a675dc32db392d

scripts/test-crds.sh
shaneutt: 1d7bafe13a2ec3ead8b3fcf1d7a675dc32db392d

scripts/test-crds.sh
shaneutt: 43af719e3fd00ffa0349a039829ba4e6defee73b

shaneutt: and d63d5eb836629e7ccb1a9317944772fe76361774

mflendrich (approved):
:100:

PR #419, opened 2021-07-23
[kong] update RBAC for KIC v2

rainest:
Updates the RBAC role permissions for KIC v2. Supersedes #364. New PR since the upstream permissions have changed a ton since #364 started and because I used a different update strategy.

Added the v2 leader election permissions alongside the v1 permissions in the same role, as v2 uses a new leader election system built into controller-runtime. There is some overlap, but trying to consolidate them in the chart will make any future updates harder. If they remain separated, you can just replace whichever block changed with the new version from upstream.

Replaced the entire ClusterRole permission set (which granted access to Ingresses and such) from v1 with the v2 permissions. The v2 permissions are a superset of the v1 permissions, so there doesn't appear to be any reason to maintain separate versions. Note that the v2 permissions _do_ have a slightly different structure. They do not group resources in the same API group and equal permission sets together; see the example at https://github.com/Kong/charts/pull/364#issuecomment-885324557

Related to Kong/kubernetes-ingress-controller#1352

This includes the change from https://github.com/Kong/kubernetes-ingress-controller/pull/1584. That should be merged imminently.

This does _not_ include permissions necessary to add finalizers. KIC 2.0.0-alpha.2 still attempts to add those, and will fail without them. We have removed finalizers from next but have not yet released alpha.3.

#### Checklist

### DIFF
--- a/charts/kong/templates/controller-rbac-resources.yaml
+++ b/charts/kong/templates/controller-rbac-resources.yaml
@@ -41,6 +41,28 @@ rules:
       - endpoints
     verbs:
       - get
+  # Begin KIC 2.x leader permissions
+  - apiGroups:
+      - ""
+      - coordination.k8s.io
+    resources:
+      - configmaps
+      - leases
+    verbs:
+      - get
+      - list
+      - watch
+      - create
+      - update
+      - patch
+      - delete
+  - apiGroups:
+      - ""
+    resources:
+      - events
+    verbs:
+      - create
+      - patch
 ---
 apiVersion: rbac.authorization.k8s.io/v1beta1
 kind: RoleBinding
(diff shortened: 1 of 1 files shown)


shaneutt (approved):
Looks good to me :+1: 

My only thought is: what are we doing about testing this?

rainest:
Was good aside from the part where I opened a copy of the role from the source tree where I didn't have KIC#1584 checked out, whoops.

Testing-wise we won't have have anything consistent until the functional tests are expanded to include all resources the controller can interact with (versus just Ingress, Service, and Endpoint so far with the basic httpbin test). Absent that it's our old friend "upgrade a release that has a bunch of config and see if it comes online with the expected config/check error logs", which worked, implies that it was able to read resources, and caught the still-inaccurate Ingress v1beta1 class.

PR #420, opened 2021-07-27
[kong] add watch-namespace automation

rainest:
Automates role creation for controllers that watch a specific set of namespaces only. When `ingressController.watchNamespaces` is set to an empty array, watches all namespaces and uses a ClusterRole to access resources, same as we've done in the past. When it is set to a list of namespaces, specifies them in `--watch-namespace` and creates a Role/RoleBinding for each individual namespace.

Related to https://github.com/Kong/kubernetes-ingress-controller/issues/1503

#### Checklist

### DIFF
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -346,6 +346,9 @@ The name of the service used for the ingress controller's validation webhook
 {{- if .Values.ingressController.admissionWebhook.enabled }}
   {{- $_ := set $autoEnv "CONTROLLER_ADMISSION_WEBHOOK_LISTEN" (printf "0.0.0.0:%d" (int64 .Values.ingressController.admissionWebhook.port)) -}}
 {{- end }}
+{{- if (not (eq (len .Values.ingressController.watchNamespaces) 0)) }}
+  {{- $_ := set $autoEnv "CONTROLLER_WATCH_NAMESPACE" (.Values.ingressController.watchNamespaces | join ",") -}}
+{{- end }}
 
 {{/*
     ====== USER-SET ENVIRONMENT VARIABLES ======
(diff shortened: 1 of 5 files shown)


shaneutt (approved):
No blockers, but the failing CI tests seem to indicate we've broken something?

rainest:
I think I failed to heed my own advice: the automation _does_ require KIC 2.x, but we don't test with that yet. CI for the new feature would require a 2.x instance to run successfully, so until then we can only test the default, which uses existing functionality with a slightly changed template.

Will probably just remove that part of the change. Poking at it a bit since it failed in a weird way (the namespace was just completely empty, which is... odd), and I should do a manual test that does use 2.x.

Q: is it feasible to automate the manual tests ?

shaneutt:
> I think I failed to heed my own advice: the automation _does_ require KIC 2.x, but we don't test with that yet. CI for the new feature would require a 2.x instance to run successfully, so until then we can only test the default, which uses existing functionality with a slightly changed template.
> 
> Will probably just remove that part of the change. Poking at it a bit since it failed in a weird way (the namespace was just completely empty, which is... odd), and I should do a manual test that does use 2.x.

Sounds good. Once you have done manual testing, and have some corrections for the tests I'm still :heavy_check_mark:

rainest:
Post manual testing updates in a096df3. Requesting review again since the changes are large enough to warrant another look IMO.

2.x will now start with a limited set of namespaces without permissions errors.

We actually won't be able to completely test watchNamespaces even with 2.x, since you can't use namespaces that don't exist and can't create new namespaces for CI. The workaround in https://github.com/Kong/charts/issues/421 is the only option. Still gets most of what we want to test (templates are correct and the controller doesn't break completely with that option) even though it's not a realistic configuration.

RBAC version bump wasn't necessary for this, but we should do it anyway because v1beta1 is deprecated, and we intended to change it per https://github.com/Kong/kubernetes-ingress-controller/issues/801. That's handled upstream but wasn't handled here because I only overwrote rules when making these changes.

charts/kong/templates/controller-rbac-resources.yaml
```diff
@@ -63,11 +63,18 @@ rules:
...
+  - apiGroups:
+      - ""
+    resources:
+      - services
+      - endpoints
+    verbs:
+      - get
 ---
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
```
shaneutt, line 74: :+1:

charts/kong/templates/_helpers.tpl
```diff
@@ -346,6 +346,10 @@ The name of the service used for the ingress controller's validation webhook
 {{- if .Values.ingressController.admissionWebhook.enabled }}
   {{- $_ := set $autoEnv "CONTROLLER_ADMISSION_WEBHOOK_LISTEN" (printf "0.0.0.0:%d" (int64 .Values.ingressController.admissionWebhook.port)) -}}
 {{- end }}
+{{- if (not (eq (len .Values.ingressController.watchNamespaces) 0)) }}
+  {{- $_ := set $autoEnv "CONTROLLER_WATCH_NAMESPACE" (.Values.ingressController.watchNamespaces | join ",") -}}
+  {{- $_ := set $autoEnv "CONTROLLER_CONTROLLER_KONGCLUSTERPLUGIN" "disabled" -}}
```
shaneutt, line 351: I'm _kinda_ lost as to how this one is relevant to the watchNamespaces,  did I miss something?

charts/kong/templates/controller-rbac-resources.yaml
```diff
@@ -1,8 +1,8 @@
 {{- if and .Values.ingressController.rbac.create .Values.ingressController.enabled -}}
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
```
shaneutt, line 2: :+1:

charts/kong/templates/controller-rbac-resources.yaml
shaneutt, line 74: :+1:

shaneutt (approved):
LGTM :+1: 

Just one comment about cluster plugin, and then when that's resolved and CI is happy, I'm happy :vulcan_salute:

charts/kong/templates/_helpers.tpl
rainest: When you watch specific namespaces only, you cannot watch any cluster scope resources. If you enable the KCP controller, it attempts to watch cluster-scoped KCPs unsuccessfully.

shaneutt: ooop ofc, ty.

Issue #386, opened 2021-06-28
Kong migration job args extension


ttony:
Hi Kong Team,

I would like to make ENV var KONG_PASSWORD only been set or used in kong migration only so that the generated yaml is clean, since KONG_PASSWORD is only applicable during the boostrap migration job only.

Also, I would like to try to highlight about the one of the best practice, such as CKV_K8S_35 (Prefer using secrets as files over secrets as environment variables).  

The generated bootstrap job yaml is something like this:

```yaml
  containers:
  - args: [ "/bin/sh", "-c", "export KONG_PASSWORD=`cat /etc/secrets/kong/password` && kong migrations bootstrap -vv" ]
```

we can use it as default values for it.

```yaml
migrations:
     additionalArgs: ""
     additionalPreUpgradeArgs: ""
     additionalPostUpgradeArgs: "" 
```

migration template is something like this:

```yaml
  containers:
  - args: [ "/bin/sh", "-c", "{{ .Values.migrations.additionalArgs }} kong migrations bootstrap" ]
```

pre-upgrade migration template:
```yaml
  containers:
  - args: [ "/bin/sh", "-c", "{{ .Values.migrations.additionalPreUpgradeArgs }} kong migrations up" ]
```

post-upgrade migration template:
```yaml
  containers:
  - args: [ "/bin/sh", "-c", "{{ .Values.migrations.additionalPostUpgradeArgs }} kong migrations finish" ]
```

Let me know your feedback

rainest:
Do you have further information on the CKV_K8S_35 recommendation? I'm not able to find much anything on it other than Bridgecrew's reference pages and blog, and the rationale there (that environment variables may show up in application logs) seems quite limited, and not applicable here (we do not log that value and intentionally mark sensitive values for redaction in our startup logs). Were this a widely accepted recommendation, I'd expect to see it in official Kubernetes documentation, e.g. https://kubernetes.io/docs/tasks/inject-data-application/distribute-credentials-secure/ and https://kubernetes.io/docs/concepts/configuration/secret/#security-properties

Injecting user-provided free text into the migrations commands in the chart is a rather roundabout way to address this in either case. I'd recommend instead:
- Continue using the environment variable, but remove it from values.yaml and running `helm upgrade` after bootstrapping to clear it from the Deployment and subsequent Jobs.
- If you do indeed want to load it from a file mount instead, submit a feature request for Kong Enterprise upstream to request alternate RBAC bootstrap options.

[...]

PR #429, opened 2021-08-03
[kong] use migration commands directly in args

rainest:
Call `kong` commands directly from migration pods. This allows them to take advantage of special entrypoint script features that require a `kong` command in args.

Related to https://github.com/Kong/charts/issues/386#issuecomment-891859681

Note that the functionality in question doesn't yet exist in the entrypoint for the `KONG_PASSWORD` variable. This is preparation assuming that will be implemented.

Holding pending CI confirmation that Postgres tests complete successfully.

IIRC there may have been some reason we used the `sh` arg originally--typically we did that specifically to _avoid_ the entrypoint script because of some incompatibility between its behavior and the chart (usually related to daemonization). Not sure if that ever actually existed though and/or if it's something that affected older versions of, but not the current entrypoint script.

#### Checklist

### DIFF
--- a/charts/kong/templates/migrations-post-upgrade.yaml
+++ b/charts/kong/templates/migrations-post-upgrade.yaml
@@ -57,7 +57,7 @@ spec:
         {{ toYaml .Values.containerSecurityContext | nindent 10 }} 
         env:
         {{- include "kong.no_daemon_env" . | nindent 8 }}
-        args: [ "/bin/sh", "-c", "kong migrations finish" ]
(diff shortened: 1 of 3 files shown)


shaneutt (approved):
LGTM and thanks for the thorough description :+1:

rainest:
Fine with CI. Tested with something ancient (Enterprise 1.5) and it doesn't appear this hits any incompatibility with older images, at least for the extent of older that I reasonably care about. Migrations completed and the main instance came online. :shipit:

PR #430, opened 2021-08-04
[kong] add controller to service monitor

rainest:
Adds the controller metrics port to the controller container template and Prometheus ServiceMonitor template. The port only appears if the controller's major version is >= 2.

Part of https://github.com/Kong/kubernetes-ingress-controller/issues/1497

This change imposes new requirements on values. The controller image tag _must_ be semver-compliant. The template will completely fail to render if it is not.

This does not pose issues for our stock image tags, but may affect users who use custom controller images. If they do not currently use semver tags, they must make them semver compliant. IMO this is a reasonable ask: they can use the base image version and include any additional info relevant to their build using semver metadata.

The helper output is annoyingly considered a string, with no available function to convert it to a boolean. Go templates are cumbersome.

#### Checklist

### DIFF
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -535,7 +535,17 @@ The name of the service used for the ingress controller's validation webhook
   {{- toYaml .Values.resources | nindent 4 }}
 {{- end -}}
 
+{{/*
+kong.controller2xplus returns "true" if the controller image tag major version is >= 2
+Note that this is a string, not a boolean, because templates vov
+*/}}
+{{- define "kong.controller2xplus" -}}
+{{- $version := semver .Values.ingressController.image.tag -}}
+{{- ge $version.Major 2 -}}
+{{- end -}}
+
 {{- define "kong.controller-container" -}}
+{{- $controllerIs2xPlus := include "kong.controller2xplus" . -}}
 - name: ingress-controller
   securityContext:
 {{ toYaml .Values.containerSecurityContext | nindent 4 }}  
@@ -550,6 +560,11 @@ The name of the service used for the ingress controller's validation webhook
   {{- if .Values.ingressController.admissionWebhook.enabled }}
   - name: webhook
     containerPort: {{ .Values.ingressController.admissionWebhook.port }}
+    protocol: TCP
+  {{- end }}
+  {{ if (eq $controllerIs2xPlus "true") -}}
+  - name: cmetrics
+    containerPort: 10255
     protocol: TCP
   {{- end }}
   env:
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -540,7 +540,12 @@ kong.controller2xplus returns "true" if the controller image tag major version i
 Note that this is a string, not a boolean, because templates vov
 */}}
 {{- define "kong.controller2xplus" -}}
-{{- $version := semver .Values.ingressController.image.tag -}}
+{{- $version := "" -}}
+{{- if .Values.ingressController.image.effectiveSemver -}}
+  {{- $version = semver .Values.ingressController.image.effectiveSemver -}}
+{{- else -}}
+  {{- $version = semver .Values.ingressController.image.tag -}}
+{{- end -}}
 {{- ge $version.Major 2 -}}
 {{- end -}}
 
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -596,6 +596,7 @@ section of `values.yaml` file:
 | enabled                            | Deploy the ingress controller, rbac and crd                                           | true                                                                         |
 | image.repository                   | Docker image with the ingress controller                                              | kong/kubernetes-ingress-controller |
 | image.tag                          | Version of the ingress controller                                                     | 1.2.0 |
+| image.effectiveSemver              | Version of the ingress controller used for version-specific features when image.tag is not a valid semantic version | |
 | readinessProbe                     | Kong ingress controllers readiness probe                                              |                                                                              |
 | livenessProbe                      | Kong ingress controllers liveness probe                                               |                                                                              |
 | installCRDs                        | Creates managed CRDs.                                                                 | false
--- a/charts/kong/templates/servicemonitor.yaml
+++ b/charts/kong/templates/servicemonitor.yaml
@@ -1,4 +1,5 @@
 {{- if and ( .Capabilities.APIVersions.Has "monitoring.coreos.com/v1" ) .Values.serviceMonitor.enabled }}
+{{- $controllerIs2xPlus := include "kong.controller2xplus" . -}}
 apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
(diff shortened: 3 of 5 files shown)


mflendrich (commented):
> This change imposes new requirements on values. The controller image tag must be semver-compliant. The template will completely fail to render if it is not.

Hm, this will not work nicely with custom builds and with our tags such as `next`.

What do you think about something like this:
- add a toggle to values.yaml that will control whether the controller is included in the service monitor, with the following semantics:
    - if unset, defaults to checking the semver version from the tag - if inconclusive, throwing a human-readable error suggesting that the user set the setting
    - if enabled/disabled: enabling/disabling the controller port in the service monitor
- control rendering of the controller endpoint in the service monitor based solely on the value of that toggle

charts/kong/templates/_helpers.tpl
```diff
@@ -535,7 +535,17 @@ The name of the service used for the ingress controller's validation webhook
   {{- toYaml .Values.resources | nindent 4 }}
 {{- end -}}
 
+{{/*
+kong.controller2xplus returns "true" if the controller image tag major version is >= 2
+Note that this is a string, not a boolean, because templates vov
+*/}}
+{{- define "kong.controller2xplus" -}}
```
mflendrich, line 542: consider an alternative that would be resilient to the `v` prefix being present or not:

`regexMatch "^v?2\." .Values.ingressController.image.tag`

(based on an understanding that the implementation using `semver` is more restrictive)

rainest: It's at least tolerant of that:

```
$ helm template ex /tmp/symkong --set serviceMonitor.enabled=true --set ingressController.image.tag=v2.2 | grep -A2 cmetrics 
        - name: cmetrics
          containerPort: 10255
          protocol: TCP

09:56:21-0700 yagody $ helm template ex /tmp/symkong --set serviceMonitor.enabled=true --set ingressController.image.tag=v1.2 | grep -A2 cmetrics 

$ helm template ex /tmp/symkong --set serviceMonitor.enabled=true --set ingressController.image.tag=dfgdfgv1.2 | grep -A2 cmetrics 
Error: template: kong/templates/deployment.yaml:93:10: executing "kong/templates/deployment.yaml" at <include "kong.controller-container" .>: error calling include: template: kong/templates/_helpers.tpl:548:27: executing "kong.controller-container" at <include "kong.controller2xplus" .>: error calling include: template: kong/templates/_helpers.tpl:543:16: executing "kong.controller2xplus" at <semver .Values.ingressController.image.tag>: error calling semver: Invalid Semantic Version

Use --debug flag to render out invalid YAML
```

rainest:
Bah, yeah, `main` and `next` won't work with this.

For custom builds, the ask is that you use [semver metadata](https://semver.org/#spec-item-10). If you want to distinguish your custom build in the tag, and not just by using your registry, you'd have something like `2.0.0+foobar` to denote that it's a 2.0.0 build that also includes your foobar modification.

We could use something compatible for our bleeding edge tags also (`2.1.0-next` or whatever), but it doesn't look like there's any simple way to do that with the docker-meta action.

I don't want to add toggles for individual version-dependent features that we want to have on by default on compatible versions. You shouldn't have to cross-reference the chart feature set against controller release notes, and shouldn't have to walk through all the values.yaml settings to toggle on things that you'd expect to be on by default.

I can add an optional user-set value for the controller semver version. If you provide it, we use that for the semver input instead of the tag. You'd only set it if you are using something like `next` or `main`. Those should be the minority of deployments by far, and probably mostly only at our instruction (we want someone to test an unreleased feature and ask them to use `next` in an issue). Not ideal, but it'd allow us to still use version gates without individual feature toggles.

rainest:
After 98e50e6:

```
$ helm template ex /tmp/symkong --set serviceMonitor.enabled=true --set ingressController.image.tag=v2.2 --set ingressController.image.effectiveSemver=v2.5.0 | grep -A2 cmetrics 
        - name: cmetrics
          containerPort: 10255
          protocol: TCP

$ helm template ex /tmp/symkong --set serviceMonitor.enabled=true --set ingressController.image.tag=vdfgfdgdfg1.2 --set ingressController.image.effectiveSemver=v2.5.0 | grep -A2 cmetrics                                                                                             
        - name: cmetrics
          containerPort: 10255
          protocol: TCP
```

charts/kong/templates/_helpers.tpl
```diff
@@ -535,7 +535,22 @@ The name of the service used for the ingress controller's validation webhook
...
+{{- $version := "" -}}
+{{- if .Values.ingressController.image.effectiveSemver -}}
```
mflendrich, line 544: consider using the `default` template for brevity instead of if/else

charts/kong/templates/_helpers.tpl
```diff
@@ -535,7 +535,22 @@ The name of the service used for the ingress controller's validation webhook
...
+  {{- $version = semver .Values.ingressController.image.effectiveSemver -}}
+{{- else -}}
+  {{- $version = semver .Values.ingressController.image.tag -}}
+{{- end -}}
+{{- ge $version.Major 2 -}}
```
mflendrich, line 549: could we make this template parametric in the following way, for reusability:
 - accept the major version as parameter
 - accept the `image` object as a parameter

so that it's reusable if we, say, need to check elsewhere if the gateway has major version >= 3?

charts/kong/templates/_helpers.tpl
```diff
@@ -552,6 +567,11 @@ The name of the service used for the ingress controller's validation webhook
     containerPort: {{ .Values.ingressController.admissionWebhook.port }}
     protocol: TCP
   {{- end }}
+  {{ if (eq $controllerIs2xPlus "true") -}}
```
mflendrich, line 570: can we simplify the `(eq $controllerIs2xPlus "true")` to just `$controllerIs2xPlus` since it's a boolean value already?

mflendrich (approved):
some of the attached comments may show my helm templating language ignorance :man_shrugging:

charts/kong/templates/_helpers.tpl
rainest: It _should_ be a boolean value :smile: Earlier versions tried exactly that, causing me a great deal of confusion when debug printouts of the helper result clearly showed that it was `false` but the block inside the `if` still rendered.

As best I can tell through black box analysis (since I didn't want to try and run Helm via delve to inspect the guts of what gotemplate is actually doing) is that the result of an `include` is always a string from the includer's perspective, regardless of what type it was within the `include`.

Thus, if it evaluates to `false`, the actual check in the `if` is a string check, and anything other than the empty string is truthy.

There's no function to convert string to boolean, so the simplest solution is to perform an exact string comparison. It's real stupid and makes me cry out in pain for a proper function call system with type guarantees. 50% of template development is "damnit, I really need to build an operator to replace this."

charts/kong/templates/_helpers.tpl
rainest: This:
```
{{- define "kong.controller2xplus" -}}
{{- $version := default (semver .Values.ingressController.image.tag) (semver .Values.ingressController.image.effectiveSemver) -}}
{{- ge $version.Major 2 -}}
{{- end -}}
```
is cleaner, but `default` evaluates all its arguments immediately, so it tries to call `semver` on the tag even if `effectiveSemver` isn't empty, and calls it on `effectiveSemver` if it isn't set. 415b48fdd does improve on it a bit though, by defaulting strings and putting the conversion after.

charts/kong/templates/_helpers.tpl
rainest: Looked at this briefly and I think I'd like to defer it for now. This function is intentionally a bit weird because we want this to work for the beta, and for that the major version comparison is necessary over a more conventional `semver.Compare`: `2.0.0-beta.1` is not greater than `2.0.0` (and IIRC it's not even greater than `1.3`, ignoring that using that would cause issues if we eventually released 1.4). Realistically our more immediate need, if any, will be to check if the controller version is >= 2.1, not >= 3.

We could work around that by converting to semver first, extracting the major, and passing that in, but IMO more work than its worth at this point--makes more sense to release a simpler version for now and then refactor into something more reusable after 2.0 GA.

Making a "function" in template land is a bit cumbersome, since parameters aren't really a thing--you have only a single parameter, so you have to construct a dict on the fly and don't have the benefit of signature checking. It's clunky but doable. 

Whether we truly want to do that is an open question--my general experience with developing entirely in templates is that it's an exercise in masochism. This is a good example, where we have odd workarounds for not having proper functions and types and lack access to the full feature set of the underlying semver library. I think a reasonable next major project is to begin porting chart functionality into a full Go operator that handles complex decisions in Go and only uses templates for a final "fill in the exact given values" step to create a manifest.

PR #442, opened 2021-08-18
[kong] add migrations.backoffLimit

### LINKED ISSUE #428: Review wait-for-db options
Hi,

Currently there are two available options in order to setup a hook for waiting for the database to show up.

1. A Kong binary one, which is database type aware (Postgres, Cassandra).
2. A low level TCP one, only applicable when Postgres is embedded via a local chart dependency.

A few comments at this stage

- The migrations jobs, does not support to setup a wait database hook outside the embedded use case.
- The Kong binary option requires the database password to be provided, hence when integrating with the Vault injector agent we have no way to prioritize yet the ordering of the init containers, to inject the database password prior to the `wait-for-db` init container execution.

Option A

Allow selecting between the TCP low level option and the Kong binary one. This may require the TCP one to support Cassandra too.
Allow setting up a wait hook on the jobs.

Option B

Build your own waiting hook, and allow this one to be injected in the jobs too.

Thoughts ?

rainest:
Add migrations.backoffLimit to values, which sets the backoffLimit in
migration Jobs' specs.

  - fixes #428 

#### Checklist

### DIFF
--- a/charts/kong/values.yaml
+++ b/charts/kong/values.yaml
@@ -347,6 +347,7 @@ migrations:
   # This is helpful in certain non-Helm installation situations such as GitOps
   # where additional control is required around this job creation.
   jobAnnotations: {}
+  backoffLimit:
   resources: {}
   # Example reasonable setting for "resources":
   # resources:
--- a/charts/kong/values.yaml
+++ b/charts/kong/values.yaml
@@ -347,6 +347,7 @@ migrations:
   # This is helpful in certain non-Helm installation situations such as GitOps
   # where additional control is required around this job creation.
   jobAnnotations: {}
+  # Optionally set a backoffLimit. If none is set, Jobs will use the cluster default
   backoffLimit:
   resources: {}
   # Example reasonable setting for "resources":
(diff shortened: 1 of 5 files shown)


charts/kong/values.yaml
```diff
@@ -347,6 +347,7 @@ migrations:
   # This is helpful in certain non-Helm installation situations such as GitOps
   # where additional control is required around this job creation.
   jobAnnotations: {}
+  backoffLimit:
```
shaneutt, line 351: It seems we should add some comments here, as we do for some of the surrounding values, which links to the relevant documentation and/or has some minimal explanation of the usage?

shaneutt (approved):
No hard blockers, but one comment I would like to see resolved please :+1:

charts/kong/values.yaml
rainest: Hurray forgetting to save. f8b678c45

ccfishk (approved):

ccfishk:
Should we also gate parameters value ?   
if I pass a string "100RPS", should  we prevent user input such value ?

PR #447, opened 2021-08-19
[kong] fix UDP ports

### LINKED ISSUE #445: udpProxy configuration uses TCP for listener
While [working on documentation for 2.x](https://github.com/Kong/docs.konghq.com/pull/3121) and testing the `udpProxy` section of the Helm chart I found that the `KONG_STREAM_LISTEN` configuration can become broken.

The following is my `values.yaml`:

```yaml
ingressController:
  image:
    tag: "2.0.0-beta.2"
udpProxy:
  enabled: true
  type: LoadBalancer
  stream:
  - containerPort: 9999
    servicePort: 9999
    protocol: UDP
```

Given the above `values.yaml` the `KONG_STREAM_LISTEN` environment variable that was produced for the Kong proxy pods was:

```yaml
- name: KONG_STREAM_LISTEN
  value: 0.0.0.0:9999
```

This lead to my `UDPIngress` resources being inoperable, as it seems the default of `TCP` will be used if the protocol isn't specified here.

I was able to work around this by manually patching the pods with the following values:

```yaml
- name: KONG_STREAM_LISTEN
  value: "0.0.0.0:9999 udp"
```

After that change my `UDPIngress` resources started to route traffic properly.

rainest:
Fixes several issues with UDP listens and ports.

  - fixes #445 

#### Checklist

### DIFF
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -160,7 +160,7 @@ spec:
   {{- end }}
   {{- if (hasKey . "stream") }}
   {{- range .stream }}
-  - name: stream-{{ if (eq (default .protocol "TCP") "UDP") }}udp-{{ end }}{{ .containerPort }}
+  - name: stream{{ if (eq (default "TCP" .protocol) "UDP") }}udp{{ end }}-{{ .containerPort }}
     port: {{ .servicePort }}
     targetPort: {{ .containerPort }}
     {{- if (and (or (eq $.type "LoadBalancer") (eq $.type "NodePort")) (not (empty .nodePort))) }}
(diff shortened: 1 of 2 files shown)


ccfishk (approved):
lgtm

shaneutt (approved):
We should add some tests to protect ourselves from making regressions in this functionality in the future, and then otherwise LGTM :+1:

shaneutt:
It doesn't seem that we ended up with any tests before merging this. My expectation is that it should be fairly straightforward to add a simple `UDPIngress` test to our existing integration tests to provide that we can enable UDP with the chart (for instance to proxy to coredns and make a simple DNS lookup in the test). Can we follow up on this?

PR #448, opened 2021-08-20
[kong] ServiceAccount can't be created if not using ingressController (#418)

### LINKED ISSUE #418: ServiceAccount can't be created if not using ingressController
Hi,

I have a problem because I use the serviceAccount but I don't use IngressController. And with the last changes, It's impossible to use it. Also I need to mount the token because I use vault-secret-webhook and for this is necessary. Can be this optional and enable it by any environment?

Is there any workaround to fix it? 

Thanks

ludovic-pourrat:
[kong] ServiceAccount can't be created if not using ingressController (#418)

Add the feature to create a service account for the deployment and the migrations

 - fixes #418 

#### Checklist

### DIFF
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -629,6 +629,9 @@ For a complete list of all configuration values you can set in the
 | deployment.daemonset               | Use a DaemonSet instead of a Deployment                                               | `false`             |
 | deployment.userDefinedVolumes      | Create volumes. Please go to Kubernetes doc for the spec of the volumes               |                     |
 | deployment.userDefinedVolumeMounts | Create volumeMounts. Please go to Kubernetes doc for the spec of the volumeMounts     |                     |
+| deployment.serviceAccount.create   | Create Service Account for the Deployment / Daemonset and the migrations              | `false`             |
+| deployment.serviceAccount.name     | Name of the Service Account, a default one will be generated if left blank            | ""                  |
+| deployment.serviceAccount.annotations | Annotations for the Service Account                                                | {}                  |
 | autoscaling.enabled                | Set this to `true` to enable autoscaling                                              | `false`             |
 | autoscaling.minReplicas            | Set minimum number of replicas                                                        | `2`                 |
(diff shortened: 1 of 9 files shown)


rainest:
This generally looks sound. I'm trying to think of some elegant way that makes it so we don't end up with two different templates for the two ServiceAccount options, so that we don't end up with some situation where we add features to one and not the other (something that's bitten us in the past).

Helm templates are not making this easy :neutral_face:

ludovic-pourrat:
@rainest I was thinking the same, but wasn't sure the way you wanted to be maintained.
Thoughts or should I proceed to update this PR in that way ?

ludovic-pourrat:
Needs rework for taking into account the `automountServiceAccountToken`

rainest:
> but wasn't sure the way you wanted to be maintained. Thoughts or should I proceed to update this PR in that way ?

I wasn't sure how I wanted to do it either--my typical standard approach is to stuff a template into _helpers.tpl and then build a context elsewhere with appropriate values for it, like how https://github.com/Kong/charts/blob/b87a69dcb92928e228aad59608b439f02ad65f6e/charts/kong/templates/service-kong-admin.yaml#L3-L10 and https://github.com/Kong/charts/blob/b87a69dcb92928e228aad59608b439f02ad65f6e/charts/kong/templates/_helpers.tpl#L101-L180 work to build a Service.

That felt a bit clunky here since we're not building multiple ServiceAccounts, we're building just one. We furthermore need to make the decision about which version to use in a few different places (e.g. the name goes in both the ServiceAccount itself and in the Deployment using it).

I think it might be possible to use a single helper that determines which context to use to avoid repeating the decision if block, so that you end up with something like (messy pseudocode):

```
var serviceAccountContext
if <we should use the controller version> {
  serviceAccountContext = ingressController.serviceAccount
} else {
  serviceAccountContext = deployment.serviceAccount
}
<further template stuff uses serviceAccountContext.name, etc.>
```
Unsure if you can use a helper for that or not--it may be that you have to assign that variable in each template that uses the ServiceAccount.

PR #455, opened 2021-09-02
[kong] move ServiceAccount configuration

### LINKED ISSUE #418: ServiceAccount can't be created if not using ingressController

rainest:
Allow creation and configuration of ServiceAccounts independent of the controller. By default, always create a ServiceAccount, changing it from a default conditional (if controller or PSP in use) on to just default on. Fix #418 

Originally a user PR, but didn't hear back after the request to reopen in https://github.com/Kong/charts/issues/418#issuecomment-912663576. Opening my own since I'd like this to go in before the next release, which is imminent.


mflendrich:
@rainest what isthe specific use for CI that wasn't fulfilled by our existing triggers? Consider filing an `area/ci` issue if it makes sense to fix going forward

rainest:
It's just on PRs rather than branches, which makes sense for the vast majority of chart stuff. This one was a bit weird since I needed to make a new PR to test out my changes on top of stuff that originated from a fork.

rainest:
Ignore the prior conversation, this is a real PR and should be reviewed now.

charts/kong/README.md
```diff
@@ -637,6 +634,9 @@ For a complete list of all configuration values you can set in the
 | deployment.daemonset               | Use a DaemonSet instead of a Deployment                                               | `false`             |
 | deployment.userDefinedVolumes      | Create volumes. Please go to Kubernetes doc for the spec of the volumes               |                     |
 | deployment.userDefinedVolumeMounts | Create volumeMounts. Please go to Kubernetes doc for the spec of the volumeMounts     |                     |
+| deployment.serviceAccount.create   | Create Service Account for the Deployment / Daemonset and the migrations              | `true`              |
+| deployment.serviceAccount.name     | Name of the Service Account, a default one will be generated if left blank.           | ""                  |
+| deployment.serviceAccount.annotations | Annotations for the Service Account                                                | {}                  |
```
shaneutt, line 639: nit: you could re-space the table so that in a non-web view it looks even

charts/kong/templates/_helpers.tpl
```diff
@@ -52,13 +52,14 @@ app.kubernetes.io/instance: "{{ .Release.Name }}"
...
-{{- if .Values.ingressController.serviceAccount.create -}}
-    {{ default (include "kong.fullname" .) .Values.ingressController.serviceAccount.name }}
+{{- if .Values.deployment.serviceAccount.create -}}
+    {{ default (include "kong.fullname" .) .Values.deployment.serviceAccount.name }}
 {{- else -}}
     {{ default "default" .Values.ingressController.serviceAccount.name }}
 {{- end -}}
 {{- end -}}
 
+
```
shaneutt: nit: seems like errant blank space?

```suggestion
```

shaneutt (approved):
No blockers just nitpicks :+1:

charts/kong/README.md
rainest: Never!

PR #464, opened 2021-09-29
[kong] docs: 2.4.0 upgrade procedures

### LINKED ISSUE #450: Chart upgrade needs to spin down controller during v1 to v2 upgrade when using DB mode
During the upgrade from `v1` to `v2` when DB mode is enabled (e.g. `postgres`) we shouldn't use a standard rollout upgrade.

We shouldn't allow any period of time where both controllers are acting on the proxy at the same time to avoid:

- 1.x and 2.x "fighting" over the data-plane configuration
- 2.x ending up writing a new parallel batch of config unrelated to 1.x's config
- other untested scenarios where both control-planes are actively trying to manage the data-plane.

**NOTE**: this doesn't affect DBLESS where the proxy pod will be completely recreated (and will be separate) anyhow.

Acceptance Criteria:
- [ ] when upgrading from v1 to v2 with DB mode backend the old controller is terminated _before_ the new controller is spun up

shaneutt:

Adds documentation for the upcoming 2.4.0 chart release relevant to the inclusion of the new KIC.

Resolves https://github.com/Kong/charts/issues/450

#### Checklist

### DIFF
--- a/charts/kong/UPGRADE.md
+++ b/charts/kong/UPGRADE.md
@@ -17,6 +17,7 @@ upgrading from a previous version.
 ## Table of contents
 
 - [Upgrade considerations for all versions](#upgrade-considerations-for-all-versions)
+- [2.4.0](#240)
 - [2.3.0](#230)
 - [2.2.0](#220)
 - [2.1.0](#210)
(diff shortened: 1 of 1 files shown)


rainest (approved):
Enough changes and realigning hard wraps that this was easiest via a commit: f1f6e31b4. Reasoning was:

- Control plane and data plane aren't hyphenated, though I've substituted ingress controller and admin API instead since we typically use CP and DP to refer to the hybrid mode concepts.
- Clarified action in section heading.
- Limited some redundant statements.
- Active voice, prescriptive statements. We _do_ want users to take specific steps unless they'll absolutely need to adjust for their environment, and they shouldn't here. I omitted the values.yaml/--set distinction since they need to change the command string regardless to use `--version`, so they may as well do everything via the command invocation.
- `--wait` since Helm has "wait until everything in the upgrade completes before the command exits" built in.

PR #465, opened 2021-09-29
[kong] fix: dbless configmap duplicate pipes

### LINKED ISSUE #463: dblessConfig.config creating invalid configmap
When installing kong with the helm command `--set-file dblessConfig.config=kong.yml` where kong.yml is a valid kong declarative configuration, it is inserting an extra `|` into the generated configmap. Kong then fails to start as it has an invalid configuration. 

The generated configmap (snipped for brevity):
```
apiVersion: v1
kind: ConfigMap
name: dbless-config-custom
namespace: kong                                                                                                                                                                                                                                                                                                                                                                              │
│ data:                                                                                                                                                                                                  │
│   kong.yml: |                                                                                                                                                                                          │
│     |-                                                                                                                                                                                                 │
│       _format_version: "1.1"     
```

The error on kong startup:
```
2021/09/29 19:08:14 [error] 1#0: init_by_lua error: /usr/local/share/lua/5.1/kong/init.lua:525: error parsing declarative config file /kong_dbless/kong.yml:                                           │
│ failed parsing declarative configuration: expected an object                                                                                                                                           
│ stack traceback:                                                                                                                                                                                       
│     [C]: in function 'error'                                                                                                                                                                           
│     /usr/local/share/lua/5.1/kong/init.lua:525: in function 'init'                                                                                                                                     
│     init_by_lua:3: in main chunk
```

usrbinkat:
DB-less ConfigMap generates faulty ConfigMap with duplicate pipes on value when using helm `--set-file dblessConfig.config` or piping into `--set dblessConfig.config`

  - fixes #463 

#### Checklist

Signed-off-by: usrbinkat <usrbinkat@braincraft.io>

### DIFF
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -146,17 +146,18 @@ read the [env](#the-env-section) section.
 
 When deploying Kong in DB-less mode(`env.database: "off"`)
 and without the Ingress Controller(`ingressController.enabled: false`),
-you have to provide a [declarative configuration]https://docs.konghq.com/gateway-oss/latest/db-less-and-declarative-config/#the-declarative-configuration-format)
-for Kong to run. You can provide an existing ConfigMap
+you have to provide a [declarative configuration]https://docs.konghq.com/gateway-oss/latest/db-less-and-declarative-config/#the-declarative-configuration-format) for Kong to run.
+You can provide an existing ConfigMap
 (`dblessConfig.configMap`) or place the whole configuration into
 `values.yaml` (`dblessConfig.config`)
 parameter. See the example configuration in the default values.yaml
 for more details. You can use `--set-file dblessConfig.config=/path/to/declarative-config.yaml`
 in Helm commands to substitute in a complete declarative config file.    
     
-Note that ConfigMap updates will not restart Kong pods when using `dblessConfig.configMap`,
-and they will not recognize updated configuration until restarted. You should run
-`kubectl rollout restart deploy` after updating the configuration ConfigMap contents.    
+Note that externally supplied ConfigMaps are not hashed or tracked in deployment annotations.
+Subsequent ConfigMap updates will require user-initiated new deployment rollouts
+to apply the new configuration. You should run `kubectl rollout restart deploy`
+after updating externally supplied ConfigMap content.    
     
 #### Using the Postgres sub-chart
 
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -146,7 +146,7 @@ read the [env](#the-env-section) section.
 
 When deploying Kong in DB-less mode(`env.database: "off"`)
 and without the Ingress Controller(`ingressController.enabled: false`),
-you have to provide a [declarative configuration]https://docs.konghq.com/gateway-oss/latest/db-less-and-declarative-config/#the-declarative-configuration-format) for Kong to run.
+you have to provide a [declarative configuration](https://docs.konghq.com/gateway-oss/latest/db-less-and-declarative-config/#the-declarative-configuration-format) for Kong to run.
 You can provide an existing ConfigMap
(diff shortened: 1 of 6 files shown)


gperanich (approved):
+1 in the values.yaml, may need to call out the importance of keeping the pipe next to config:
`config: |`

usrbinkat:
> +1 in the values.yaml, may need to call out the importance of keeping the pipe next to config: `config: |`

I would consider this standard notation in the industry and suggest over-documenting standard notation may clutter the inline documentation. Thoughts?

charts/kong/README.md
```diff
@@ -146,13 +146,22 @@ read the [env](#the-env-section) section.
...
+supplied under the `dblessConfig.config` paramater of `values.yaml`, or with the
+helm flag `--set-file` parameter. See the example configuration in the default 
+[values.yaml](values.yaml#L371) or [example-values/dbless-config.yaml](example-values/dbless-config.yaml)
+for syntax and more details.    
+    
+Alternatively, the configuration can be provided separately using an existing
+ConfigMap name supplied to Kong in (`dblessConfig.configMap`). Note, externally
+supplied ConfigMaps are not hashed or tracked in deployment annotations.
+Subsequent ConfigMap updates will require user-initiated new deployment rollouts
+to apply the new configuration.
```
rainest: - Latest doc since we don't usually scan through docs to bump versions.
- values.yaml locations change all the time, so only including the key names here.
- The example-values directory is for values.yamls only rather than general example content. I recommend omitting any separate example here since the linked doc page already includes its own example, which doesn't need to change at all for `--set-file` use.
- Reworded the ConfigMap rollout warning to apply across the board. The template version/`dblessConfig.config` only updates the generated ConfigMap and won't change the Kong Deployment, so Helm won't perform its own rollout if that's the only thing you're changing.
 
```suggestion
you have to provide a [declarative configuration]https://docs.konghq.com/gateway-oss/latest/db-less-and-declarative-config/#the-declarative-configuration-format) for Kong to run.
You can provide an existing ConfigMap
(`dblessConfig.configMap`) or place the whole configuration into
`values.yaml` (`dblessConfig.config`)
parameter. See the example configuration in the default values.yaml
for more details. You can use `--set-file dblessConfig.config=/path/to/declarative-config.yaml`
in Helm commands to substitute in a complete declarative config file.
    
Note that ConfigMap updates will not restart Kong pods when using `dblessConfig.configMap`,
and they will not recognize updated configuration until restarted. You should run
`kubectl rollout restart deploy` after updating the configuration ConfigMap contents.
```

charts/kong/example-values/dbless-config.yaml
```diff
@@ -0,0 +1,9 @@
+_format_version: "1.1"
+services:
+  # Example configuration
+  - name: example.com
+    url: http://example.com
+    routes:
+    - name: example
+      paths:
+      - "/example"
```
rainest: Per the other comment, just remove this, since we use this directory for other content and don't need to provide anything different than upstream docs.

rainest (changes requested):
The test failures are from some example values that unnecessarily include the config block from the default and were still using the version before the pipe was added to the default. 
[examples.diff.txt](https://github.com/Kong/charts/files/7255398/examples.diff.txt) should clear those.

charts/kong/example-values/dbless-config.yaml
usrbinkat: [done](https://github.com/Kong/charts/pull/465/commits/071a59910032ba6435c86197ce0a93ab5b800a9f)

charts/kong/README.md
usrbinkat: I believe 5848f6e & 27e8aee resolves these items.

rainest (approved):

PR #471, opened 2021-10-07
Firm up CI conditions

rainest:
The cause of the CI breakage in the 2.4.0 release job is unclear: https://github.com/Kong/charts/actions/runs/1317477468

Unfortunately, `ct list-changed` doesn't provide much anything in terms of output describing how it decides what did or didn't change. As we've had no CI changes since 2.3, however, whatever broke was hopefully a fluke that will be fixed by just bumping the version again :shrug: 

The only thing I can think of is that there's a race condition with the sync master job that causes this. We set main as the base branch for most commands, but didn't apply the default args to `list-changed`. Moving default args to a ct.yaml means that all ct invocations will use those settings by default, without explicitly adding them to the individual workflow step commands.

### DIFF
--- a/.github/workflows/main-pr.yaml
+++ b/.github/workflows/main-pr.yaml
@@ -7,8 +7,6 @@ on:
 
 jobs:
   lint-test:
-    env:
-      COMMON_CT_ARGS: "--chart-repos bitnami=https://charts.bitnami.com/bitnami --remote origin --target-branch main"
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
@@ -30,6 +28,7 @@ jobs:
 
       - name: Run chart-testing (list-changed)
         id: list-changed
+        working-directory: charts
         run: |
           changed=$(ct list-changed)
           if [[ -n "$changed" ]]; then
@@ -37,11 +36,13 @@ jobs:
           fi
 
       - name: Run chart-testing (lint)
-        run: ct lint $COMMON_CT_ARGS
+        run: ct lint
+        working-directory: charts
 
       - name: Create kind cluster
         uses: helm/kind-action@v1.2.0
         if: steps.list-changed.outputs.changed == 'true'
 
       - name: Run chart-testing (install)
-        run: ct install $COMMON_CT_ARGS
+        run: ct install
+        working-directory: charts
--- a/.github/workflows/main-push.yaml
+++ b/.github/workflows/main-push.yaml
@@ -26,14 +26,25 @@ jobs:
       - name: Set up chart-testing
         uses: helm/chart-testing-action@v2.0.1
 
+      - name: Run chart-testing (list-changed)
+        id: list-changed
+        working-directory: charts
+        run: |
+          changed=$(ct list-changed)
+          if [[ -n "$changed" ]]; then
+            echo "::set-output name=changed::true"
+          fi
+
       - name: Run chart-testing (lint)
         run: ct lint
+        working-directory: charts
 
       - name: setup testing environment (kind-cluster)
         run: ./scripts/test-env.sh
 
       - name: run chart-testing (install)
         run: ct install
+        working-directory: charts
 
       - name: run integration tests (integration)
         run: ./scripts/test-run.sh
--- a/.github/workflows/non-main.yaml
+++ b/.github/workflows/non-main.yaml
@@ -10,8 +10,6 @@ on:
 
 jobs:
   lint-test:
-    env:
-      COMMON_CT_ARGS: "--chart-repos bitnami=https://charts.bitnami.com/bitnami --remote origin --target-branch next"
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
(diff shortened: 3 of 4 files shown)


mflendrich (changes requested):
If my hunch is right, the 2.4.0 release failure was caused by the `if` condition suppressing creation of a `kind` cluster in the test, but not suppressing subsequent use of that (nonexistent) cluster.

A PR will follow that will try to fix that. If it works, we can try to release 2.4.0 again. I don't think I'm on board with bumping the patch number: the 2.4.0 release didn't happen effectively, and the release notes (and version numbers) are too public-facing to be influenced by such CI hiccup.

Unless you feel strongly that we should do otherwise, let's see if we can fix the CI and re-release 2.4.0. Hence the "changes requested" (so that nobody merges this accidentally). (edit: https://github.com/Kong/charts/pull/473 does that)

mflendrich:
2.4.0 released successfully with #473. So, as far as I understand, the 2.4.1 release does not have a reason to happen anymore.

rainest:
> If my hunch is right, the 2.4.0 release failure was caused by the if condition suppressing creation of a kind cluster in the test, but not suppressing subsequent use of that (nonexistent) cluster.

This is true, it's just that it's not clear why the change check failed here. It was working fine for previous releases, so there's not an obvious reason it should go away.

What I think may have happened here (hard to confirm from action output) was a race condition between the separate "release" and "sync master to main" jobs. Assuming "release" runs first, it was checking for any changed charts on a master branch that didn't have the changes from main yet. No changed chart in master, `list-changed` returns nothing, and the rest of "release" fails. My change was to make `list-changed` run against main as well, which should remove that potential issue.

I think we probably do still want that, though we wouldn't need the version bump since #473 did release 2.4.0 (I figured that the change check wouldn't detect anything if there were no changes; removing that check made it a moot point). AFAIK, the reason that exists is because the tool was originally designed for multi-chart repos, and the main push job would attempt to release all charts without the change check. That's less of an issue for us since we only have the Kong chart in here, though it'd still trigger release jobs on other pushes to main (CI-only changes, for example). Those should be benign since the release job just fails if the release already exists, but it would be a bit annoying.

rainest:
Rebase and force push drops the 2.4.1 release and updates CI only. Including ct.yaml and using the working directory always means we should have no chance of looking at master before it gets a mirrored update. This PR now includes the CI changes only and leaves the chart as is.

Per https://github.com/Kong/charts/pull/471#issuecomment-938847611 I do think we want the checks removed in #473 to remain in place (though they should perhaps fail more gracefully than halting and catching fire if the test cluster exists). main pushes that do not include changes to a chart aren't as common for our repo, but they're not unheard of, and those changes _should_ avoid release attempts.

mflendrich (commented):
I am consistently failing to understand how `ct list-changed` is supposed to work in the `main-push` workflow, and therefore readdition of the `ct list-changed` run to **that specific workflow** is something I remain unconvinced about:

`main-push.yaml` runs on each commit after it gets **pushed** to main. It's reasonable to assume that (saving for corner cases) `main-push.yaml` runs for a commit if and only if it's the head of `main`.

`ct list-changed`, as configured in this PR, takes two inputs (below) and compares them for equality:
- input 1 is the working copy,
- input 2 is `origin/main` (as configured by `remote` and `target-branch` parameters).

The working copy (due to the current CI triggers) _is always the same commit as_ `origin/main`. Therefore, the comparison between input 1 and input 2 will _always_ (if no pushes happen during the CI run) respond that these two inputs are equal.

If my reasoning above is correct, what is the point of performing that check in the `main-push` workflow at all?

rainest:
I'm not entirely sure either :smile: I just know that it normally does work, and haven't dug too deeply into what the source does and/or how that works in practice in a GH Actions environment.

AFAIK the reason why it's this way is that `ct` was originally designed for https://github.com/helm/charts/, where the repo contained many charts and pushes to main would only update a single chart. The release job should only act on the one updated chart, since it will just fail on others.

Our repo isn't quite the same since it only contains the one chart--we'd considered incorporating other Kong products into it in the past, but that never happened (Brain never completed its chart and Kuma maintains its own repo). Absent additional charts, the only pushes to main that don't bump the Kong chart version are CI changes, and I believe those will attempt a release without the `list-changed` gate. Attempting to release an already-released version will fail benignly, since the release job includes its own check to see if a version already exists. It will emit an error, but otherwise does nothing.

shaneutt:
Looking at the help file for `ct list-changed` here:

https://github.com/helm/chart-testing/blob/main/doc/ct_list-changed.md

It appears the entire purpose is to list the charts changed from your current local repository against a remote branch, in order to identify which charts you should be testing.

It does seem to me that given our use case we can leave this out, as we're never using it to test one branch against another we're using it entirely from the context of the current HEAD of origin/main. We also have only one chart, and we want to specifically test it on every main/next push for thoroughness, no?

My impression is that the only reason we would want to use `ct list-changed` in the future is if we had multiple charts and a lot of PR's because then we could use this to selectively target which charts to test on pushes to main/next. I am in favor of leaving it out until we have more than 1 chart.

rainest:
Partially continued in https://github.com/Kong/charts/pull/479

PR #479, opened 2021-10-22
[kong] rework chart-testing CI config

rainest:
Continuation of https://github.com/Kong/charts/pull/471, sorta.

This now just adds additional config that we should have for `ct` runs to the config YAML, and bumps the action version for good measure.

Removes the condition for the test environment from other branches also, though it leaves a comment about what's broken instead of pruning it entirely.

#### Checklist

### DIFF
--- a/.github/workflows/main-pr.yaml
+++ b/.github/workflows/main-pr.yaml
@@ -7,8 +7,6 @@ on:
 
 jobs:
   lint-test:
-    env:
-      COMMON_CT_ARGS: "--chart-repos bitnami=https://charts.bitnami.com/bitnami --remote origin --target-branch main"
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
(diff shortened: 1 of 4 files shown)


rainest:
Originally attempted to restore the list-changed behavior. While the original logic to compare against `HEAD` seems wrong, the latest version offers an a new `--since` flag to allow comparisons against a specific commit.

This arguably _should_ work with the previous commit, so https://github.com/rainest/charts-citest/blob/ba2820e1749285962c3a9476d6f525431acd10fa/.github/workflows/main-push.yaml#L33-L37 should be fine.

For some reason it works fine locally, specifying the same commit that Github subbed in, but not in the action. Not really sure what's going on there.

As such, this PR just rearranges config and removes the condition from the other affected job pending some solution to make list-changed do what we want.

shaneutt (approved):

PR #480, opened 2021-10-27
[kong] clear stale PIDs from prefix

### LINKED ISSUE #466: Clean up the Kong PID file on pod restart
Summary:

When a Kong container gets abruptly restarted (e.g. due to a power failure of the node), the PID file can prevent Kong from starting again in the same pod. See https://github.com/Kong/kong/issues/6696 for root cause analysis.

Expected behavior:
1. Kong Gateway is running in a pod. The PID file is present in the container.
2. Power gets cut and restored
3. Kong Gateway fails to start up because of the PID file being present in the container.

Expected behavior:
1. Kong Gateway is running in a pod. The PID file is present in the container.
2. Power gets cut and restored
3. The Kong Gateway pod starts up again successfully

rainest:
Add an init container to delete the "pids/" directory from the prefix. This allows Kong to start in the event that the Kong container crashed abruptly (leaving its PIDfile behind in the prefix emptyDir) and restarted in the same Pod.

Fix #466

#### Checklist

### DIFF
--- a/charts/kong/templates/deployment.yaml
+++ b/charts/kong/templates/deployment.yaml
@@ -70,15 +70,23 @@ spec:
         - name: {{ . }}
       {{- end }}
       {{- end }}
-      {{- if (or (and (not (eq .Values.env.database "off")) .Values.waitImage.enabled) .Values.deployment.initContainers) }}
       initContainers:
+      - name: clear-stale-pid
+        image: {{ include "kong.getRepoTag" .Values.image }}
+        command:
+        - "rm"
+        - "-vrf"
+        - "$KONG_PREFIX/pids"
+        env:
+        {{- include "kong.env" . | nindent 8 }}
+        volumeMounts:
+        {{- include "kong.volumeMounts" . | nindent 8 }}
         {{- if .Values.deployment.initContainers }}
         {{- toYaml .Values.deployment.initContainers | nindent 6 }}
         {{- end }}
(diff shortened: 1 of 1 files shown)


charts/kong/templates/deployment.yaml
```diff
@@ -70,15 +70,23 @@ spec:
...
+      - name: clear-stale-pid
+        image: {{ include "kong.getRepoTag" .Values.image }}
+        command:
+        - "rm"
+        - "-vrf"
+        - "$KONG_PREFIX/pids"
+        env:
+        {{- include "kong.env" . | nindent 8 }}
+        volumeMounts:
+        {{- include "kong.volumeMounts" . | nindent 8 }}
```
shaneutt, line 83: It seems like we're looking to _always_ have this initcontainer, but there could be cases where there are no `volumeMounts` and so there's no persisted pidfiles right?

rainest: We always mount the temporary Kong prefix and /tmp volumes. `kong.volumeMounts` doesn't have any condition around those.

Technically you can change the prefix location but you shouldn't. If you do that, this will still run successfully (`rm -f` doesn't care and will succeed even if its target doesn't exist), albeit ineffectively--the volume and mount still exist, even if you don't put files in them.

shaneutt (approved):

PR #484, opened 2021-10-29
[kong] fix(bug): APIVersion of v1 does not guarantee Ingress capability/resource definition exists

ttyS0e:
There is a bug when installing this chart on Kubernetes 1.18 and lower, which is fixed here.

The existing check for the API namespace "networking.k8s.io/v1", when deciding on the Ingress object API, does not work on 1.18 and below. This is because the logic assumes that if "networking.k8s.io/v1" exists at all, then the Ingress object must also exist.

This isn't the case in some Kubernetes releases - the API namespace itself can still exist, but the Ingress object definition is still in the beta namespace.

#### Checklist


mflendrich (approved):
LGTM!

PR #446, opened 2021-08-19
[kong] use latest available Ingress version

rainest:
#### What this PR does / why we need it:
Add support for networking.k8s.io/v1beta1 and networking.k8s.io/v1
Ingress resources. Generate Ingress resources with the newest available
version on the cluster.

Add support for the ingressClassName field and IngressClass resource.

We need Ingress v1 support for 1.22. These are fairly recent, however, so instead of blanket changes to v1, this auto-detects the latest available version and creates resources according.

#### Which issue this PR fixes
*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*
  - fixes #444 

#### Special notes for your reviewer:
Does anyone know of a local K8S environment <1.18 that still works? KIND and Minikube are both failing to start for me on 1.17 and 1.16 images, and I can't find related issues with a fix in those projects.

I can force the extensions content to render with invalid versions:
```
diff --git a/charts/kong/templates/_helpers.tpl b/charts/kong/templates/_helpers.tpl
index 268182f45..386ade324 100644
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -1076,9 +1076,9 @@ Kubernetes resources it uses to build Kong configuration.
 {{- end -}}
 
 {{- define "kong.ingressVersion" -}}
-{{- if (.Capabilities.APIVersions.Has "networking.k8s.io/v1") -}}
+{{- if (.Capabilities.APIVersions.Has "anetworking.k8s.io/v1") -}}
 networking.k8s.io/v1
-{{- else if (.Capabilities.APIVersions.Has "networking.k8s.io/v1beta1") -}}
+{{- else if (.Capabilities.APIVersions.Has "anetworking.k8s.io/v1beta1") -}}
 networking.k8s.io/v1beta1
 {{- else -}}
 extensions/v1beta1
```
Which makes `helm template trk /tmp/symkong -f ~/src/charts/charts/kong/example-values/minimal-kong-controller.yaml --set admin.enabled=true --set admin.ingress.enabled=true > /tmp/fakeold.yaml` render [fakeold.yaml.txt](https://github.com/Kong/charts/files/7017969/fakeold.yaml.txt)

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [x] PR is based off the current tip of the `next` branch and targets `next`, not `main`
- [x] New or modified sections of values.yaml are documented in the README.md
[...]

PR #485, opened 2021-10-29
[kong] fix IngressClass detection

rainest:
The criteria for creating a "kong" IngressClass were previously
backwards: the chart would attempt to create a "kong" IngressClass only
if one already existed. This change inverts that logic to create a
"kong" IngressClass only if one does /not/ already exist.

Reported at https://github.com/Kong/charts/pull/446#issuecomment-954520414

#### Special notes

13ea4d8 is an unrelated CI workaround. The [original test run](https://github.com/Kong/charts/runs/4049754088?check_suite_focus=true) failed because of issues unrelated to the PR. KTF failed to come up due to a timeout in one of the jobs, but not the other. I'm guessing this is due to worker CPU contention, since it appears to have sat around doing nothing for 5 minutes.

May be a fix elsewhere that does allow them to run concurrently, but making them not is the simplest solution I have for now.

#### Checklist

### DIFF
--- a/.github/workflows/non-main.yaml
+++ b/.github/workflows/non-main.yaml
@@ -57,6 +57,8 @@ jobs:
         run: ./scripts/test-env.sh cleanup
 
   integration-test:
+    # Not a real requirement, but due to worker CPU contention, this can fail if it runs concurrent with lint-test
+    needs: lint-test
     runs-on: ubuntu-latest
     steps:
       - name: checkout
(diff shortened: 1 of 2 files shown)


shaneutt (approved):
Thanks for the solid writeup, makes sense to me :+1:

Nica-Alex:
Not sure if this is the root cause, but I'm getting
```Error: unable to recognize "": no matches for kind "IngressClass" in version "networking.k8s.io/v1"```
when trying to upgrade my ingress controller from chart version 2.3.0 to 2.6.0 in k8s version 1.18.

PR #491, opened 2021-11-08
[kong] Release 2.6 to main

rainest:
Continued from https://github.com/Kong/charts/pull/488. This must use a merge commit.

### DIFF
--- a/.github/workflows/main-pr.yaml
+++ b/.github/workflows/main-pr.yaml
@@ -7,8 +7,6 @@ on:
 
 jobs:
   lint-test:
-    env:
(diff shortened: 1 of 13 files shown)


rainest:
Not recalling exactly how the CLA setup for this repo works. We apparently have no workflow for it, and while I can see _something_ related to this repo when I log in, it's showing that it's not actually linked, and attempting to add the Kong org and recheck doesn't appear to do anything.

tysoekong is an employee though, so we can disregard it.

shaneutt (approved):

PR #500, opened 2021-11-23
fix(webhook) exclude Helm Secrets

rainest:
Hurray! We added Secret validation! Oh no! We added Secret validation!

Helm charts with webhooks that handle Secrets run into an issue that
prevents changes after an action that enables the webhook:
https://github.com/helm/helm/issues/10023

Because Helm's Secret for release information is subject to the webhook,
Kubernetes will attempt to validate it, likely before the webhook
service comes online (because Helm just created the Pod that will
provide it). If the service is not online, validation fails, and Helm
cannot update its Secret to mark the release status, usually leaving it
stuck in a pending state that blocks future interactions.

This change excludes Helm Secrets from our validation, because we have
no need to validate them.

#### Checklist

### DIFF
--- a/charts/kong/templates/admission-webhook.yaml
+++ b/charts/kong/templates/admission-webhook.yaml
@@ -34,6 +34,9 @@ metadata:
     {{- include "kong.metaLabels" . | nindent 4 }}
 webhooks:
 - name: validations.kong.konghq.com
+  objectSelector:
+    matchLabels:
+      owner: !helm
   failurePolicy: {{ .Values.ingressController.admissionWebhook.failurePolicy }}
   sideEffects: None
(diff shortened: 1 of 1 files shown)


mflendrich (approved):

mflendrich:
I'm wondering whether this commands any change to the deploy/kustomize manifests we have in the KIC repo, but I don't see an obvious reason that it would.

rainest:
The risk of hitting this with plain manifests should be dramatically less: you'd need to intentionally deploy the manifest and then install or upgrade some unrelated Helm release immediately after.

The proper solution is that we require labels on credential Secrets so we can use a selector for them only and get rid of all unnecessary Secret checks, but that's a breaking change.

PR #503, opened 2021-11-30
Apply the SecurityContext to the PID removal initContainer

rainest:
This should be the same as the proxy container, but was omitted earlier.

#### Checklist


KongCarl:
Can we ensure the init Container also respects the `resources: {}` section?

shaneutt (approved):
Let's make sure @KongCarl's comment gets resolution and then LGTM :+1:

PR #508, opened 2021-12-01
fix(init) rework wait-for-db socket workaround

### LINKED ISSUE #295: Leftover socket file interferes with startup when wait-for-db initContainer is enabled
The Kong container within our Deployment uses the same prefix emptyDir volume as its wait-for-db initContainer.

Because of https://github.com/Kong/kong/issues/6827, wait-for-db will leave a stream_rpc.sock file in the prefix, which prevents the Kong container from starting.

We can work around this by manually removing the file at the end of wait-for-db, but want to remove the workaround once the underlying bug is fixed.

rainest:
Use a temporary KONG_PREFIX for wait-for-db, rather than using the standard prefix and removing leftover files. This avoids the general class of issue described in #295. As that issue is still present on versions of Kong after 2.3, it looks like we'll need to keep the workaround indefinitely, unless we can figure out why we're still leaking socket files despite the upstream fix.

  - fixes #295 

Found while testing the 2.7 nightly. This adds an additional socket (and somehow makes the existing `rm` not work): https://github.com/Kong/kubernetes-testing-framework/pull/168#issuecomment-983180853

#### Checklist

### DIFF
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -529,7 +529,6 @@ The name of the service used for the ingress controller's validation webhook
 {{- end -}}
 
 {{- define "kong.wait-for-db" -}}
-{{ $sockFile := (printf "%s/stream_rpc.sock" (default "/usr/local/kong" .Values.env.prefix)) }}
 - name: wait-for-db
   image: {{ include "kong.getRepoTag" .Values.image }}
   imagePullPolicy: {{ .Values.image.pullPolicy }}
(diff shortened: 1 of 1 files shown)


shaneutt (approved):
This seems like a reasonable way to deal with the problem to me. I can't think of any gotchas for this, and given that it's for the `wait-for-db` container where the end-user is highly unlikely to need to do anything in the prefix directory I feel confident about it.

I would only say maybe give this a couple more cycles and some extra special care testing, but LGTM :heavy_check_mark:

rainest:
Effectively it's just starting Kong from a clean slate same as it would in DB-less mode. Existing tests are fine for this since any issue with it would prevent the Postgres-backed instance from starting at all (either wait-for-db would never succeed or proxy would try to start in some environment where that was impossible).

PR #509, opened 2021-12-01
chore(*) release Kong chart 2.6.2

rainest:
Please review/merge https://github.com/Kong/charts/pull/508 before this. If we need to rework it, I'll remove it from this release (it's only needed for 2.7, so not urgent).

Release 2.6.2 to next. No version bump in Chart.yaml, it was previously added in https://github.com/Kong/charts/blame/0565c0372d2561ff96bc111eb63e686f807b2d56/charts/kong/Chart.yaml#L13

For the future, note that PRs to next other than the actual release PR shouldn't bump that number; we make a determination at release time whether we need a minor or patch version bump.

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -12,7 +12,24 @@ kubectl apply -f https://raw.githubusercontent.com/Kong/charts/main/charts/kong/
 
 ### Improvements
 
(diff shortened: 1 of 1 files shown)


shaneutt (approved):
I'm :heavy_check_mark: for #508 after we've completed testing, so I'm :heavy_check_mark: with this with #508 included.

PR #520, opened 2021-12-15
feat(kong) add optional prometheus rules

EJKinger:
  - fixes #

#### Checklist

### DIFF
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -678,6 +678,9 @@ For a complete list of all configuration values you can set in the
 | serviceMonitor.targetLabels        | ServiceMonitor targetLabels                                                           | `{}`                |
 | serviceMonitor.honorLabels         | ServiceMonitor honorLabels                                                            | `{}`                |
 | serviceMonitor.metricRelabelings   | ServiceMonitor metricRelabelings                                                      | `{}`                |
+| prometheusRules.enabled            | Create PrometheusRules for Kong                                                       | `false`             |
+| prometheusRules.namespace          | Where to create PrometheusRules                                                       |                     |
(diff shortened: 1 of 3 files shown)


andrewgkew:
@shaneutt - when do you think this will be merged into master? I have a client that is wanting to use this new addition to the chart

shaneutt:
> @shaneutt - when do you think this will be merged into master? I have a client that is wanting to use this new addition to the chart

I put this into review on our project board, so it will be reviewed as soon as possible and if there are no changes needed merged quickly. Given the context you've provided that someone is eagerly anticipating this change I've marked this as a medium severity to indicate that anything with low severity should be prioritized _after_ it.

rainest:
Do we have a reason this would need to go into the chart? It looks like it could be deployed independent of the chart, as the spec doesn't include any templated values and and the rules are mapped to the static metric names, rather than being tied to any specific release.

The configuration here is prescriptive and likely won't be valid for all users. Your application may return many non-2XX statuses in the course of normal operation, or may naturally have very spiky traffic. You may have stricter latency requirements and want alerting at >20ms latency.

Between lack of an obvious link to the specific release and the need for different rules in different environments, I'd recommend we leave this out. It should either be deployed independently without Helm, or if the user in question wants to use Helm to manage it alongside their Kong release, that they create an umbrella chart that includes the Kong chart and resources they want to deploy alongside it in a separate chart.

rainest:
Per conversation elsewhere, the plan here is to open a PR that updates https://docs.konghq.com/gateway/2.7.x/vitals/vitals-prometheus-strategy/ and with a link from https://docs.konghq.com/kubernetes-ingress-controller/2.1.x/guides/prometheus-grafana/, so going ahead and closing the PR here.

PR #527, opened 2022-01-17
feat(conf) Adds support startupProbe on Kong Ingress Controller

danibaeyens:
Some big clusters can contain a high number of objects, which might
require more time to be processed during Kong IC boot time.
The new Kubernetes way to protect startup processes is using
`startupProbes`, leaving `livenessProbes` to processes which are
already running and `ready`.

This commit adds support for configuring startupProbe for Kong IC.

As it states, Kubernetes provides 3 different probes for controlling containers: liveness, readiness and startup ones. liveness and readiness are common ones, as they are since the beginning of Kubernetes. Startup probes are on beta since 1.18. As in big clusters Kong Ingress Controller can take longer to start, it's safer to have a separate probe for its boot process and keep  the liveness probe without changes. This way, boot can safely take longer and pod can be restarted quicker if needed.

As the default value for liveness probe is set to 3 retries (failureThreshold) of 10 seconds each, I've just doubled the number of retries for the startup probe. If it boots faster, it's fine, as the readiness probe will set it as ready before it reaches the 60s threshold (6 retries * 10 seconds).

First contribution to kong project, so I've tried to follow the guide as good as possible

#### Checklist

### DIFF
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -611,6 +611,7 @@ section of `values.yaml` file:
 | image.effectiveSemver              | Version of the ingress controller used for version-specific features when image.tag is not a valid semantic version | |
 | readinessProbe                     | Kong ingress controllers readiness probe                                              |                                                                              |
 | livenessProbe                      | Kong ingress controllers liveness probe                                               |                                                                              |
+| startupProbe                       | Kong ingress controllers startup probe                                                |                                                                              |
 | installCRDs                        | Creates managed CRDs.                                                                 | false
 | env                                | Specify Kong Ingress Controller configuration via environment variables               |                                                                              |
 | ingressClass                       | The name of this controller's ingressClass                                                | kong                                                                         |
--- a/charts/kong/templates/deployment.yaml
+++ b/charts/kong/templates/deployment.yaml
@@ -263,6 +263,8 @@ spec:
 {{ toYaml .Values.readinessProbe | indent 10 }}
         livenessProbe:
 {{ toYaml .Values.livenessProbe | indent 10 }}
+        startupProbe:
+{{ toYaml .Values.startupProbe | indent 10 }}
         resources:
 {{ toYaml .Values.resources | indent 10 }}
         {{- end }} {{/* End of Kong container spec */}}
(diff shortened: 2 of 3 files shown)


rainest:
Interesting. I don't think we should block liveness on reading resources (we should block readiness, but that's fine) initially, though I haven't looked in the code yet to see why that's the case. A controller-side fix is probably a better option here.

Edit: which version are you using? I'm unsure offhand what we did in 1.x, but in 2.x this relies on standard controller-runtime functionality to specify a basic "return success if started at all" function for `/healthz` and start it immediately. The other functions we bind to watch K8S resources should not block that healthz check.

That said, per https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/ this is GA as of 1.20. We generally only enable features once they hit GA since our experience with managed providers is that they don't usually have beta features on by default and may not even allow you to enable them.

Since we still maintain support for older versions, this should have a [KubeVersion](https://helm.sh/docs/chart_template_guide/builtin_objects/) gate so that it only populates on versions >= 1.20, and the values comment/README should indicate as much.

danibaeyens:
Just to add some more context, it's related to this discussion: https://github.com/Kong/kubernetes-ingress-controller/discussions/2170

I'll have a look at your comment and modify the PR

rainest:
Ah, gotcha, it's still controller 1.2. Without reviewing the old code in more detail, I'm guessing the liveness handler there is indeed blocked on a successful read of the configuration.

As of 2.x we've replaced a lot of our bespoke generic functionality with [controller-runtime](https://github.com/kubernetes-sigs/controller-runtime), and its liveness handler should just start immediately. Do you still observe the issue if you upgrade/is there any reason you wouldn't be able to use 2.x yet? I'd probably recommend that instead, since ideally we want to avoid additional config surface in the chart unless it's truly necessary.

rainest:
Minor note if 2.x doesn't end up fixing this: we've switched away from using the next branch, and everything goes into main. I've retargeted the PR but you probably still want to rebase on your end.

ubergesundheit:
@rainest is this something you would still accept a PR for?

PR #540, opened 2022-02-05
feat(conf) add sidecar containers to migration pods

Mirdrack:
The side containers are only included in the main deployment
definition, these updates include the sidecar containers definition
into the migration pods definitions allowing the implementation of the
sidecar pattern on the migration handleiding

Now sidecars will be able for migrations, pre-upgrade and post-upgrade
containers.

None

This PR was motivated for being unable to run the migrations on a system that is using a sql proxy as a sidecar for the database connection
Sidecar containers are already documented but not sure if I should add an extra note about this

#### Checklist


rainest (changes requested):
This needs to be a separate `migrations.sidecarContainers` key (with accompanying changelog and readme entries), since not all sidecars are appropriate for the Jobs. If the sidecar runs indefinitely, the Job will not actually complete. I'm actually surprised that the SQL proxy wouldn't have the same issue; do you have a means of terminating it once the main container is done?

Can you also test cases for these to ci/test2-values.yaml?

```
deployment:
  kong:
    sidecarContainers:
      - name: sidecar
        image: gcr.io/kubernetes-e2e-test-images/echoserver:2.2
migrations:
  sidecarContainers:
     - name: sidecar
       image: bash:4.4
       args: ["true"]
```

Mirdrack:
Thank you @rainest you helped me realize one of my mistakes running the migrations. Indeed, the proxy sidecar container never stops and the deployment of the chart fails.
Regardless, I think that it would be nice to have the option to include sidecars on the migration containers with the disclaimer that they should be finite.

rainest (approved):

PR #550, opened 2022-02-21
fix(helpers) handle istio request authz to proxy

### LINKED ISSUE #549: Unable to reach Kong proxy service when using Istio
Istio's envoy proxy fails to forward traffic to Kong Helm Chart proxy service with message "Upstream connect error or disconnect/reset before headers. reset reason: connection termination" [^1] when access control is enforced with Istio's Authorization Policies [^2].

The issue seems to be caused by Istio's inability to determine the network protocol: Istio expects the Service's port name to be prefixed with the protocol (e.g. "http-kong-proxy" or "tls-kong-proxy"), but this is not happening for Kong Helm Chart [^3][^4].

[^1]: https://discuss.istio.io/t/upstream-connect-error-or-disconnect-reset-before-headers-reset-reason-connection-termination/4434
[^2]: https://istio.io/latest/docs/reference/config/security/authorization-policy/
[^3]: https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/
[^4]: https://github.com/Kong/charts/blob/main/charts/kong/templates/_helpers.tpl#L115

ghost:
Kong proxy service is unreachable from Istio's envoy proxy when access control is enforced with Istio's Authorization Policies. This happens because Istio relies on the Service resource port name to determine the protocol. Issue is solved by prefixing the Service resource port names with the protocol name i.e. tls- and http-.

*(optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close that issue when PR gets merged)*
  - fixes #549

#### Checklist

### DIFF
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -156,6 +156,7 @@ spec:
   - name: kong-{{ .serviceName }}
     port: {{ .http.servicePort }}
     targetPort: {{ .http.containerPort }}
+    appProtocol: http
   {{- if (and (or (eq .type "LoadBalancer") (eq .type "NodePort")) (not (empty .http.nodePort))) }}
     nodePort: {{ .http.nodePort }}
   {{- end }}
@@ -166,6 +167,7 @@ spec:
   - name: kong-{{ .serviceName }}-tls
     port: {{ .tls.servicePort }}
     targetPort: {{ .tls.overrideServiceTargetPort | default .tls.containerPort }}
+    appProtocol: tls
   {{- if (and (or (eq .type "LoadBalancer") (eq .type "NodePort")) (not (empty .tls.nodePort))) }}
     nodePort: {{ .tls.nodePort }}
   {{- end }}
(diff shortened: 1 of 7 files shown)


rainest (changes requested):
Can you change this to use https://kubernetes.io/docs/concepts/services-networking/service/#application-protocol instead? https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/ indicates support for it and we'd strongly prefer to use official Kubernetes standards where available.

ghost:
Good point, I completely missed that. Done!

charts/kong/templates/_helpers.tpl
```diff
@@ -166,6 +167,7 @@ spec:
   - name: kong-{{ .serviceName }}-tls
     port: {{ .tls.servicePort }}
     targetPort: {{ .tls.overrideServiceTargetPort | default .tls.containerPort }}
+    appProtocol: tls
```
rainest:
```suggestion
    appProtocol: https
```

rainest (changes requested):

charts/kong/templates/_helpers.tpl
ghost: Sure. By the way, this is actually the same for Istio, I just wanted to make it consistent with the existing identifiers

> TLS Encrypted data. Because the Sidecar does not decrypt TLS traffic, this is the same as tls

rainest (approved):

PR #552, opened 2022-02-23
fix(conf) fix creation of non-default IngressClasses

ONordander:
When using a custom IngressClass we should check for existence of that class, not the default 'kong' one.

If a user wants to have two Kong instances installed with different IngressClasses the second installation will not actually create the IngressClass since it checks for the existence of the 'kong' IngressClass.

Tested with:
```yaml
ingressController:
  ingressClass: test-kong
```
and
`helm template --validate -f values.yaml kong -n kong .`
#### Checklist


shaneutt (approved):
This LGTM, would also like a double check from @rainest

PR #556, opened 2022-03-09
fix(helpers) disable keyring in wait-for-db container

fffonion:

The wait-for-db container checks for migration readiness. Since
it starts the full kong cluster, it also runs keyring bootstrap process:
the kong clster in that container generates the initial key,
store in its memory, and writes the key_id to keyring_meta table.
But it doesn’t have enough time to broadcast its key to other kong
container that serves proxy, admin, manager etc. By design,
keyring doesn't automatically generate a new key if there's
anything in the keyring_meta table, even no nodes have
knowledge to it; this it to avoid unnecessary rotation of keys.

This PR skips keyring bootstrap at wait-for-db container, so the
bootstrap can be done in an actual kong cluster.

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -14,7 +14,7 @@
 * Fixed traffic routing from Istio's envoy proxy to Kong proxy when using Istio's AuthorizationPolicy.
   ([#550](https://github.com/Kong/charts/pull/550))
 * Fixed creation of non-default IngressClasses ([#552](https://github.com/Kong/charts/pull/552))
-* Fixed initial key in keyring not generated in a new Kong cluster ([#556](https://github.com/Kong/charts/pull/556))
+* Fixed: wait_for_db no longer tries to instantiate the keyring in Kong Enterprise ([#556](https://github.com/Kong/charts/pull/556))
 
 ## 2.7.0
 
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -14,6 +14,7 @@
 * Fixed traffic routing from Istio's envoy proxy to Kong proxy when using Istio's AuthorizationPolicy.
   ([#550](https://github.com/Kong/charts/pull/550))
 * Fixed creation of non-default IngressClasses ([#552](https://github.com/Kong/charts/pull/552))
+* Fixed initial key in keyring not generated in a new Kong cluster ([#556](https://github.com/Kong/charts/pull/556))
 
 ## 2.7.0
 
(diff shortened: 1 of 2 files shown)


charts/kong/templates/_helpers.tpl
```diff
@@ -538,9 +538,15 @@ The name of the service used for the ingress controller's validation webhook
...
+  {{ $waitForDBEnterpriseEnvs := "true" }}
+  {{- if .Values.enterprise.enabled }}
+  {{/* unconditionally disable keyring at wait-for-db container, so that it doesn't
+       generate a bootstrap key without acknowledge any other kong cluster */}}
+    {{- $waitForDBEnterpriseEnvs = "export KONG_KEYRING_ENABLED=off" }}
+  {{- end }}
 {{/* TODO the prefix override is to work around https://github.com/Kong/charts/issues/295
      Note that we use args instead of command here to /not/ override the standard image entrypoint. */}}
-  args: [ "/bin/sh", "-c", "export KONG_NGINX_DAEMON=on; export KONG_PREFIX=`mktemp -d`; until kong start; do echo 'waiting for db'; sleep 1; done; kong stop"]
+  args: [ "/bin/sh", "-c", "export KONG_NGINX_DAEMON=on; export KONG_PREFIX=`mktemp -d`; {{ $waitForDBEnterpriseEnvs }}; until kong start; do echo 'waiting for db'; sleep 1; done; kong stop"]
```
mflendrich: Is there a reason to suppress adding `KONG_KEYRING_ENABLED=off` in non-enterprise?
I think that we should optimize for template complexity rather than for minimizing the set of options passed to Kong.

Suggestion: drop `$waitForDBEnterpriseEnvs` and the `.Values.enterprise.enabled` condition, and inline the `KONG_KEYRING_ENABLED=off` alongside other env vars already there.

mflendrich: readability suggestion: 
now that we're adding a third env var, we can replace `export A=B; export C=D; export E=F;` with a single call: `export A=B C=D E=F;`

charts/kong/CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 * Fixed traffic routing from Istio's envoy proxy to Kong proxy when using Istio's AuthorizationPolicy.
   ([#550](https://github.com/Kong/charts/pull/550))
 * Fixed creation of non-default IngressClasses ([#552](https://github.com/Kong/charts/pull/552))
+* Fixed initial key in keyring not generated in a new Kong cluster ([#556](https://github.com/Kong/charts/pull/556))
```
mflendrich:
```suggestion
* Fixed: wait_for_db no longer tries to instantiate the keyring in Kong Enterprise ([#556](https://github.com/Kong/charts/pull/556))
```

fffonion:
Thanks for the review @mflendrich, I have addressed the comments, could you take another look : )?

mflendrich (approved):

PR #559, opened 2022-03-16
update template to allow wildcard domain names

hanfi:
Small change to template to allow using wildcard domain names (ex: *.example.com)

- quoted the hostname value to be able to set a wildcard hostname

#### Checklist
- [x] Commits follow the [Kong commit message guidelines]

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* add quote on hostname template to be able to use wildcard domains.
+
 ### Improvements
 
 * Enables the option to add sidecar containers to the migration containers.
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -84,7 +84,7 @@ spec:
   ingressClassName: {{ .ingress.ingressClassName }}
 {{- end }}
   rules:
-  - host: {{ $hostname }}
+  - host: {{ $hostname | quote }}
     http:
       paths:
         - backend:
@@ -106,7 +106,7 @@ spec:
   {{- if (hasKey .ingress "tls") }}
   tls:
   - hosts:
-    - {{ $hostname }}
+    - {{ $hostname | quote }}
     secretName: {{ .ingress.tls }}
   {{- end -}}
 {{- end -}}


rainest (commented):
Can you clarify what prompted this change?

The Ingresses we generate from the template are standard Ingresses, not Knative Ingresses; they should not require any changes to avoid issues with our Knative implementation.

The ClusterIngress resource [was removed from Knative quite some time ago](https://github.com/knative/serving/issues/5024) and we do not have/never had an implementation of it in the controller code, so there shouldn't be any reason to add those permissions.

hanfi:
> Can you clarify what prompted this change?

i was trying to make knative work with kong ingress controller. and those are the changes that i thought i needed to do to make it work.

> The Ingresses we generate from the template are standard Ingresses, not Knative Ingresses; they should not require any changes to avoid issues with our Knative implementation.

the point is that knative routes are as specified {{servicename}}.{{namespace}}.{{domainname}} i wanted that the proxy ingress hostname can be like *.example.com and since the ingress hostname is not quoted you can't create an ingress with asterisk at the beginning.

it's useful also when you use the chart without the controller and with external DNS  simply need to set *.example.com as your proxy ingress and setup via api all your custom routes in kong.

the need is just to be able to set something starting with an asterisk in the hostname of the ingress and the tls domains.

> The ClusterIngress resource https://github.com/knative/serving/issues/5024 and we do not have/never had an implementation of it in the controller code, so there shouldn't be any reason to add those permissions.

i was following this answer for Ambassador https://stackoverflow.com/a/60854616 -> you are right i'm deleting those permissions since they disappeared from knative serving. (i was trolled by my knative that only worked once i changed this 😢  )


***i cleaned my PR and rename everything to make it only related to wildcard domain names usage with ingresses.*** 


thank you and sorry for my mistake 😄

rainest:
In general, I would recommend against using the proxy Ingress. It's essentially a holdover from a very old design of the chart that didn't align with the intended use of Ingress. We [intentionally removed](https://github.com/Kong/charts/blob/main/charts/kong/UPGRADE.md#removal-of-multi-host-proxy-ingress) functionality similar to what you're describing because Ingresses should be specific to the application they configure, and advise anyone that does want to use Ingress that way to manage their own independent of the chart.

In normal usage, Knative Services will create routes for their specific hostname, e.g. Service `foo` will create a `foo.example.com` route. You can still set external DNS for `*.example.com` to point to the proxy IP to route this traffic; it's not necessary to have an Ingress for the wildcard hostname.

Having that entry in the proxy Ingress should only be necessary if you're chaining controller-managed proxies together, e.g. if you place a controller-manged ALB in front of Kong with a blanket rule to direct all traffic to Kong. Such configurations are outside the scope of what we intend to support in the chart/are one of the cases where we advise managing the proxy Ingress outside it.

Does that clarify the intended usage/edge cases that the chart doesn't support. Do you have some other environment setup outside those that we should consider?

chrbo:
Having that entry in the proxy Ingress should only be necessary if you're chaining controller-managed proxies together, e.g. if you place a controller-manged ALB in front of Kong with a blanket rule to direct all traffic to Kong. Such configurations are outside the scope of what we intend to support in the chart/are one of the cases where we advise managing the proxy Ingress outside it.
> @rainest Enable a in front placed ALB to route its traffic with a wildcard hostname rule to kong is exactly what you can very easily achieve with the provided change. If this (in my opinion very common use case) is out of scope of the intended usage of the chart why is it then at all possible to generate an incoming ingress? On the other hand, it is generally not a bad practice to quote values and charts should be a swiss army knife. Otherwise you end up with a huge toolbox and no tools fit right. 
> @hanfi You can still achieve what you want with an additional ALB annotation: You have to unset the [proxy.ingress.hostname](https://github.com/Kong/charts/blob/e4c7f914a31c7090e59b254b7915dcd68860b468/charts/kong/values.yaml#L287) value and set the [proxy.ingress.path](https://github.com/Kong/charts/blob/e4c7f914a31c7090e59b254b7915dcd68860b468/charts/kong/values.yaml#L291) to /*. The inbound kong ingress listens now to all hostname and all paths. To set the wildcard hostname rule to the ALB you can use the [alb.ingress.kubernetes.io/conditions.${conditions-name}(https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/guide/ingress/annotations/#traffic-routing) annotation. Example: ```alb.ingress.kubernetes.io/conditions.kong-ingress-kong-proxy: [{"field":"host-header","hostHeaderConfig":{"values":["*.example.com"]}}]```

hanfi:
i'll explain why i'm using it this way : 
- ingresses creates route54 (dns) entries for me, so usually i hava a new service i need to deploy on k8s it has it's own ingress (i think everybody does the same). i also installed knative on the same cluster, so i need an ingress where i can set *.mydomain.com so externalDns creates the dns entry to my loadbalancer then it will be routed internally via kong to the matching service.

the whole PR is quoting the host value so you can add a value like *.mydomain.com if you want.
normally in a standard helm chart (a simple helm create chart) you can add *.domain.com if you just set it quoted on vaues.yaml, but in the case of kong chart since the hole ingress yaml is built inside helper template, the quotes that you set on the values are lost and when rendering the chart you find yourself with an unquoted value for  hostname: *.mydomain.com ,  to which yaml fail :/ 

i'm sure it's not related to my use case or kong usage, i understand the kong helm chart guideline choices. 

but i think that quoting the hostname string is usefull sometimes for a lot of persons.

rainest (approved):
Gotcha--presumably Route53 isn't picking up the individual hostnames because it's only looking at standard Ingresses, whereas Knative uses its own Knative group Ingresses.

Not sure if that's built-in functionality or using ExternalDNS. Knative previously had docs for the latter, though it looks like they were inadvertently deleted during a reorg: https://github.com/ZhiminXiang/docs-1/blob/22ea1955f7979dfb1aadc239b2a892457420e8f7/serving/using-external-dns.md

If using ExternalDNS, using an `external-dns.alpha.kubernetes.io/hostname: hostname` annotation on the proxy Service would be my recommendation if creating the DNS record is the only goal.

That said, the actual change to just quote the hostnames isn't harmful, so I'll go ahead and merge this.

PR #564, opened 2022-03-22
feat(pathType) support user-defined ingress pathType

vzell:
Add support for user-defined ingress pathType. Right now it was hard-coded to be `ImplementationSpecific`.
According to https://kubernetes.io/docs/concepts/services-networking/ingress/#path-types the values can be one of
`ImplementationSpecific`, `Exact` or `Prefix`.

  - fixes #

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Improvements
 
+* Ingress `pathType` field is now configurable.
+  ([564](https://github.com/Kong/charts/pull/564))
 * Added IngressClass resources to RBAC roles.
   ([563](https://github.com/Kong/charts/pull/563))
 * Ingresses now support wildcard hostnames.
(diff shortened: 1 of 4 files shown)


rainest:
@vzell can you share what prompted this? For the various Kong services, I'd expect that you only ever want prefix behavior, since they all expose many more endpoints than their Ingress lists (e.g. for the admin API you want to match requests for both `/` and `/routes`).

ImplementationSpecific for Kong is de facto prefix (though it does not require a trailing slash, which the Ingress Prefix spec does), and as far as we know that works fine for all our services, so it was originally omitted to reduce the number of configuration options, with the expectation that there'd be no reason to change it in practice.

vzell:
@[rainest](https://github.com/rainest) Our use case on Amazon EKS only works with pathType `Prefix`. Until now it was hardcoded to be `ImplementationSpecific` and I could not overwrite it which prompted me to the patch.

rainest (changes requested):
Sounds reasonable--to clarify, you're handling them through the ALB controller instead of ours then?

If you can the following to the changelog under the Unreleased > Improvements section we should be all set:

```
* Ingress `pathType` field is now configurable.                                                                                             
  ([564](https://github.com/Kong/charts/pull/564))
```

vzell:
Yes, we are handling them through the ALB controller...

Added to CHANGELOG

rainest (approved):

PR #573, opened 2022-04-05
feat(addOpenShiftSupport) allow adding OCP routes

FabianHardt:
Add support for generating OCP routes to expose API gateway with
standard OCP routes.

This feature adds the possibility to generate OpenShift routes, to expose API gateway or Kong manager.

  - fixes #

#### Checklist


rainest:
We don't intend to add OpenShift Routes to the chart to limit the number of configuration decisions needed and number of templates we need to maintain, since OpenShift [supports Ingress also](https://cloud.redhat.com/blog/kubernetes-ingress-vs-openshift-route), including support for [Ingress annotations that implement some of their features via auto-generated Routes](https://docs.openshift.com/container-platform/4.9/networking/routes/route-configuration.html#nw-ingress-creating-a-route-via-an-ingress_route-configuration).

Future work on our end will focus on supporting SIG-Network's [Gateway APIs](https://gateway-api.sigs.k8s.io/), an upcoming standard that addresses similar concerns.

That aside, it should be possible to largely handle this through an umbrella chart that includes your templates: the new templates are largely independent of existing ones, so you'd be able to change the inputs to use `kong.<values path>` via a subchart and continue to use helper functions (`kong.fullname` and such normally). The one exception are the hint variable (`KONG_ADMIN_API_URI` and such), which you'd need to set manually--you can add those to `kong.env` to override anything we set automatically, though unfortunately there's no way to template those to always match the generated Route with this approach.

Did you have any questions regarding that approach?

FabianHardt:
Thank's for your answer. You're right, OpenShift supports Ingress and I could use the implicit route creation. But in this case I will need an Ingress defintion for cluster and clustertelemetry resources as well.
Is it possible to add the following code snippet in the templates _service-kong-cluster-telemetry.yaml_ and _service-kong-cluster.yaml_?
```
{{ if .Values.<SVC>.ingress.enabled }}
---
{{ include "kong.ingress" $serviceConfig }}
{{- end -}}
```
Then I will also get two OpenShift routes for that. So I can communicate between CP and DP, which are located in two different clusters.
If this would be fine for you as well, I could change this pull request. Otherwise I will use the umbrella chart as you proposed.

Support of the new Gateway-API would be a great way I think. This might be a very nice feature (I'm looking foreward to that).
Thank's for your ideas!

FabianHardt:
Will be closed as #583 has been implemented and merged.

PR #577, opened 2022-04-08
docs(mesh) add document mesh annotation overwrites

### LINKED ISSUE #547: Document mesh annotation overrides
Follow-up from #522 

Some configurations should _not_ have mesh proxying disabled, which is [the default](https://github.com/Kong/charts/blob/c2bee5c467ca0336bb7e426835f98a473abb15f2/charts/kong/values.yaml#L627-L628).

Handling the override correctly is non-intuitive because you cannot simply delete these annotations unless you use the empty set, because Helm will merge the keys from the default values.yaml. To override them, you must explicitly set those keys, e.g. to use the opposite of the default configuration:

```
  kuma.io/gateway: disabled
  traffic.sidecar.istio.io/includeInboundPorts: "*"
```

FabianHardt:
Add documentation from mesh annotation overwrites

  - fixes #547 

#### Checklist


charts/kong/README.md
```diff
@@ -772,6 +772,16 @@ containerSecurityContext: # run as root to bind to lower ports
...
 
+**Note:** The default pod annotations in this chart disables mesh proxing for Kuma and istio mesh.
+If you don't want to run Kong with disabled proxing inside this meshes, you can tweak the `podAnnotations` configurations as in the example:
+
+```yaml
+# Enable mesh proxing for kuma and istio
+podAnnotations:
+  kuma.io/gateway: disabled
+  traffic.sidecar.istio.io/includeInboundPorts: "*"
+```
```
rainest: Can you clarify the intent here/environments you'd expect to use this configuration, and and share any test results you have for it?

The standard configuration disables only inbound mesh proxying (`gateway` is basically [Kuma's alias for exclude all inbound](https://kuma.io/docs/1.6.x/reference/kubernetes-annotations/#kuma-io-gateway)). Outbound proxying is still enabled, and as far as we know that makes sense for our expected topology where Kong serves as an edge gateway.

The explanation here suggests that the defaults disable the mesh proxies entirely. I expect that the intent here is to provide a configuration for topologies where Kong is somewhere in the center of the cluster network, but the explanation doesn't indicate as much.

FabianHardt: You're right. Here's a new attempt (99c715d425a0dd535f9aba0265c07602c9c8371d), with a somewhat more detailed explanation of these annotations.

charts/kong/README.md
```diff
@@ -772,6 +772,22 @@ containerSecurityContext: # run as root to bind to lower ports
...
+external Envoy listener for the service and making it resolvable inbound. This is the typical usecase when you
+want to use Kong as a “doorway” into the service mesh.
+
+If you want to use Kong gateway as a mesh proxy with advanced plugin functionality, you should disable the
+Kuma and istio gateway mode (as in the following example). In this modes the gateway is only reachable inside the
+service mesh with it's matching _.mesh_ address. The equivalent annotation for istio service mesh is
+`traffic.sidecar.istio.io/includeInboundPorts: "*"`, which redirects all incoming traffic to envoy first.
+
+```yaml
+# Enable mesh proxing (mesh to mesh) for kuma and istio
```
rainest: Brevity, capitalization, and removal of the `.mesh` mention (Kuma-specific and only for multi-zone meshes)

```suggestion
**Note:** The default `podAnnotations` values disable inbound proxying for Kuma 
and Istio. This is appropriate when using Kong as a gateway for external 
traffic inbound into the cluster.

If you want to use Kong as an internal proxy within the cluster network, you 
should enable inbound the inbound mesh proxies:

```yaml
# Enable inbound mesh proxying for Kuma and Istio
```

rainest (changes requested):

rainest (approved):

PR #578, opened 2022-04-08
#Suport ingress controller with configmap

tquocthang97:
#Suport ingress controller with configmap

  - fixes #

#### Checklist

### DIFF
--- a/charts/kong/templates/config-dbless.yaml
+++ b/charts/kong/templates/config-dbless.yaml
@@ -1,5 +1,5 @@
 {{- if .Values.deployment.kong.enabled }}
-{{- if (and (not .Values.ingressController.enabled) (eq .Values.env.database "off")) }}
+{{- if (and (.Values.ingressController.enabled) (eq .Values.env.database "off")) }}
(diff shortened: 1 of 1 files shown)


rainest (commented):
@tquocthang97 can you clarify the expected use case here? As far as we know, loading DB-less configuration from a ConfigMap only makes sense when not using the controller, because the controller is the source of configuration when it's enabled.

The controller would overwrite any configuration in the ConfigMap shortly after startup, so the ConfigMap config would not serve any apparent purpose.

tquocthang97:
@rainest Yah,
I want deploy Kong Ingress Controller using Daemonset -DB-less and  use Declarative Configuration
 File in ENV options to load config when pod start.  In my use case, i want load  some plugin when pod start deploy with helm, and no use kubectl apply some plugin manual.
 Because i want add "tcp-log-plugin-" and set host-port to load plugin

rainest:
You understand that configuration will be overwritten by the controller more or immediately, correct? Why would you want to load something you'll never actually be able to use?

PR #583, opened 2022-04-14
feat(services) add Ingress for cluster sync

FabianHardt:
This is an alternative pull request to #573 . This one uses the default Ingress feature of K8s to create OpenShift routes. I just need to add Ingress option for cluster communication as well, that's why I edited the matching service templates. I need this for automatic creation of OpenShift routes as I need them for cluster to cluster communication between CP and DP.
I also had to remove an unnecessary if statement, because ingress.path already has a default value in values.yml. I need the option to be able to leave the ingress.path blank.

As I said, this is an alternative to #573 and I would be happy if we could find a common solution here.

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -7,6 +7,8 @@
 
 ### Improvements
 
+* feat(services) add Ingress for cluster sync
+  ([583](https://github.com/Kong/charts/pull/583))
 * feat(deployment) allow custom environment variables to be configured for the ingress-controller container
   ([568](https://github.com/Kong/charts/pull/568))
 * Ingress `pathType` field is now configurable.
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -98,12 +98,10 @@ spec:
               port:
                 number: {{ $servicePort }}
             {{- end }}
-          {{- if $path }}
           path: {{ $path }}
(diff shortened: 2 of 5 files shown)


rainest (approved):
What does including an empty path do in your case?

It's unclear why we had the `if` clause given that all of our configuration sets the field by default in values.yaml and v1 Ingresses without it were invalid. However, they were invalid because `pathType` was absent, not because `path` was. Not including `path` is valid, but a bit unclear since you can't really issue an HTTP request without one.

Best guess is that the original template was somehow relevant to extensions/v1beta1 Ingress if anything, but can't really test that anymore for lack of a readily available test instance that supports it. That's old and not of clear use there either, so not too worried about it.

FabianHardt:
Thank's for the quick merge!
With this possibility to leave the _path_ empty, I'm able to get auto generated OpenShift routes in _passthrough_ mode as described [here](https://docs.openshift.com/container-platform/4.10/networking/routes/route-configuration.html#nw-ingress-creating-a-route-via-an-ingress_route-configuration). I tired it, but if _path_ attribute is not empty, OpenShift doesn't generate a passthrough route automatically.
Combined with new Ingress resources for cluster and clustertelemetry I'm able to deploy my existing OpenShift hybrid mode setup completely with this chart. Everything works fine now, thanks!
If you're interested in an OpenShift example _values.yml_ please let me know.

PR #584, opened 2022-04-17
Feat/generate postgres secret

### LINKED ISSUE #214: Test and document existingSecret for Postgres passwords
When using the Postgres sub-chart, our chart defaults to creating a random password. Upgrading the chart then leads to https://github.com/Kong/charts/blob/master/charts/kong/FAQs.md#kong-fails-to-start-after-helm-upgrade-when-postgres-is-used-what-do-i-do

Setting `postgresql.postgresqlPassword` avoids this, but poorly, as it leaves a plaintext password in values.yaml. Using `postgresql.existingSecret` should be better, but it requires a specific Secret format. From initial research on the Bitnami Postgres chart:

```
$ grep -r key: .
./templates/statefulset-slaves.yaml:                  key: postgresql-replication-password
./templates/statefulset-slaves.yaml:                  key: postgresql-postgres-password
./templates/statefulset-slaves.yaml:                  key: postgresql-password
./templates/statefulset.yaml:                  key: postgresql-replication-password
./templates/statefulset.yaml:                  key: postgresql-postgres-password
./templates/statefulset.yaml:                  key: postgresql-password
./templates/statefulset.yaml:                  key: postgresql-ldap-password
./templates/statefulset.yaml:                  key: postgresql-password
```
We need to conduct further research to understand the the different keys' use and special requirements (if any) for their contents. Most will likely just be random strings, but we should provide instructions for creating these (we can't rely on Helm functionality for it anymore; https://unix.stackexchange.com/a/230676 should be a decent source of inspiration).

We may also need new tooling to set `KONG_PG_PASSWORD` from the existingSecret Secret.

FabianHardt:
Changes the default behaviour of setting the postrges password. As described in #214 the _existingSecret_ attribute is used. The matching secret is generated by this chart and won't get overwritten when an upgrade is done.

  - fixes #214

It's a kind of a breaking change, but the actual default behaviour also, because it generates a new password everytime the chart runs. I don't know if my documentation for that is sufficient. Please give me a hint about this.

#### Checklist


rainest:
Do you know if this is actually still needed in any scenario? I think our upgrade to the latest Bitnami chart version obsoleted it. Per https://github.com/bitnami/charts/pull/4416#issuecomment-1094880576 part of that was migrating to a lookup-based password Secret manager, and a brief test indicates that it's indeed no longer replaced on upgrade: 

```
11:13:50-0700 esenin $ kubectl get secrets aaa-postgresql -oyaml
apiVersion: v1                 
data:
  password: eEQ0ME9lZU8xWA==
  postgres-password: RkpiYVFqVVpYWg==
kind: Secret
metadata:
  annotations:
    meta.helm.sh/release-name: aaa
    meta.helm.sh/release-namespace: default
  creationTimestamp: "2022-04-18T18:12:35Z"
  labels:
    app.kubernetes.io/instance: aaa
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: postgresql
    helm.sh/chart: postgresql-11.1.15
  name: aaa-postgresql
  namespace: default
  resourceVersion: "1493"
  uid: ecd99333-f063-4a35-928d-4ea706ac7467
type: Opaque

11:13:57-0700 esenin $ helm upgrade aaa kong/kong -f /tmp/values.yaml       
Release "aaa" has been upgraded. Happy Helming!
NAME: aaa
LAST DEPLOYED: Mon Apr 18 11:14:09 2022
NAMESPACE: default
STATUS: deployed
REVISION: 2
NOTES:
...

11:14:14-0700 esenin $ kubectl get secrets aaa-postgresql -oyaml     
apiVersion: v1
data:
  password: eEQ0ME9lZU8xWA==
  postgres-password: RkpiYVFqVVpYWg==
kind: Secret
metadata:
  annotations:
    meta.helm.sh/release-name: aaa
    meta.helm.sh/release-namespace: default
  creationTimestamp: "2022-04-18T18:12:35Z"
  labels:
    app.kubernetes.io/instance: aaa
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: postgresql
    helm.sh/chart: postgresql-11.1.15
  name: aaa-postgresql
  namespace: default
  resourceVersion: "1493"
  uid: ecd99333-f063-4a35-928d-4ea706ac7467
type: Opaque
```
So I think all we need to do now is mark that the FAQ only applies to Kong chart versions <2.8.

FabianHardt:
You're right, it's fixed for the update case in the new bitnami chart. But the annotation _"helm.sh/resource-policy": "keep"_ isn't used for the secret, so you can't do a _helm delete_ and reinstall the chart like you often do in development situations.
I think with explicit secret generation it's more robust than the original way of bitnami. But it's just my personal opinion.
It's no problem if you won't merge it, cause I still could use explicit secret generation in my use case.

rainest:
The upgrade case was more of a concern, since it meant losing something you didn't intend to (access to the current DB you're still trying to use) with no way to recover. The delete one less so, since that's instead something you should have deleted but didn't. The old PVC that breaks subsequent installs should be deleted regardless since it will also contain old DB rows and schemas, not just the previous password, so closing this in favor of improving that doc in https://github.com/Kong/charts/pull/586

PR #592, opened 2022-05-06
Feature proxy and mtls with cert-manager

### LINKED ISSUE #591: [RFE] Add support for issuing mtls and proxy certs via cert-manager
Enhancement request for cert-manager support

usrbinkat:
This enables use of cert-manager to issue default proxy and hybrid cluster mtls certs. 

  - fixes #591 

This enhancement is based on high value enhancement needs observed during field work hands on customer testing.

This enhancement also supports [work in progress docs enhancement](https://github.com/Kong/docs.konghq.com/pull/3886).

#### Checklist


shaneutt:
I marked this draft because the title indicated it was still a draft and I see that linters were failing. Please let @Kong/team-k8s know if you need any help moving this one forward!

charts/kong/README.md
```diff
@@ -19,6 +19,9 @@ $ helm install kong/kong --generate-name
 
 ## Table of contents
 
+- [Kong for Kubernetes](#kong-for-kubernetes)
+- [TL;DR;](#tldr)
+- [Table of contents](#table-of-contents)
```
rainest: We don't want the prelude or a link back to the TOC in the TOC either. Does whatever TOC generator you use have a means of excluding sections?
```suggestion
```

charts/kong/values.yaml
```diff
@@ -581,6 +581,68 @@ postgresql:
...
+#     the cluster mtls shared certificate for use in kong hybrid and multiple-dataplane
+#     scenarios.
+#   - Kong pre-prod & local development where the self-signed cert-manager issuer
+#     is leveraged for developer and testing workflows only. In this scenario it
+#     is assumed that valid certs will be utilized in production for public facing
+#     traffic.
+
+# NOTICE: The cert-manager cluster cert integration generates a self signed
+# cluster issuer certificate authority for hybrid cluster mtls when enabled.
+
```
rainest: Comments in values should ideally be brief, to describe the use of individual keys or show example objects, not complete copies of the readme docs.

```suggestion
# Configure cert-manager integration
```

charts/kong/templates/cluster-certificate.yaml
```diff
@@ -0,0 +1,21 @@
...
+# Self Signed Kong Cluster Certificate
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: kong-cluster
+  labels:
+    {{- include "kong.metaLabels" . | nindent 4 }}
+spec:
+  secretName: kong-cluster
```
rainest:
```suggestion
  secretName: {{ include "kong.fullname" . }}-kong-cluster
```

charts/kong/templates/proxy-certificate.yaml
```diff
@@ -0,0 +1,24 @@
...
+# Kong Proxy Certificate
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: kong-cert
+  labels:
+    {{- include "kong.metaLabels" . | nindent 4 }}
+spec:
+  secretName: kong-cert
```
rainest:
```suggestion
  secretName: {{ include "kong.fullname" . }}-kong-cert
```

charts/kong/templates/cluster-certificate.yaml
```diff
@@ -0,0 +1,21 @@
+{{- if and ( .Capabilities.APIVersions.Has "cert-manager.io/v1" ) .Values.certificate.cluster.enabled }}
+# Self Signed Kong Cluster Certificate
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: kong-cluster
```
rainest: Generated resource names should be such that they cannot conflict with other releases' resources. We don't require a dedicated namespace per release.

We also add the namespace explicitly to include it when rendered through `helm template` instead of `helm install`.

```suggestion
  name: {{ include "kong.fullname" . }}-kong-cluster
  namespace:  {{ template "kong.namespace" . }}
```

charts/kong/templates/proxy-certificate.yaml
```diff
@@ -0,0 +1,24 @@
+{{- if and ( .Capabilities.APIVersions.Has "cert-manager.io/v1" ) .Values.certificate.proxy.enabled }}
+# Kong Proxy Certificate
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: kong-cert
```
rainest:
```suggestion
  name: {{ include "kong.fullname" . }}-kong-cert
  namespace:  {{ template "kong.namespace" . }}
```

charts/kong/templates/proxy-certificate.yaml
```diff
@@ -0,0 +1,24 @@
...
+  labels:
+    {{- include "kong.metaLabels" . | nindent 4 }}
+spec:
+  secretName: kong-cert
+  commonName: kong_clustering
+  dnsNames:
+  - "kong_clustering"
```
rainest: There's no reason this should be included on the proxy certificate, correct?
```suggestion
```

charts/kong/templates/proxy-certificate.yaml
```diff
@@ -0,0 +1,24 @@
...
+  {{- range .Values.certificate.dnsNames }}
```
rainest:
```suggestion
  {{- range .Values.certificate.proxy.dnsNames }}
```

charts/kong/values.yaml
```diff
@@ -581,6 +581,68 @@ postgresql:
...
+    # is not defined with a valid alternative.
+    issuer: ""
+    # Use commonName and dnsNames to set the common name and dns alt names which this
+    # certificate is valid for. Wildcard records are supported by the included self-signed issuer.
+    commonName: "example.com"
+    dnsNames:
+    - "example.com"
+    - "*.example.com"
+    - "*.apps.example.com"
+    - "*.kong.example.com"
```
rainest: Default array values will actually get used, so those should be comment examples with `{}` as the actual default. Using `example.com` versus the `.example` TLD is basically personal preference, but I like the latter a bit more if we're going to use actual strings rather than commenting those out as well (not sure if leaving it empty would break the resulting Certificate resources, but ambivalent since you have to enable this anyway and presumably will notice that you also need to change the CN).

```suggestion
    commonName: "kong.example"
    # Remove the "{}" and uncomment/change the examples to add SANs
    dnsNames:{}
    #- "kong.example"
    #- "*.kong.example"
```

charts/kong/README.md
```diff
@@ -425,6 +432,44 @@ documentation on Service
...
+    and the kong proxy and services directly.
+  - Kong helm chart cert-manager integration is used only for provisioning
+    the cluster mtls shared certificate for use in kong hybrid and multiple
+    dataplane scenarios.
+  - Kong pre-prod & local development where the self-signed cert-manager issuer
+    is leveraged for developer and testing workflows only. In this scenario it
+    is assumed that valid certs will be utilized in production for public
+    facing traffic.
+
+#### NOTICE: The cert-manager cluster cert integration generates a self signed cluster issuer certificate authority for hybrid cluster mtls when enabled.
```
rainest: Ideally we do just leave Issuer management to the user, but header directives as a means to specify appearance are anathema.

```suggestion
**The cert-manager cluster cert integration generates a self signed cluster issuer certificate authority for hybrid cluster mtls when enabled.**
```
If we end up keeping the chart-managed Issuers, any such warnings should move to the top of the section (or at least should be in the middle above some relevant subsection, not at the end).

charts/kong/templates/_helpers.tpl
```diff
@@ -21,6 +21,12 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
 {{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
 
+{{- define "kong.certificate.issuer" -}}
+{{- $default := printf "%s-%s" .Release.Name "kong-selfsigned-issuer" -}}
+{{- $name := default $default .Values.certificate.proxy.issuer -}}
+{{- printf "%s" $name -}}
+{{- end -}}
```
rainest: This should be possible to condense to a single line using parentheses to contain the components, which will make it a bit easier to read what the final result is. I think, roughly, `{{- default (printf "%s-%s" .Release.Name "kong-selfsigned-issuer") .Values.certificate.proxy.issuer -}}`, though not positive.

That said, see elsewhere in the Issuer template re using it consistently.

charts/kong/templates/selfsigned-issuer.yaml
```diff
@@ -0,0 +1,45 @@
...
+  issuerRef:
+    name: {{ .Release.Name }}-kong-clusterissuer-selfsign-root
+    group: cert-manager.io
+    kind: ClusterIssuer
+---
+# Self Signed Issuer
+apiVersion: cert-manager.io/v1
+kind: ClusterIssuer
+metadata:
+  name: {{ template "kong.fullname" . }}-selfsigned-issuer
```
rainest: We shouldn't build names in multiple places, else we can inadvertently change one version and not the other later. When using a template for a name, use it throughout:

```suggestion
  name: {{ template "kong.certificate.issuer" . }}
```
As such, in cases where we may want to instead default to a values.yaml value instead of the generated name, the `default` that handles that should be inside the template that needs to reference the name, rather than in the name template itself.

rainest (changes requested):
- I'm not sure if we should be creating an Issuer. We have, AFAIK, no special requirements for the cluster certificates and do not need separate Issuers per release. I'd expect that most cert-manager users will have an existing Issuer or ClusterIssuer for internal certificates. We also have a significant subset of users that are very opposed to granting permissions to manage cluster-level permissions. Assuming a provided Issuer or ClusterIssuer, do we expect an issues?

  Creating the Issuers would definitely make sense in an entire pre-packaged environment, but the goal of the chart is to target a production environment on a cluster that runs applications other than Kong alongside Kong, with the expectation that common cluster resources (certificate issuers, Prometheus installs, etc.) and related configuration not specific to Kong are managed externally. For example, we provide ServiceMonitors for Prometheus--they're bound to Kong Services--but not AlertmanagerConfig, which is shared across the Prometheus install. We could make an umbrella chart to install all of Kong, things it needs (Postgres, Redis), and things that make it easier to use (Prometheus, cert-manager) along with broader configuration (AlertmanagerConfig and/or ClusterIssuers) for demonstration environments or single-application clusters that do want a pre-made "the works" chart to handle most everything other than their applications, but don't yet.

- This adds a lot of interacting parts and needs at least a documented manual end to end test run to confirm that it works as expected.
- I'd recommend having a top-level `certificate.enabled=false` that disables everything, and then making toggles under it `true` by default but only actually enabled if relevant. Individual certificate templates render conditions should be something like `if certificate.enabled: if certficate.cluster.enabled and (env.role == control_plane or env.role == data_plane) ... end; end;`
- Should we use cert-manager for the admission webhook certificate as well? It seems like we probably should if it's available.
- We should offer certificate provisioning for the various Ingress resources as toggles under their config sections.
- This looks like it's missing the configuration to actually use the certificates in question. We create them, but do not mount them anywhere and set up autoEnv values referencing the mounts or annotate the generated Ingress resources to request one.
- Actual hybrid mode usage presumably has additional challenges because of the separate releases, where you'll either need to reference a Secret managed by only one of those releases (`shared` mode) or an Issuer managed by only one of those releases (or externally) (`pki` mode). The docs should cover that. Not many other comments on docs yet since I expect a fair amount to change.

charts/kong/README.md
usrbinkat: aaaah, just found "auto update TOC on save" was set for a new plugin I've been using. Disabled that nonsense.

charts/kong/templates/_helpers.tpl
usrbinkat: done

usrbinkat:
@rainest 

Documented manual method is in staging on [netlify kong docs previews](https://deploy-preview-3886--kongdocs.netlify.app/gateway/latest/install-and-run/helm-quickstart-enterprise/).

This document currently supports local testing on both kind and docker desktop kubernetes.

This documentation effort is the early work in a larger effort to reduce friction in first impression scenarios which I believe cert-manager adds significant value to achieving. 



> * I'm not sure if we should be creating an Issuer. We have, AFAIK, no special requirements for the cluster certificates and do not need separate Issuers per release.

I converted the ClusterIssuer to be a release namespaced issuer. I cannot create kong hybrid cluster certificates with a CN attribute using the `selfsigned` spec. This attribute requires an issuer.

Additionally, Google Chrome and Firefox both refuse traffic from the dataplane when issuing a kong proxy certificate without the CN attribute and a chain of trust which I also cannot accomodate with the selfsigned spec alone.

I believe this means that some level of issuer is required to use cert-manager for kong as I have seen no way around it yet.

>      I'd expect that most cert-manager users will have an existing Issuer or ClusterIssuer for internal certificates. We also have a significant subset of users that are very opposed to granting permissions to manage cluster-level permissions. Assuming a provided Issuer or ClusterIssuer, do we expect an issues?

For this see comment above where I converted the issuer chain of trust from being a cluster scoped ClusterIssuer type, to being a kong release namespace constrained Issuer type. I believe that resolves most of this.

Also, I exposed a per-certificate ability to explicitly set the preferred cert issuer name. 

While in the field, many of the customers I engaged with have been at the beginning of their container adoption journey and things like educating them on cert-manager for it's value to their kong operations story was rather non-trivial. This enhancement will significantly improve future stories following those same lines.

>       Creating the Issuers would definitely make sense in an entire pre-packaged environment, but the goal of the chart is to target a production environment on a cluster that runs applications other than Kong alongside Kong, with the expectation that common cluster resources (certificate issuers, Prometheus installs, etc.) and related configuration not specific to Kong are managed externally. For example, we provide ServiceMonitors for Prometheus--they're bound to Kong Services--but not AlertmanagerConfig, which is shared across the Prometheus install. We could make an umbrella chart to install all of Kong, things it needs (Postgres, Redis), and things that make it easier to use (Prometheus, cert-manager) along with broader configuration (AlertmanagerConfig and/or ClusterIssuers) for demonstration environments or single-application clusters that do want a pre-made "the works" chart to handle most everything other than their applications, but don't yet.

As a rule, I am opposed to adding bloat and complexity where there is not a significant impact to large numbers of our users and broad use case for the feature. I believe that cert-manager support pairs very well with the existing optional built in postgres support to enable both developer focused ease of use as well as adding value to enterprise scenarios where customers want a streamlined way to leverage their existing cert-manager issuers for kong service certificates. This has been a repeated conversation of difficulty with more advanced customers in the field where they felt our cert-manager ease of use was lacking.

> 
>     * This adds a lot of interacting parts and needs at least a documented manual end to end test run to confirm that it works as expected.
> 
>     * I'd recommend having a top-level `certificate.enabled=false` that disables everything, and then making toggles under it `true` by default but only actually enabled if relevant. Individual certificate templates render conditions should be something like `if certificate.enabled: if certficate.cluster.enabled and (env.role == control_plane or env.role == data_plane) ... end; end;`



> 
>     * Should we use cert-manager for the admission webhook certificate as well? It seems like we probably should if it's available.

Can you point me in the direction of what you are speaking about specifically for this point? I'm not sure I am aware of any remaining certs not addressed in my latest updates.

> 
>     * We should offer certificate provisioning for the various Ingress resources as toggles under their config sections.

Resolved in latest commits

> 
>     * This looks like it's missing the configuration to actually use the certificates in question. We create them, but do not mount them anywhere and set up autoEnv values referencing the mounts or annotate the generated Ingress resources to request one.

I have added an example values.yaml file. I would probably benefit from a short pair programming session to see how to properly implement what you are describing here if I am going to go off and do that on my own.

> 
>     * Actual hybrid mode usage presumably has additional challenges because of the separate releases, where you'll either need to reference a Secret managed by only one of those releases (`shared` mode) or an Issuer managed by only one of those releases (or externally) (`pki` mode). The docs should cover that. Not many other comments on docs yet since I expect a fair amount to change.

This is a good callout. I believe that in more complex architectures we can document that other methods may be required to support more complicated implementations as a start. If opportunity arises with practice in the field for seeing how to accommodate those other needs based on this foundational implementation, then I can see having some time in the field with this may bring those opportunities to light in a clearer fashion.

I will be adding an additional KIC deployment to the doc previews linked above which leverages the name of the secret to be able to use the mtls cert in establishing the hybrid nature of this deployment.

A follow on to this quick start document will include both an enterprise free mode, as well as a more advanced document diving into administrator decisions, hybrid mode implementation practice, and explanation of each of these different moving parts.

rainest (commented):
> some level of issuer is required to use cert-manager for kong

Right, but we don't have any particular requirements for it. _Any_ issuer will do. If none is present, is the Issuer we create meangingfully different from the [example in the cert-manager documentation](https://cert-manager.io/docs/configuration/selfsigned/#bootstrapping-ca-issuers)? I don't think it is, since we're essentially just changing names (and fields that reference those names). Templated resources should generally be resources that require specific information from elsewhere in the chart, or at least resources where we don't expect a general-purpose instance of that resource would work.

Customers new to Kubernetes entirely may not be familiar with tools like cert-manager, but that doesn't mean we need to handle it for them entirely (again, I think that'd make sense in a pre-packaged environment umbrella chart, but that's not what this chart is). Referencing vendor documentation for generic resources is preferable because those docs will be kept up to date better. Advanced users should already have Issuers available, so all they need for the integration is the Certificate resource we provision and use.

> I have added an example values.yaml file. I would probably benefit from a short pair programming session to see how to properly implement what you are describing here if I am going to go off and do that on my own.

It's to handle the stuff you were discussing re limiting the need for duplicate config across values.yaml. If we have the affirmative "yes I want to create this" from somewhere, which `certificate.proxy.enabled == true` is, we should handle the mounts and such for users. Briefly, the [kong.volumes template](https://github.com/Kong/charts/blob/64fb4be81f7b264d12542dea6a72664997adb68b/charts/kong/templates/_helpers.tpl#L404) and [kong.volumeMounts template](https://github.com/Kong/charts/blob/64fb4be81f7b264d12542dea6a72664997adb68b/charts/kong/templates/_helpers.tpl#L476) can contain a predefined mount that goes into the Pod/Container sections you'd expect, and something like [the license environment variable setter](https://github.com/Kong/charts/blob/64fb4be81f7b264d12542dea6a72664997adb68b/charts/kong/templates/_helpers.tpl#L770-L773) can then set `KONG_SSL_CERT` to whatever path you used. I'll set up a time.

charts/kong/values.yaml
```diff
@@ -581,6 +581,65 @@ postgresql:
...
+    # is not defined with a valid alternative.
+    issuer: ""
+    # Use commonName and dnsNames to set the common name and dns alt names which this
+    # certificate is valid for. Wildcard records are supported by the included self-signed issuer.
+    commonName: "kong.example"
+    # Remove the "{}" and uncomment/change the examples to add SANs
+    dnsNames: []
+    # - "manager.kong.example"
+
+  # Set proxy.enabled to true to issue default kong-proxy certificate with cert-manager
```
rspurgeon: Is this a copy & paste error?

```suggestion
  # Set portal.enabled to true to issue a kong dev portal certificate with cert-manager
```

usrbinkat: Thank you. I resolved along with additional cleanup.

usrbinkat:
> * I'm not sure if we should be creating an Issuer. We have, AFAIK, no special requirements for the cluster certificates and do not need separate Issuers per release. I'd expect that most cert-manager users will have an existing Issuer or ClusterIssuer for internal certificates. We also have a significant subset of users that are very opposed to granting permissions to manage cluster-level permissions. Assuming a provided Issuer or ClusterIssuer, do we expect any issue?
>   Creating the Issuers would definitely make sense in an entire pre-packaged environment, but the goal of the chart is to target a production environment on a cluster that runs applications other than Kong alongside Kong, with the expectation that common cluster resources (certificate issuers, Prometheus installs, etc.) and related configuration not specific to Kong are managed externally. For example, we provide ServiceMonitors for Prometheus--they're bound to Kong Services--but not AlertmanagerConfig, which is shared across the Prometheus install. We could make an umbrella chart to install all of Kong, things it needs (Postgres, Redis), and things that make it easier to use (Prometheus, cert-manager) along with broader configuration (AlertmanagerConfig and/or ClusterIssuers) for demonstration environments or single-application clusters that do want a pre-made "the works" chart to handle most everything other than their applications, but don't yet.

 - [x] Removed the built-in issuer so that it is now created out-of-band.

> * This adds a lot of interacting parts and needs at least a documented manual end-to-end test run to confirm that it works as expected.

- [x] Current state of manual testing proceedure is in [netlify docs preview](https://deploy-preview-3886--kongdocs.netlify.app/gateway/latest/install-and-run/helm-quickstart-enterprise/) from [WIP PR #3886](https://github.com/Kong/docs.konghq.com/pull/3886)

> * I'd recommend having a top-level `certificate.enabled=false` that disables everything, and then making toggles under it `true` by default but only actually enabled if relevant. Individual certificate templates render conditions should be something like `if certificate.enabled: if certficate.cluster.enabled and (env.role == control_plane or env.role == data_plane) ... end; end;`

- [x] This is now implemented.

> * We should offer certificate provisioning for the various Ingress resources as toggles under their config sections.

- [x] This is now implemented.

> * This looks like it's missing the configuration to actually use the certificates in question. We create them, but do not mount them anywhere and set up autoEnv values referencing the mounts or annotate the generated Ingress resources to request one.

- [x] This is now implemented.

> * Actual hybrid mode usage presumably has additional challenges because of the separate releases, where you'll either need to reference a Secret managed by only one of those releases (`shared` mode) or an Issuer managed by only one of those releases (or externally) (`pki` mode). The docs should cover that. Not many other comments on docs yet since I expect a fair amount to change.
> * Should we use cert-manager for the admission webhook certificate as well? It seems like we probably should if it's available.

- [ ] I believe we decided to leave this for future enhancement as the go templating required was beyond scope of this PR

mheap:
@rainest Would we be able to merge this then revisit the meta-chart idea? It's blocking some other work we'd like to move on

charts/kong/templates/_helpers.tpl
```diff
@@ -406,6 +422,16 @@ The name of the service used for the ingress controller's validation webhook
...
+{{- if .Values.clusterCaSecretName -}}
+- name: cluster-ca
+  secret:
+    secretName: {{ .Values.clusterCaSecretName }}
+{{- end }}
+{{- if and ( .Capabilities.APIVersions.Has "cert-manager.io/v1" ) .Values.certificates.enabled .Values.certificates.proxy.enabled }}
+- name: {{ include "kong.fullname" . }}-proxy-cert
+  secret:
+    secretName: {{ include "kong.fullname" . }}-proxy-cert
+{{- end }}
```
rainest: This needs to be copy-pastaed for each of the remaining certificate types (admin, portal, etc.).

charts/kong/templates/_helpers.tpl
```diff
@@ -478,6 +504,14 @@ The name of the service used for the ingress controller's validation webhook
...
 - name: {{ template "kong.fullname" . }}-tmp
   mountPath: /tmp
+{{- if .Values.clusterCaSecretName -}}
+- name: cluster-ca
+  mountPath: /etc/cluster/
+{{- end }}
+{{- if and ( .Capabilities.APIVersions.Has "cert-manager.io/v1" ) .Values.certificates.enabled .Values.certificates.proxy.enabled }}
+- name: {{ include "kong.fullname" . }}-proxy-cert
+  mountPath: /etc/cert-manager/proxy/
+{{- end }}
```
rainest: This needs to be copy-pastaed for each of the remaining certificate types (admin, portal, etc.).

charts/kong/templates/_helpers.tpl
```diff
@@ -670,6 +704,15 @@ the template that it itself is using form the above sections.
...
 {{- end -}}
 
+{{- if .Values.clusterCaSecretName -}}
+  {{- $_ := set $autoEnv "KONG_CLUSTER_CA_CERT" "/etc/cluster/ca.crt" -}}
+{{- end -}}
+
+{{- if and ( .Capabilities.APIVersions.Has "cert-manager.io/v1" ) .Values.certificates.enabled .Values.certificates.proxy.enabled }}
+  {{- $_ := set $autoEnv "KONG_SSL_CERT" "/etc/cert-manager/proxy/tls.crt" -}}
+  {{- $_ := set $autoEnv "KONG_SSL_CERT_KEY" "/etc/cert-manager/proxy/tls.key" -}}
+{{- end -}}
```
rainest: This needs to be copy-pastaed for each of the remaining certificate types (admin, portal, etc.).

charts/kong/README.md
```diff
@@ -425,6 +429,29 @@ documentation on Service
...
+deploying Kong.
+
+Kong Gateway's supported cert-manager integration includes:
+  - Kong Hybrid mode mtls certificates
+  - Kong Gateway default proxy certificates
+  - Kong Admin API & Manager certificates
+  - Kong Developer API & Portal certificates
+
+Currently cert-manager support on hybrid mode clusters is only valid
+on shared certificate type hybrid mode configuration.
```
rainest: Minor suggestion just place the links inline--the shorthand names (or whatever Markdown calls the `[name]:URL` syntax) don't render on their own.

More broadly, this should indicate how to use these features. We don't want to rely on an example set of values alone with the feature already enabled; users need to understand where they're configured and the baseline of what they need to customize. Something like:

---

For example, if you'd like to provision an admin API certificate using the self-signed issuer and your admin API hostname is `kong-admin.example.com`, you can create a certificate for it by including the following in values.yaml:

```
certificates:
  enabled: true
  admin:
    commonName: "kong-admin.example.com"
```
```
admin:
  annotations:
    konghq.com/protocol: https
  enabled: true
  ingress:
    annotations:
      konghq.com/https-redirect-status-code: "301"
      konghq.com/protocols: https
      konghq.com/strip-path: "true"
    enabled: true
    ingressClassName: kong
    hostname: kong-admin.example.com
    path: /api
```

```suggestion
Kong services and proxied services have various certificate requirements.
By default, this chart installs Kong without [cert-manager](https://cert-manager.io/docs/) integration and
dataplanes generate a fallback self-signed certificate on each container if
no other source for the default proxy certificate is explicitly provided.

This cert-manager support requires that cert-manager be installed and
a public or [self-signed](https://cert-manager.io/docs/configuration/selfsigned/) valid chain-of-trust issuer be ready before
deploying Kong.

Kong Gateway's supported cert-manager integration includes:
  - Kong Hybrid mode mtls certificates
  - Kong Gateway default proxy certificates
  - Kong Admin API & Manager certificates
  - Kong Developer API & Portal certificates

Currently cert-manager support on hybrid mode clusters is only valid
on shared certificate type hybrid mode configuration.
```

charts/kong/templates/_helpers.tpl
```diff
@@ -21,6 +25,18 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
...
+
+{{- define "kong.certificate.issuer.ca" -}}
+{{- printf "%s-%s-%s" .Release.Name .Chart.Name "selfsigned-issuer-ca" -}}
+{{- end -}}
+
+{{- define "kong.certificate.issuer" -}}
+{{- printf "%s-%s-%s" .Release.Name .Chart.Name "selfsigned-issuer" -}}
+{{- end -}}
+
```
rainest: The first two are no longer used outside their definition, so they can be removed entirely. Probably something similar will be re-added for further hybrid automation.

`kong.certificate.issuer` is still in use, but since we no longer need it in multiple places (for some definitions of multiple), we can inline it within the template(s) that use it--more in the Certificate template comments.

```suggestion
```

charts/kong/templates/certificate-kong-admin.yaml
```diff
@@ -0,0 +1,25 @@
...
+  - {{ .Values.certificates.admin.commonName }}
+  {{- range .Values.certificates.admin.dnsNames }}
+  - {{ . | quote }}
+  {{- end }}
+  renewBefore: 360h
+  duration: 2160h
+  isCA: false
+  issuerRef:
+    name: {{ template "kong.certificate.issuer" . }}
+    kind: Issuer
```
rainest: We'll want to rework this to use the user-provided issuer and to support either Issuers or ClusterIssuers. To that end, we probably want to add an additional `.certificates.SERVICE.clusterIssuer` field and top-level `.certificates.issuer` and `.certificates.clusterIssuer` fields (assuming many users probably don't need to configure a different one per-Service).

The template can then generate the complete `issuerRef`, with logic like:
```
if .certificates.SERVICE.clusterIssuer:
issuerRef:
   name: .certificates.SERVICE.clusterIssuer
   kind: ClusterIssuer
else if .certificates.SERVICE.issuer:
<similar>
else if .certificates.clusterIssuer:
<similar>
else if .certificates.issuer:
<similar>
else
<use the self-signed clusterissuer from https://cert-manager.io/docs/configuration/selfsigned/ as a final fallback>
```
I.e. prefer the specific issuer if set, else the global. Arbitrarily prefer clusterissuers because you shouldn't set both, but we'll try to generate valid YAML even if we get conflicting instructions (Helm doesn't provide a great way to reject logically wrong configuration).

You can optionally reduce these into a single generic template that's used for each Certificate. Similar to the [Ingress generator](https://github.com/Kong/charts/blob/dedd27073980e866bd97e2187600f1a60f9af453/charts/kong/templates/_helpers.tpl#L62-L111), you can define a `kong.certmanagercertificate` and pass it a non-root context, and then just use `{{ .commonName }}` instead of the full paths. In an ideal world you'd just pass `.Values.certificates.admin` as the context, but in practice, since we want to use fullname and such, you'll probably need to [build a context object while you're still at the root](https://github.com/Kong/charts/blob/dedd27073980e866bd97e2187600f1a60f9af453/charts/kong/templates/service-kong-portal.yaml) (Helm supposedly provides a way to get to the root from a non-root context, but I've never been able to get it to work). I'd recommend condensing to a single template now, but it's easy enough to do after, so if you'd prefer not to just create an issue to do it later.

rainest (changes requested):
We want to handle Ingress certificates also without relying on magic values from an example. Enabling certificates should inject that TLS configuration somehow. Adding an additional `else if .Values.certificates.enabled .Values.certificates.admin.enabled` clause to the [ingress generator TLS block](https://github.com/Kong/charts/blob/dedd27073980e866bd97e2187600f1a60f9af453/charts/kong/templates/_helpers.tpl#L105-L110) that sets the secret to `{{ .fullName }}-{{ .serviceName }}-cert` should work. You can optionally add the `cert-manager.io/cluster-issuer` annotation but it appears pointless to do so if we've created the Certificate already (it does not appear to cause any conflicts with an existing Certificate).

If we're deferring the hybrid mode automation and/or instructions for now, is there reason to include any of the cluster certificate stuff in this iteration? IMO including it and exposing values settings for it implies there's something available that isn't really, and we should avoid deciding API surface (what values.yaml looks like) before the implementation is ready--we may find that our initial expectations were off and will need to make breaking changes if we do.

shaneutt:
Just wanted to check in on this one, is there any help needed here to move this one forward?

usrbinkat:
I believe I need to hack on the helm chart some more to incorporate feedback from rainest's last review when I can.

charts/kong/templates/_helpers.tpl
usrbinkat: added in [dbcdca5](https://github.com/Kong/charts/pull/592/commits/dbcdca5b3930aa555b4a3e4e694b162090116dea):
- proxy
- admin
- portal

charts/kong/templates/_helpers.tpl
usrbinkat: added in [dbcdca5](https://github.com/Kong/charts/pull/592/commits/dbcdca5b3930aa555b4a3e4e694b162090116dea):
- proxy
- admin
- portal

charts/kong/templates/_helpers.tpl
usrbinkat: added in [dbcdca5](https://github.com/Kong/charts/pull/592/commits/dbcdca5b3930aa555b4a3e4e694b162090116dea):
- proxy
- admin
- portal

charts/kong/templates/_helpers.tpl
usrbinkat: resolved in. 97764e4

rdotterer09:
what I don't understand is why this was set up as an EVM contract when I specifically asked for this contract to be updated so I could run it on the circleci pipeline and have control of the contract and funds within it???? can someone answer this for me, I am the contract owner and was the one who created the initial contract in Ethereum. This conversion has cost a lot of money and I'm still not getting the results I asked for rdotterer09@gmail.com

charts/kong/templates/certificate-kong-admin.yaml
usrbinkat: Added support for user defined top level, and service specific `issuer` and `clusterIssuer` [422b493](https://github.com/Kong/charts/pull/592/commits/422b4939f7eee498e86fbc37790b37cd84b9abbf).

Opened #654 to track template consolidation.

charts/kong/README.md
```diff
@@ -442,6 +446,26 @@ documentation on Service
...
+### Cert Manager Integration
+
+Kong services and proxied services have various certificate requirements.
+By default, this chart installs Kong without [cert-manager](https://cert-manager.io/docs/) integration and
+dataplanes generate a fallback self-signed certificate on each container if
+no other source for the default proxy certificate is explicitly provided.
+
+This cert-manager support requires that cert-manager be installed and
+a public or [self-signed](https://cert-manager.io/docs/configuration/selfsigned/) valid chain-of-trust issuer be ready before
+deploying Kong.
```
shaneutt: If I'm not mistaken we support using specific issuers, but we'll fall back to a _default issuer_ if one is available. If that's the case is that worth pointing out here to clarify?

charts/kong/example-values/quickstart-enterprise-licensed-aio.yaml
```diff
@@ -0,0 +1,291 @@
...
+    containerPort: 8100
+    enabled: true
+  tls:
+    containerPort: 8543
+    enabled: false
+updateStrategy:
+  rollingUpdate:
+    maxSurge: 100%
+    maxUnavailable: 100%
+  type: RollingUpdate
```
shaneutt: I would suggest since we're here we add some level of testing for this example if it's not too much trouble. Ideally we would run this new example through our integration test suite, and verify the deployment at least at a basic level to help avoid future regressions.

shaneutt (commented):
Couple of smaller comments only, mostly LGTM. I'll defer to Travis for full review however.

charts/kong/templates/certificate-kong-admin.yaml
usrbinkat: > ```
> else
> <use the self-signed clusterissuer from https://cert-manager.io/docs/configuration/selfsigned/ as a final fallback>
> ```

I'm unsure it makes sense to have an `{{ else }}` fallback in the template if we do not also have a template to generate the namespaced selfsigned issuer with the release.

I would like to add a selfsigned issuer template that can deploy with the release as a fallback to support this `else` behavior.

I acknowledge that this issuer fallback would float up to a meta chart and be removed from this core kong chart once a meta chart is released.

charts/kong/README.md
usrbinkat: in-discussion on comment https://github.com/Kong/charts/pull/592#discussion_r938189394

charts/kong/templates/certificate-kong-admin.yaml
rainest: Added https://github.com/usrbinkat/kong-charts-feat-cert-manager/pull/1 for consolidation. Manual smoke test log at [test.txt](https://github.com/Kong/charts/files/9285431/test.txt). After that and a merge of upstream main/fixing conflicts, this should be good to go.

We can probably take a doc/umbrella approach instead of a fallback and just remove the `else`:

- In the umbrella chart, it includes an issuer and sets `kong.certificates.clusterIssuer: <issuer name>` in the default values.yaml.
- If installing alone, the readme provides concise instructions. Two settings where we can provide explicit values are not an undue burden.

The error you hit if you install with no fallback and no configuration isn't perfect, but is at least in the right ballpark:

```
Error: INSTALLATION FAILED: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(Certificate.spec): missing required field "issuerRef" in io.cert-manager.v1.Certificate.spec
```

rainest (changes requested):
Aside from the changes mentioned in the open thread/my PR, I'd like to move `quickstart-enterprise-licensed-aio.yaml` to either the docs site itself (ostensibly, it should be fine to place files under the doc directory and add a relative link to create a download) or a new `example-values/doc-examples` directory with a readme that indicates that those files are part of examples on docs.konghq.com. Since it fills out everything with assumptions about the environment, it doesn't really fit with the others, where the intent is to fill out as little as possible to switch from the default DB-less+controller to hybrid, Enterprise, no controller, etc.

Guaris:
I prefer `example-values/doc-examples` directory.

usrbinkat:
Unsure how to root-cause current chart-testing failure.

<details><summary>Expandable</summary>
Local error log:

```
Installing chart with values file 'charts/kong/ci/test1-values.yaml'...                                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                                                                                       
Creating namespace 'kong-xdujy76mdp'...                                                                                                                                                                                                                                                                                                                                                
namespace/kong-xdujy76mdp created                                                                                                                                                                                                                                                                                                                                                      
W0812 13:51:44.634993   79697 warnings.go:70] policy/v1beta1 PodSecurityPolicy is deprecated in v1.21+, unavailable in v1.25+                                                                                                                                                                                                                                                          
W0812 13:51:44.670670   79697 warnings.go:70] autoscaling/v2beta2 HorizontalPodAutoscaler is deprecated in v1.23+, unavailable in v1.26+; use autoscaling/v2 HorizontalPodAutoscaler                                                                                                                                                                                                   
W0812 13:51:44.722089   79697 warnings.go:70] policy/v1beta1 PodSecurityPolicy is deprecated in v1.21+, unavailable in v1.25+                                                                                                                                                                                                                                                          
W0812 13:51:44.753191   79697 warnings.go:70] autoscaling/v2beta2 HorizontalPodAutoscaler is deprecated in v1.23+, unavailable in v1.26+; use autoscaling/v2 HorizontalPodAutoscaler                                                                                                                                                                                                   
Error: INSTALLATION FAILED: Deployment.apps "kong-xdujy76mdp-kong" is invalid: spec.template.spec.containers[0].volumeMounts[0].name: Not found: "webhook-cert"                                                                                                                                                                                                                        
==================================================================
===> Events of namespace kong-xdujy76mdp                                                                                                                                                                                                                                                                                                                                                
No resources found in kong-xdujy76mdp namespace.                                                                                                                                                                                                                                                                                                                                       
<== Events of namespace kong-xdujy76mdp                                                                                                                                                                                                                                                                                                                                                
==================================================================
Deleting release 'kong-xdujy76mdp'...                                                                                                                                                                                                                                                                                                                                                  
release "kong-xdujy76mdp" uninstalled                                                                                                                                                                                                                                                                                                                                                  
W0812 13:51:46.781874   79703 warnings.go:70] autoscaling/v2beta2 HorizontalPodAutoscaler is deprecated in v1.23+, unavailable in v1.26+; use autoscaling/v2 HorizontalPodAutoscaler                                                                                                                                                                                                   
W0812 13:51:46.801489   79703 warnings.go:70] policy/v1beta1 PodSecurityPolicy is deprecated in v1.21+, unavailable in v1.25+                                                                                                                                                                                                                                                          
Deleting namespace 'kong-xdujy76mdp'...                                                                                                                                                                                                                                                                                                                                                
namespace "kong-xdujy76mdp" deleted                                                                                                                                                                                                                                                                                                                                                    
Namespace 'kong-xdujy76mdp' terminated.                                                      
------------------------------------------------------------------ 
✖︎ kong => (version: "2.12.0", path: "charts/kong") > Error waiting for process: exit status 1                                                                                             
------------------------------------------------------------------
Error: Error installing charts: Error processing charts                                      
Error installing charts: Error processing charts
```


Runner error log:

```
-----------------------------------------------------------------------------------------------
==> Logs of container kong-zu33j3zzoi-kong-7c499fd588-8nptd
-----------------------------------------------------------------------------------------------
2022/08/12 20:39:12 [warn] 1#0: the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /kong_prefix/nginx.conf:6
nginx: [warn] the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /kong_prefix/nginx.conf:6
Notice: 12 20:39:12 [notice] 1#0: using the "epoll" event method
Notice: 12 20:39:12 [notice] 1#0: openresty/1.19.9.1
Notice: 12 20:39:12 [notice] 1#0: built by gcc 6.4.0 (Alpine 6.4.0) 
Notice: 12 20:39:12 [notice] 1#0: OS: Linux 5.15.0-1014-azure
Notice: 12 20:39:12 [notice] 1#0: getrlimit(RLIMIT_NOFILE): 1048576:1048576
Notice: 12 20:39:12 [notice] 1#0: start worker processes
Notice: 12 20:39:12 [notice] 1#0: start worker process 1115
Notice: 12 20:39:12 [notice] 1#0: start worker process 1116
Notice: 12 20:39:12 [notice] 1115#0: *1 [lua] init.lua:260: purge(): [DB cache] purging (local) cache, context: init_worker_by_lua*
Notice: 12 20:39:12 [notice] 1115#0: *1 [lua] init.lua:260: purge(): [DB cache] purging (local) cache, context: init_worker_by_lua*
Notice: 12 20:39:12 [notice] 1116#0: *2 [kong] init.lua:312 only worker #0 can manage, context: init_worker_by_lua*
127.0.0.1 - - [12/Aug/2022:20:39:13 +0000] "GET / HTTP/2.0" 200 10598 "-" "Go-http-client/2.0"
127.0.0.1 - - [12/Aug/2022:20:39:13 +0000] "GET /tags HTTP/2.0" 200 23 "-" "Go-http-client/2.0"
127.0.0.1 - - [12/Aug/2022:20:39:13 +0000] "GET / HTTP/2.0" 200 10598 "-" "Go-http-client/2.0"
Notice: 12 20:39:21 [notice] 1115#0: *5 [lua] init.lua:260: purge(): [DB cache] purging (local) cache, client: 127.0.0.1, server: kong_admin, request: "POST /config?check_hash=1 HTTP/2.0", host: "localhost:8444"
Notice: 12 20:39:21 [notice] 1115#0: *5 [lua] init.lua:260: purge(): [DB cache] purging (local) cache, client: 127.0.0.1, server: kong_admin, request: "POST /config?check_hash=1 HTTP/2.0", host: "localhost:8444"
127.0.0.1 - - [12/Aug/2022:20:39:21 +0000] "POST /config?check_hash=1 HTTP/2.0" 201 2 "-" "Go-http-client/2.0"
Error: /12 20:39:24 [error] 1115#0: *5 [lua] api_helpers.lua:511: handle_error(): /usr/local/share/lua/5.1/lapis/application.lua:424: /usr/local/share/lua/5.1/kong/api/routes/health.lua:45: http2 requests not supported yet
stack traceback:
	[C]: in function 'capture'
	/usr/local/share/lua/5.1/kong/api/routes/health.lua:45: in function 'fn'
	/usr/local/share/lua/5.1/kong/api/api_helpers.lua:285: in function </usr/local/share/lua/5.1/kong/api/api_helpers.lua:[268](https://github.com/Kong/charts/runs/7813398846?check_suite_focus=true#step:10:269)>

stack traceback:
	[C]: in function 'error'
	/usr/local/share/lua/5.1/lapis/application.lua:424: in function 'handler'
	/usr/local/share/lua/5.1/lapis/application.lua:146: in function 'resolve'
	/usr/local/share/lua/5.1/lapis/application.lua:183: in function </usr/local/share/lua/5.1/lapis/application.lua:181>
	[C]: in function 'xpcall'
	/usr/local/share/lua/5.1/lapis/application.lua:189: in function 'dispatch'
	/usr/local/share/lua/5.1/lapis/nginx.lua:231: in function 'serve'
	/usr/local/share/lua/5.1/kong/init.lua:1423: in function 'admin_content'
	content_by_lua(nginx-kong.conf:358):2: in main chunk, client: 127.0.0.1, server: kong_admin, request: "GET /status HTTP/2.0", host: "localhost:8444"
127.0.0.1 - - [12/Aug/2022:20:39:24 +0000] "GET /status HTTP/2.0" 500 42 "-" "Go-http-client/2.0"
Error: /12 20:39:27 [error] 1115#0: *5 [lua] api_helpers.lua:511: handle_error(): /usr/local/share/lua/5.1/lapis/application.lua:424: /usr/local/share/lua/5.1/kong/api/routes/health.lua:45: http2 requests not supported yet
stack traceback:
	[C]: in function 'capture'
	/usr/local/share/lua/5.1/kong/api/routes/health.lua:45: in function 'fn'
	/usr/local/share/lua/5.1/kong/api/api_helpers.lua:285: in function 'fn'
	/usr/local/share/lua/5.1/kong/api/api_helpers.lua:285: in function </usr/local/share/lua/5.1/kong/api/api_helpers.lua:268>

stack traceback:
	[C]: in function 'error'
	/usr/local/share/lua/5.1/lapis/application.lua:424: in function 'handler'
	/usr/local/share/lua/5.1/lapis/application.lua:146: in function 'resolve'
	/usr/local/share/lua/5.1/lapis/application.lua:183: in function </usr/local/share/lua/5.1/lapis/application.lua:181>
	[C]: in function 'xpcall'
	/usr/local/share/lua/5.1/lapis/application.lua:189: in function 'dispatch'
	/usr/local/share/lua/5.1/lapis/nginx.lua:231: in function 'serve'
	/usr/local/share/lua/5.1/kong/init.lua:1423: in function 'admin_content'
	content_by_lua(nginx-kong.conf:358):2: in main chunk, client: 127.0.0.1, server: kong_admin, request: "GET /status HTTP/2.0", host: "localhost:8444"
127.0.0.1 - - [12/Aug/2022:20:39:27 +0000] "GET /status HTTP/2.0" 500 42 "-" "Go-http-client/2.0"
-----------------------------------------------------------------------------------------------
<== Logs of container kong-zu33j3zzoi-kong-7c499fd588-8nptd
-----------------------------------------------------------------------------------------------
===============================================================================================
Deleting release 'kong-zu33j3zzoi'...
release "kong-zu33j3zzoi" uninstalled
Deleting namespace 'kong-zu33j3zzoi'...
namespace "kong-zu33j3zzoi" deleted
Namespace 'kong-zu33j3zzoi' terminated.

Installing chart with values file 'charts/kong/ci/test1-values.yaml'...

Creating namespace 'kong-35mktr9ywv'...
namespace/kong-35mktr9ywv created
W0812 20:39:56.068551    9441 warnings.go:70] policy/v1beta1 PodSecurityPolicy is deprecated in v1.21+, unavailable in v1.25+
W0812 20:39:56.090571    9441 warnings.go:70] autoscaling/v2beta2 HorizontalPodAutoscaler is deprecated in v1.23+, unavailable in v1.26+; use autoscaling/v2 HorizontalPodAutoscaler
W0812 20:39:56.168[281](https://github.com/Kong/charts/runs/7813398846?check_suite_focus=true#step:10:282)    9441 warnings.go:70] policy/v1beta1 PodSecurityPolicy is deprecated in v1.21+, unavailable in v1.25+
W0812 20:39:56.247131    9441 warnings.go:70] autoscaling/v2beta2 HorizontalPodAutoscaler is deprecated in v1.23+, unavailable in v1.26+; use autoscaling/v2 HorizontalPodAutoscaler
Error: INSTALLATION FAILED: Deployment.apps "kong-35mktr9ywv-kong" is invalid: spec.template.spec.containers[0].volumeMounts[0].name: Not found: "webhook-cert"
===============================================================================================
...............................................................................................
==> Events of namespace kong-35mktr9ywv
...............................................................................................
...............................................................................................
<== Events of namespace kong-35mktr9ywv
...............................................................................................
===============================================================================================
Deleting release 'kong-35mktr9ywv'...
release "kong-35mktr9ywv" uninstalled
W0812 20:39:56.789744    9507 warnings.go:70] autoscaling/v2beta2 HorizontalPodAutoscaler is deprecated in v1.23+, unavailable in v1.26+; use autoscaling/v2 HorizontalPodAutoscaler
W0812 20:39:56.817486    9507 warnings.go:70] policy/v1beta1 PodSecurityPolicy is deprecated in v1.21+, unavailable in v1.25+
Deleting namespace 'kong-35mktr9ywv'...
namespace "kong-35mktr9ywv" deleted
Namespace 'kong-35mktr9ywv' terminated.
-----------------------------------------------------------------------------------------------
 ✖︎ kong => (version: "2.12.0", path: "charts/kong") > Error waiting for process: exit status 1
-----------------------------------------------------------------------------------------------
Error installing charts: Error processing charts
Error: Error installing charts: Error processing charts
Error: Process completed with exit code 1.
```

</details>

rainest:
How did you handle the rebase? There are actual commits for it, which is unusual (I guess it lets you commit changes separately instead of adding a fixed file and continuing?) and there are some additions that don't look like they should be there, e.g. https://github.com/Kong/charts/pull/592/files?show-viewed-files=true&file-filters%5B%5D=#diff-3e8a71242505abb72903a338eef1cdea2cd2c674f877c59378eff807951a1fa8R677-R680 was originally https://github.com/Kong/charts/blob/main/charts/kong/templates/_helpers.tpl#L631-L635

Ultimately https://github.com/Kong/charts/blob/70a9aa675487b93c5cfb7b772abe65196aa53f10/charts/kong/templates/_helpers.tpl#L443 (combined with removing the the `end` in https://github.com/Kong/charts/pull/592/commits/f133cf7586d6a750b3bf8b3460cb5eab571b00c3) is terminating the volumes template early, but that rebase may have done something else unwanted, so you'd want to [use reflog](https://www.atlassian.com/git/tutorials/rewriting-history/git-reflog) find the `HEAD@{x}` before the rebase began, and `git reset --hard HEAD@{x}` to restart it.

The fix for that conflict should be to move the origin/main emptydir definition under the `-tmp` mount and strip the leftover empty one. The actual edits look like:

```diff
11:46:04-0700 esenin $ diff -u /tmp/diff_helpers.tpl templates/_helpers.tpl
--- /tmp/diff_helpers.tpl	2022-08-15 11:47:21.112759971 -0700
+++ templates/_helpers.tpl	2022-08-15 11:47:37.469479044 -0700
@@ -416,7 +416,8 @@
   emptyDir: 
     sizeLimit: {{ .Values.deployment.prefixDir.sizeLimit }}
 - name: {{ template "kong.fullname" . }}-tmp
-<<<<<<< HEAD
+  emptyDir: 
+    sizeLimit: {{ .Values.deployment.tmpDir.sizeLimit }}
-  emptyDir: {}
 {{- if and ( .Capabilities.APIVersions.Has "cert-manager.io/v1" ) .Values.certificates.enabled .Values.certificates.cluster.enabled }}
 - name: {{ include "kong.fullname" . }}-cluster-cert
@@ -438,10 +439,6 @@
   secret:
     secretName: {{ include "kong.fullname" . }}-portal-cert
 {{- end }}
-=======
-  emptyDir: 
-    sizeLimit: {{ .Values.deployment.tmpDir.sizeLimit }}
->>>>>>> origin/main
 {{- if (and (.Values.postgresql.enabled) .Values.waitImage.enabled) }}
 - name: {{ template "kong.fullname" . }}-bash-wait-for-postgres
   configMap:
```
And the final result is [_helpers.tpl.txt](https://github.com/Kong/charts/files/9339235/_helpers.tpl.txt)

charts/kong/example-values/doc-examples/README.md
```diff
@@ -0,0 +1,5 @@
+# Kong Gateway Helm Docs Samples
+This directory contains sample values files written in support of official [docs.konghq.com](https://docs.konghq.com/gateway/2.8.x/install-and-run/) workflows.
+
+###### Disclaimer:
+> While these sample values are written by Kong to demonstrate many features of Kong Gateway, the samples here are published for educational purposes and should not be considered production ready as is.[](https://www.youtube.com/watch?v%3DwzJdj1YoCSo)
```
rainest:
```suggestion
## Disclaimer
The samples here are published for educational purposes and should not be considered production ready as-is.
```

rainest (changes requested):

rainest (approved):

PR #599, opened 2022-05-28
Disable jobs flag

### LINKED ISSUE #598: Add disabled flag for test-jobs.yaml
Please add a values disabled flag option for [test-jobs.yaml](https://github.com/Kong/charts/blob/main/charts/kong/templates/tests/test-jobs.yaml), like you have for test-resources.yaml.

The service url being curled is hard coded to `-kong-proxy....`. As a result, my tests have restarted over 1000 times since that doesn't match the name of my service.

arm749:
This adds a conditional option to disable test jobs from running. The test jobs are currently formatted to look at the release name `{{ .Release.Name }}-kong-proxy.{{ .Release.Namespace }}.svc.cluster.local/httpbin` with kong-proxy being hard coded. There are cases where the service name does not match that url resulting in unlimited restarts on the test pods. For example, if the Helm `nameOverride` value is used or if the service is created by something outside of the chart (Terraform).
  - fixes #598 

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## Unreleased
 
-Nothing yet.
+* Added option to disable test job pods.
+  [#598](https://github.com/Kong/charts/issues/598)
 
 ## 2.9.0
 
--- a/charts/kong/templates/tests/test-jobs.yaml
+++ b/charts/kong/templates/tests/test-jobs.yaml
@@ -1,3 +1,4 @@
+{{- if  .Values.deployment.test.enabled }}
 ---
 apiVersion: v1
 kind: Pod
@@ -43,3 +44,4 @@ spec:
       command:
         - curl
         - "http://{{ .Release.Name }}-kong-proxy.{{ .Release.Namespace }}.svc.cluster.local/httpbin-httproute"
+{{- end }}
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -726,6 +726,7 @@ kong:
 | deployment.serviceAccount.create   | Create Service Account for the Deployment / Daemonset and the migrations              | `true`              |
 | deployment.serviceAccount.name     | Name of the Service Account, a default one will be generated if left blank.           | ""                  |
 | deployment.serviceAccount.annotations | Annotations for the Service Account                                                | {}                  |
+| deployment.test.enabled            | Enable creation of test resources for use with "helm test"                            | `false`             |
 | autoscaling.enabled                | Set this to `true` to enable autoscaling                                              | `false`             |
 | autoscaling.minReplicas            | Set minimum number of replicas                                                        | `2`                 |
 | autoscaling.maxReplicas            | Set maximum number of replicas                                                        | `5`                 |


charts/kong/CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-Nothing yet.
+* Added option to disable test job pods.
```
shaneutt, line 5:
```suggestion
* Added option to disable test job pods.
  [#598](https://github.com/Kong/charts/issues/598)
```

charts/kong/templates/tests/test-jobs.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if  .Values.deployment.testJobs.enabled }}
```
shaneutt: It seems like rather than adding a new boolean, we could just leave the existing `test.enabled` boolean and check it conditionally here rather than having a separate one. I'm not sure I'm seeing a situation where you would want one, but not the other? :thinking: Let me know your thoughts

shaneutt (dismissed):

charts/kong/templates/tests/test-jobs.yaml
arm749: @shaneutt That's how I initially had it, but I went with separating them because I noticed ```.Values.deployment.tests.enabled``` defaulted to ```false``` and the test jobs were enabled by default. I'm with you though, I think they should use the same condition. The question though is what should they default to?

shaneutt: I would say just apply the existing value as a boolean check on the Jobs, and keep the default to false. You only want the jobs to test the test resources, and testing resources and jobs should be an opt-in because 99.9% of deployments don't want to run those tests (they're mostly used for CI purposes, where we run the tests for PRs).

shaneutt:
Looks like a rebase against main is needed, the diff is showing removal of recent changes.

rainest (changes requested):
Generally I'd like to avoid adding further `helm template`-specific fields if possible. If there's behavior from `helm install` (e.g. not rendering test templates unless you're running `helm test`) that `template` cannot mimic and needs to be able to mimic, that should be handled upstream in Helm itself. This one already exists, however:

```
08:52:51-0700 esenin $ helm template fexample kong/kong | grep -c httpbin-v1beta1             
1
08:53:01-0700 esenin $ helm template --skip-tests fexample kong/kong | grep -c httpbin-v1beta1 
0
```

@arm749 does using https://registry.terraform.io/providers/hashicorp/helm/latest/docs/data-sources/template#skip_tests result in what you want?

arm749:
> Generally I'd like to avoid adding further `helm template`-specific fields if possible. If there's behavior from `helm install` (e.g. not rendering test templates unless you're running `helm test`) that `template` cannot mimic and needs to be able to mimic, that should be handled upstream in Helm itself. This one already exists, however:
> 
> ```
> 08:52:51-0700 esenin $ helm template fexample kong/kong | grep -c httpbin-v1beta1             
> 1
> 08:53:01-0700 esenin $ helm template --skip-tests fexample kong/kong | grep -c httpbin-v1beta1 
> 0
> ```
> 
> @arm749 does using https://registry.terraform.io/providers/hashicorp/helm/latest/docs/data-sources/template#skip_tests result in what you want?

@rainest The issue is we are using Jenkins X for our CI. Jenkins X has their own cli command that wraps around helm template, so we don't easily have a way to add the --skip-tests flag. As for Terraform, we have a case where someone created a service using the Kubernetes provider, and then disables the service on the Helm chart... ends up linking the two together. Obviously that is not ideal, and is easier to change if we must.

rainest:
Can you elaborate on what you mean regarding the Service? That sounds unrelated to the original issue.

If Jenkins lacks support for that feature (which does appear to be the case looking at https://github.com/jenkins-x-plugins/jx-gitops/blob/6f6cfceade01384dde7f5dd827bca0a91465f5bd/pkg/cmd/helm/step_helm_template.go#L178-L207) then it's something they should add/you should submit a PR there to add it.

arm749:
@rainest The service issue is to just show a use case where the service name may not always match the value of what is set on the tests: ```{{ .Release.Name }}-kong-proxy.{{ .Release.Namespace }}.svc.cluster.local/httpbin```. But the root stems from the tests being spun up due to not passing in ```--skip-tests```(Jenkins X problem).

I would like to mention, of all the Helm charts we are using, all provide us the flexibility to turn off the tests. Kong is the only problem we have with tests restarting and not being able to disable them. I.e [Grafana](https://artifacthub.io/packages/helm/grafana/grafana?modal=template&template=tests/test.yaml) & [Sonarqube](https://artifacthub.io/packages/helm/sonarqube/sonarqube?modal=template&template=tests/sonarqube-test.yaml) to name a few.

charts/kong/templates/tests/test-jobs.yaml
rainest: It should remain defaulted to `false`. We set it to `true` in the test configs that need it already. Currently these are effectively default `false` when using `helm install` because they have the test hook annotation, and Helm strips them out after templating.

I curse the ecosystem for not fully evaluating if the tools they were using were fit for purpose and do still think the burden of handling this should be upstream in the CD systems, but if we already have this flag, expanding its reach is fine.

arm749: Updated my pr so it's in sync with main. Appreciate you guys entertaining this change.

rainest (approved):

PR #601, opened 2022-06-01
feat(deployment) Add cmd / args to deployment.yaml

craig-rueda:
This PR adds two new fields to the `deployment.yaml` manifest (`command` and `args`) which optionally allows users of this chart to override the defaults provided by the docker image's `CMD` and `ENTRYPOINT` directives.

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Added `deployment.kong.args` and `deployment.kong.args` options for Proxy container
+  ([601](https://github.com/Kong/charts/pull/601))
 * Added terminationDelaySeconds for Ingress Controller.
   ([597](https://github.com/Kong/charts/pull/597))
 
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -717,6 +717,8 @@ kong:
 | Parameter                          | Description                                                                           | Default             |
 | ---------------------------------- | ------------------------------------------------------------------------------------- | ------------------- |
 | namespace                          | Namespace to deploy chart resources                                                   |                     |
+| deployment.kong.args               | Optionally override args field in proxy container                                     |                     |
+| deployment.kong.command            | Optionally override cmd field in proxy container                                      |                     |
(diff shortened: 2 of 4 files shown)


rainest:
What was the motivation behind this/what did you want to override with?

AFAIK we generally do want everything to run through the [standard entrypoint](https://github.com/Kong/docker-kong/blob/master/docker-entrypoint.sh), and I'd expect that most reasons to override it are either something we could incorporate into that (if it's suitable for all users) or something that'd make use of a custom image (in which the entrypoint should be changed during build).

craig-rueda:
This gives a little more flexibility. I would like to change up the CMD that's run and don't want to be coupled to the docker entrypoint that ships with the image. In the future, should that thing change, it will make this chart brittle. 

Also note that my change leaves the default behavior and only allows the CMD to be optionally overridden.

rainest:
Right, we're asking why you want to run a different command/what other command you'd want to run to evaluate whether this is indeed the best approach.

We want to avoid having values.yaml essentially expose all fields of the resources it generates unless there's a demonstrated need to do so.

craig-rueda:
Ok, understood. We have a slightly custom docker image built from the official one and we would like to pass custom args / cmd without running all of the stuff that the entrypoint does currently. :) 

Also - prob best for another Issue/PR, but have you thought about splitting the ingress controller container from the proxy/admin container? Typically one would only need 1-2 replicas of the controller whereas you might want to have several replicas (> 2) of the proxy.

rainest:
You should be able to override the existing script by adding your own `ENTRYPOINT /bin/sh` and `CMD whatever` to your Dockerfile, correct? The only thing that should require the standard script is if your `CMD` is `kong docker-start`, and we don't actually include that in the chart since it's the default.

The standard entrypoint script should actually be fine for most alternative commands regardless--its job is essentially to look at whether the command is `kong <something>` and perform some helper tasks we want for a Docker Kong environment, or just run the command as-is if it's something else.

https://github.com/Kong/charts/blob/main/charts/kong/README.md#standalone-controller-nodes is already available. We wouldn't be able to handle Deployments that include both containers (which you need for DB-less mode) if they were separate charts.

PR #603, opened 2022-06-02
fix lint ( yamllint ) reported issues

rkazak:
fix lint ( yamllinter ) reported issues.

  - fixes #

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+* Fixed lint reported issue.
 * Added terminationDelaySeconds for Ingress Controller.
   ([597](https://github.com/Kong/charts/pull/597))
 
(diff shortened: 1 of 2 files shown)


rainest:
The YAML specification leaves whether to include an actual whitespace indent in sequences or to count the `- ` as the indent up to personal preference. Which yamllint actually complains about depends on [how it is configured](https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.indentation), though AFAIK the default is what we're currently using (do include the additional whitespace), not what is proposed here (count the sequence notation as the indent).

As a general rule, I want to avoid whitespace-only PRs. Managing whitespace throughout Helm templates is difficult due to the nature of modular templated YAML, and ensuring clean consistent whitespace throughout is a time-consuming endeavor. Our goal is to render valid YAML containing valid Kubernetes resources, and as long as the templates do that I'm not concerned with inconsistent indents or extra newlines much. Ultimately Helm's output is temporary, since the API server will later return YAML representations according to its own rules.

PR #611, opened 2022-06-11
Support cluster-scoped resources when using watchNamespaces

### LINKED ISSUE #607: New installs of v2.8.2 missing RBAC permissions for IngressClass
When installing v2.8.2 with `ingressController.watchNamespaces` set to ["default"], the `ingress-controller` logs the following error and remains in cyclic restarts:

E0606 17:16:56.258537       1 reflector.go:138] pkg/mod/k8s.io/client-go@v0.23.5/tools/cache/reflector.go:167: Failed to watch *v1.IngressClass: failed to list *v1.IngressClass: ingressclasses.networking.k8s.io is forbidden: User "system:serviceaccount:default:kong-kong" cannot list resource "ingressclasses" in API group "networking.k8s.io" at the cluster scope

It seems that the ingress-controller requires access to the IngressClass API resources, which are cluster level resources, even if a specific namespace to watch is specified.

See issue [#2541 in kubernetes-ingress-controller.](https://github.com/Kong/kubernetes-ingress-controller/issues/2541)

The following patch to `kong/templates/controller-rbac-resources.yaml` fixes the problem:

```diff
89d88
< {{- if eq (len .Values.ingressController.watchNamespaces) 0 }}
97a97
> {{- if eq (len .Values.ingressController.watchNamespaces) 0 }}
98a99
> {{- end }}
115c116
< {{- else }}
---
> {{- if gt (len .Values.ingressController.watchNamespaces) 0 }}
```

rainest:
When watchNamespaces is non-empty, create a ClusterRole that contains
cluster-scoped resources only.

Remove endpoints from the cluster resource template. They appear to have
been included by mistake.

Configure one of the CI values to use watchNamespaces.

  - fixes #607

When watchNamespaces was originally developed, the only cluster-scoped resource we had was KongClusterPlugin, and KongPlugin could substitute for it with an acceptable loss of functionality. Since, CNCF has introduced IngressClass and GatewayClass, which have no namespace-scoped equivalent. We need to read these to properly implement the v1 Ingress and vAny Gateway specs.

Ready for review, but do not merge pending discussion with product and field team stakeholders in KUBE-61.

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -2,7 +2,10 @@
 
 ## Unreleased
 
-Nothing yet.
+### Improvements
+
+* Added ClusterRole for cluster-scoped resources when using watchNamespaces.
+  [#611](https://github.com/Kong/charts/issues/611)
 
 ## 2.11.0
 
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -532,27 +532,17 @@ resource.
 ### Removing cluster-scoped permissions
 
 You can limit the controller's access to allow it to only watch specific
-namespaces for resources. By default, the controller watches all namespaces.
-Limiting access requires several changes to configuration:
+namespaces for namespaced resources. By default, the controller watches all
+namespaces. Limiting access requires several changes to configuration:
 
 - Set `ingressController.watchNamespaces` to a list of namespaces you want to
   watch. The chart will automatically generate roles for each namespace and
   assign them to the controller's service account.
-- Set `ingressController.env.enable_controller_kongclusterplugin=false` and
-  `ingressController.env.enable_controller_ingress_class_networkingv1=false`.
-  These are cluster-scoped resources, and controllers with no ClusterRole
-  cannot access them.
 - Optionally set `ingressContrller.installCRDs=false` if your user role (the
   role you use when running `helm install`, not the controller service
   account's role) does not have access to get CRDs. By default, the chart
   attempts to look up the controller CRDs for [a legacy behavior
   check](#crd-management).
-
-Because there is no namespaced version of IngressClass, controllers without
-cluster-scoped permissions cannot access them. The controller will rely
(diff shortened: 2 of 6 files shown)


shaneutt (commented):
:+1:

shaneutt (dismissed):

jasiek-zywczak:
the behaviour of charts is (almost) correct, when we are running kong in the cluster we should not require cluster role, the problem is that chart is not disable watching the ingressClass in that case
https://github.com/Kong/charts/issues/648

rainest:
@jasiek-zywczak while the controller should indeed disable that lookup when disabling the controller, the plan is to continue with this change (we haven't received objections from the stakeholders we reached out to), which should resolve your issue, correct? The install would have permissions for IngressClass, and the attempts to list them would not fail. Did you have any concerns with this approach?

jasiek-zywczak:
It is a matter of being consequent. Why disabling access to cluster plugins then? I mean, in case watch namespace is set?
And btw chart's docs precisely mention that access to cluster objects should be disabled.
When I set that I want to watch a particular namespace it means I don't want to watch, in some cases, the whole cluster.

rainest:
> Why disabling access to cluster plugins then? I mean, in case watch namespace is set?

That is a good catch--this change should indeed remove the KongClusterPlugin disable as well. I've updated this to do so and document the new behavior.

> When I set that I want to watch a particular namespace it means I don't want to watch, in some cases, the whole cluster.

That will be the case for namespaced resources. The library we use to implement watchNamespaces has two modes: it can add cluster-level watches for everything or [create separate watches for each namespace and a cluster-level watch for cluster-scoped resources only](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/cache#MultiNamespacedCacheBuilder). It's not possible to watch cluster-scoped resources at anything other than the cluster level, so trying to impose a namespace-restricted model on them doesn't make sense.

shaneutt (approved):

PR #621, opened 2022-06-28
Fix token format and release 2.10.1

### LINKED ISSUE #620: Helm Chart Breaks Deployment in Kuma
With the latest version of Helm chart (2.10), deploying KIC when Kuma fails. It is related to this PR, #619 

We need a way to manually mount. The proposed solution from mesh team is to use a ContainerPatch. Use the following as a guide:

```
apiVersion: kuma.io/v1alpha1
kind: ContainerPatch
metadata:
  name: container-patch-kong-token
  namespace: kong-mesh-system
spec:
  sidecarPatch:
    - op: add
      path: /volumeMounts/-
      value: '{ "mountPath": "/var/run/secrets/kubernetes.io/serviceaccount/token", "name":"kong-1656434665-kong-token", "readOnly": "true", "subPath": "token" }'
    - op: add
      path: /volumeMounts/-
      value: '{ "mountPath": "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt", "name":"kong-1656434665-kong-token", "readOnly": "true", "subPath": "ca.crt" }'
    - op: add
      path: /volumeMounts/-
      value: '{ "mountPath": "/var/run/secrets/kubernetes.io/serviceaccount/namespace", "name":"podinfo", "readOnly": "true", "subPath": "namespace" }'
```

Could have a `kuma: enabled` flag or something to trigger the creation of ContainerPatch.

## Prerequisites

- [ ] https://github.com/Kong/kubernetes-testing-framework/issues/303

rainest:

Remove the separate SA token/namespace mounts and instead mount the `namespace` key in the SA token we mount for the controller container. This matches the format used in the automount mount. Kuma [has logic to find the token on a per-container basis without automount on the Pod](https://github.com/kumahq/kuma/blob/5e12d423e9a6f3f6f9b5eefaed4a8759395c14b7/pkg/plugins/runtime/k8s/webhooks/injector/injector.go#L390-L401), but it requires the mount use [its standard path](https://github.com/kumahq/kuma/blob/5e12d423e9a6f3f6f9b5eefaed4a8759395c14b7/pkg/plugins/runtime/k8s/webhooks/injector/injector.go#L32) to find it.

We didn't need the separate downward API mount, since the namespace key is already populated in the SAT.

This is easier than the ContainerPatch since we don't need a new template that requires CRD availability checks.

  - fixes #620 

Manual testing pending https://github.com/Kong/kubernetes-testing-framework/issues/303: [testing.txt](https://github.com/Kong/charts/files/9003924/testing.txt). I had some concern that whatever updates the Secret to stuff the token into it would overwrite the entire key set and remove the namespace key, but it is still present when checking after.

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -4,6 +4,13 @@
 
 Nothing yet.
 
+## 2.10.1
+
+### Fixed
+
+* Updated manual ServiceAccount Secret mount format for compatibility with
+  Kuma.
+
 ## 2.10.0
 
 ### Added
--- a/charts/kong/Chart.yaml
+++ b/charts/kong/Chart.yaml
@@ -10,7 +10,7 @@ maintainers:
   email: traines@konghq.com
 name: kong
 sources:
-version: 2.10.0
+version: 2.10.1
 appVersion: "2.8"
 dependencies:
 - name: postgresql
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -630,16 +630,7 @@ The name of the service used for the ingress controller's validation webhook
 {{- end }}
 {{- if or .Values.deployment.serviceAccount.create .Values.deployment.serviceAccount.name }}
   - name: {{ template "kong.serviceAccountTokenName" . }}
-    mountPath: /var/run/secrets/kubernetes.io/serviceaccount/token
-    subPath: token
-    readOnly: true
-  - name: {{ template "kong.serviceAccountTokenName" . }}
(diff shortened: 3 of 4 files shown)


rainest:
I should remember to save files before committing them, they work better then.

Checked to see if we could actually do the subPath mount of the downward API over the regular mount, but that breaks the container, so nope.

rainest:
Nevermind, it was there all along without us populating it, we just weren't mounting it, so you don't need to manipulate the Secret manually.

shaneutt (approved):

PR #625, opened 2022-07-01
Add Kuma to integration tests

rainest:
Adds Kuma to the integration tests

Kuma installs a webhook that apparently tries to piggyback off the upstream Gateway webhook for HTTPRoute and such. Not quite sure why (if present, it'll run on its own), but in any case I wasn't able to get it to run: it would complain about being unable to find resources or not trusting certificates without a lot of debug information. Since I don't care about it for the purpose of tests, I've changed its policy.

#### Checklist

### DIFF
--- a/scripts/test-run.sh
+++ b/scripts/test-run.sh
@@ -28,19 +28,50 @@ KUBECTL="kubectl --context kind-${TEST_ENV_NAME}"
 KUBERNETES_VERSION="$($KUBECTL version -o json | jq -r '.serverVersion.gitVersion')"
 
 # ------------------------------------------------------------------------------
+# Deploy Kuma configuration and test namespace
+# ------------------------------------------------------------------------------
+
+echo "---
+apiVersion: kuma.io/v1alpha1
+kind: Mesh
+metadata:
+  name: default
+spec:
+  mtls:
+    backends:
+    - conf:
+        caCert:
+          RSAbits: 2048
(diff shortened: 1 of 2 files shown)


scripts/test-run.sh
```diff
@@ -27,20 +27,51 @@ TEST_ENV_NAME="${TEST_ENV_NAME:-kong-charts-tests}"
...
+" | kubectl --context "kind-${TEST_ENV_NAME}" apply -f -
+
 # ------------------------------------------------------------------------------
 # Deploy Chart - Kubernetes Ingress Controller
 # ------------------------------------------------------------------------------
 
 if [[ "${TAG}" == "default" ]]
 then
     echo "INFO: installing chart as release ${RELEASE_NAME} to namespace ${RELEASE_NAMESPACE}"
-    helm install --create-namespace --namespace "${RELEASE_NAMESPACE}" "${RELEASE_NAME}" \
```
pmalek, line 37: Why remove this 🤔?

pmalek (approved):

scripts/test-run.sh
rainest: We're now creating the namespace along with the Mesh resource above. We can't set `kuma.io/sidecar-injection: enabled` on it with Helm and want that annotation in place before Kong's Deployment is created, since Kuma won't update existing resources without a restart.

PR #627, opened 2022-07-04
chore: bump controller version to 2.4

pmalek:

Bumps KIC version to "2.4"

#### Checklist

- [x] Changes are documented under the "2.11.0" header in CHANGELOG.md

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -1,14 +1,11 @@
 # Changelog
 
-## 2.10.3
-
-### Fixed
-
-* Fix URL for more detail on Command Line Arguements / environment variables for Helm.
+## Unreleased
 
 ### Improvements
 
-* Bump controller version to 2.4 [#627](https://github.com/Kong/charts/issues/627)
+* Bump controller version to 2.4.
+  [#627](https://github.com/Kong/charts/issues/627)
 
 ## 2.10.2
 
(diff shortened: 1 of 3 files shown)


jrsmroz (dismissed):

rainest (changes requested):
We'll actually want to use `"2.4"` and `2.10.3`. We don't consider default version bumps a minor version change since they should only affect new installs--properly-configured releases should pin their image versions.

The KIC repo manifests do use the full version because they're tagged along with that specific release--if you get the manifest for tag 2.4.1, it should indeed install 2.4.1 even if there's a newer patch available. The chart isn't so tightly coupled and we want the version that will get new patch versions as they roll out.

Edit: well, actually just strip the chart bump altogether. There are a few changes in flight and it'll be easier to release them all at once.

pmalek:
@rainest 👍 PTAL now

jrsmroz:
> We'll actually want to use `"2.4"` and `2.10.3`. We don't consider default version bumps a minor version change since they should only affect new installs--properly-configured releases should pin their image versions.
> 

Thanks for your clarification, I was not aware of this.

jrsmroz:
@pmalek Please change the PR title and description too.

pmalek:
@rainest BTW: what you're trying to say is that by bumping to 2.4 the user will ultimately get latest patch from 2.4.x?

rainest (approved):
@pmalek yes, we push a minor `x.y` tag every time we release, so it points to the latest patch for that minor.

Some added changes to back out the release and drop a changelog item we shouldn't have had (it was docs only).

PR #631, opened 2022-07-06
chore(ci) bump CI dependency versions

rainest:
Versions of CI dependencies had diverged across workflows, causing jobs run on main push to fail (it used an older Helm version than the PR job, and the older version lacks features KTF needs).

This bumps all the actions and such to their latest version across all jobs.


pmalek (dismissed):

rainest:
Curse YAML.

jrsmroz (approved):

PR #632, opened 2022-07-06
Size limits for Empty Dirs

chaitanya14:

Empty directory without size limit is consuming all the ephemeral local storage resources. Need to have a capability to assign limits on it. 

  - fixes #

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixed
+
+* Fix to avoid running out of ephmeral storage on kubernetes worker nodes.
+
 ### Improvements
 
 * Bump controller version to 2.4.
--- a/charts/kong/values.yaml
+++ b/charts/kong/values.yaml
@@ -58,7 +58,7 @@ deployment:
     sizeLimit: 256Mi
   # tmp empty dir size  
   tmpDir: 
-    sizeLimit: 256Mi
+    sizeLimit: 1Gi
 # Override namepsace for Kong chart resources. By default, the chart creates resources in the release namespace.
 # This may not be desirable when using this chart as a dependency.
 # namespace: "example"
--- a/charts/kong/values.yaml
+++ b/charts/kong/values.yaml
@@ -54,10 +54,10 @@ deployment:
   daemonset: false
   hostNetwork: false
   # kong_prefix empty dir size
-  prefixDir: 
+  prefixDir:
     sizeLimit: 256Mi
   # tmp empty dir size  
-  tmpDir: 
+  tmpDir:
     sizeLimit: 1Gi
 # Override namepsace for Kong chart resources. By default, the chart creates resources in the release namespace.
 # This may not be desirable when using this chart as a dependency.
--- a/charts/kong/values.yaml
+++ b/charts/kong/values.yaml
@@ -56,7 +56,7 @@ deployment:
   # kong_prefix empty dir size
   prefixDir:
     sizeLimit: 256Mi
-  # tmp empty dir size  
+  # tmp empty dir size
   tmpDir:
     sizeLimit: 1Gi
 # Override namepsace for Kong chart resources. By default, the chart creates resources in the release namespace.
--- a/charts/kong/values.yaml
+++ b/charts/kong/values.yaml
@@ -53,7 +53,12 @@ deployment:
   # Use a DaemonSet controller instead of a Deployment controller
   daemonset: false
   hostNetwork: false
-
+  # kong_prefix empty dir size
+  prefixDir: 
+    sizeLimit: 256Mi
+  # tmp empty dir size  
+  tmpDir: 
+    sizeLimit: 256Mi
 # Override namepsace for Kong chart resources. By default, the chart creates resources in the release namespace.
 # This may not be desirable when using this chart as a dependency.
 # namespace: "example"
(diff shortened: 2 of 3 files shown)


rainest:
What were you seeing these growing to in practice? What were the largest files in the mounts?

I'd prefer to make these static limits rather than configurable if possible, since they should be small and consistent enough in size that users shouldn't need to worry about configuring them. I don't know offhand what'd consume much space in those locations other than logs (which we want to divert to stdout and stderr)--I'm checking with the gateway engineers, but info from your instance would be helpful.

If we do end up using configurable values, they need to be in the default values.yaml and in documentation, else you'll hit the error that's currently failing in CI:

```
Error:  templates/: template: kong/templates/deployment.yaml:288:10: executing "kong/templates/deployment.yaml" at <include "kong.volumes" .>: error calling include: template: kong/templates/_helpers.tpl:414:25: executing "kong.volumes" at <.Values.deployment.prefixDir.sizeLimit>: nil pointer evaluating interface {}.sizeLimit
```

chaitanya14:
> What were you seeing these growing to in practice? What were the largest files in the mounts?
> 
> I'd prefer to make these static limits rather than configurable if possible, since they should be small and consistent enough in size that users shouldn't need to worry about configuring them. I don't know offhand what'd consume much space in those locations other than logs (which we want to divert to stdout and stderr)--I'm checking with the gateway engineers, but info from your instance would be helpful.
> 
> If we do end up using configurable values, they need to be in the default values.yaml and in documentation, else you'll hit the error that's currently failing in CI:
> 
> ```
> Error:  templates/: template: kong/templates/deployment.yaml:288:10: executing "kong/templates/deployment.yaml" at <include "kong.volumes" .>: error calling include: template: kong/templates/_helpers.tpl:414:25: executing "kong.volumes" at <.Values.deployment.prefixDir.sizeLimit>: nil pointer evaluating interface {}.sizeLimit
> ```

I don't have certain numbers on limits but we are facing disk pressure issues on AKS nodes. These directories may or may not be directly effecting our issue but as a best practice it is better to evict the pods if empty dirs are using more ephemeral storage  than defined. This will improve things in kubernetes cluster .

chaitanya14:
> > What were you seeing these growing to in practice? What were the largest files in the mounts?
> > I'd prefer to make these static limits rather than configurable if possible, since they should be small and consistent enough in size that users shouldn't need to worry about configuring them. I don't know offhand what'd consume much space in those locations other than logs (which we want to divert to stdout and stderr)--I'm checking with the gateway engineers, but info from your instance would be helpful.
> > If we do end up using configurable values, they need to be in the default values.yaml and in documentation, else you'll hit the error that's currently failing in CI:
> > ```
> > Error:  templates/: template: kong/templates/deployment.yaml:288:10: executing "kong/templates/deployment.yaml" at <include "kong.volumes" .>: error calling include: template: kong/templates/_helpers.tpl:414:25: executing "kong.volumes" at <.Values.deployment.prefixDir.sizeLimit>: nil pointer evaluating interface {}.sizeLimit
> > ```
> 
> I don't have certain numbers on limits but we are facing disk pressure issues on AKS nodes. These directories may or may not be directly effecting our issue but as a best practice it is better to evict the pods if empty dirs are using more ephemeral storage than defined. This will improve things in kubernetes cluster .

Also, we have a data plane where we have used 256Mi size of empty dirs that never caused an issue since 4 months. So IMO default to 256Mi is more meaningful.

charts/kong/values.yaml
```diff
@@ -53,7 +53,12 @@ deployment:
   # Use a DaemonSet controller instead of a Deployment controller
   daemonset: false
   hostNetwork: false
-
+  # kong_prefix empty dir size
+  prefixDir: 
+    sizeLimit: 256Mi
+  # tmp empty dir size  
+  tmpDir: 
+    sizeLimit: 256Mi
```
rainest, line 58:
```suggestion
    sizeLimit: 1Gi
```

rainest (changes requested):
We'll want to merge https://github.com/Kong/charts/pull/634 first. Several of those were wrong and would eventually fill the prefix on a long-lived Pod.

There's one unbounded consumer of /tmp/ space whose actual needs will vary depending on use characteristics, so we do indeed need to make these configurable. NGINX will buffer large client bodies to disk, and we [set no default upper bound on their size](https://github.com/Kong/kong/blob/5d721ac9ae1df36049013599d0253bd39cb6f758/kong.conf.default#L919-L949). An instance with a large number of concurrent large bodies (e.g. something that handles video uploads) is expected to consume a bunch of /tmp/ space even though most instances won't, so IMO we should give that one more breathing room.

charts/kong/values.yaml
```diff
@@ -53,7 +53,12 @@ deployment:
   # Use a DaemonSet controller instead of a Deployment controller
   daemonset: false
   hostNetwork: false
-
+  # kong_prefix empty dir size
+  prefixDir:
+    sizeLimit: 256Mi
+  # tmp empty dir size
```
```suggestion
  # tmp empty dir size
```

rainest (changes requested):

rainest (approved):

chaitanya14:
@rainest  When will this be released ?

rainest:
Probably end of this week or start of next; there are a few other things I'd like to work in if possible.

charts/kong/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+###
```
pmalek, line 5: We missed a section name in here I suppose

PR #651, opened 2022-07-26
Added RDS capabilities

GarethIMGArena:
- Values file is currently incompatible with GitOps tooling. Needs the env set inside the proxy.
- Added AWS Secret Manager capabilities to allow for external RDS clusters.

#### Checklist
- [x ] PR is based off the current tip of the `main` branch.


rainest:
Can you clarify exactly what issue you're running into in your environment?

For the AWSSecret resource, since it's a new custom resource that isn't closely integrated with an existing one (it's entirely derived from values.yaml, not some helper that produces names for another resource), we'd ask you to create the resource independent of the chart, or create an umbrella chart with ours as a child if you want it to be in the same release. To include it as part of our chart, we'd want an issue with interest from several users and, ideally, demonstration that doing so provides more functionality than creating the resource independent of the chart release.

For the other issue, it's not clear why the existing values layout would be incompatible with ArgoCD, or what this change would do. Values are only used if they're referenced by a template (e.g. [the current `.env` handler](https://github.com/Kong/charts/blob/aa8ac50d5d8e97493175ccf257c2dc86a8198668/charts/kong/templates/_helpers.tpl#L827-L831)) and this does not add any template code that uses `.Values.proxy.env`.

PR #656, opened 2022-08-03
docs(upgrade) Correct 2.4.0 upgrade with KIC enabled

didi-at-work:
When upgrading 2.3.0 with ingress-controller and postgresql enabled, the originally suggested helm upgrade always fails (for all my installations at least): the service account gets deleted and then the update of the deployment is failing because of the missing service account. Also tried with disabling pre- and post-migrations, but still did not work.

So I'd suggest doing a manual deployment update instead of the helm upgrade.

Documented 2.4.0 upgrade fails

- none -

#### Checklist

Dieter Rothacker dieter.rothacker@mercedes-benz.com, Mercedes-Benz Mobility AG on behalf of Mercedes-Benz Tech Innovation GmbH.
https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md

### DIFF
--- a/charts/kong/UPGRADE.md
+++ b/charts/kong/UPGRADE.md
@@ -150,19 +150,17 @@ proxy configuration. We recommend you establish an active maintenance window
 under which to perform this upgrade and inform users and stakeholders so as to
 avoid unexpected disruption.
 
-First, run an upgrade that keeps the current version, but disables the ingress
-controller:
+First, run an update on your deployment that disables the ingress controller:
 
 ```console
-$ helm upgrade --wait \
-  --set ingressController.enabled=false \
-  --version <CURRENT_CHART_VERSION> \
+$ kubectl patch deployment \
+  <YOUR_RELEASE_NAME>-kong \
   --namespace <YOUR_RELEASE_NAMESPACE> \
-  <YOUR_RELEASE_NAME> kong/kong
+  --type json -p='[{"op": "remove", "path": "/spec/template/spec/containers/0"}]'
 ```
 
-Once the upgrade completes you will only have the proxy itself running, and can
-run a second upgrade to re-enable the ingress controller and update to chart
+Once the update completes you will only have the proxy itself running, and can
+run a helm upgrade to re-enable the ingress controller and update to chart
 version 2.4:
 
 ```console


rainest:
Can you provide an example values.yaml and sequence of upgrade commands that demonstrate the issue?

Are you using a service mesh proxy? The ServiceAccount shouldn't be an issue, since only the controller (which is disabled) needs it, but service meshes usually also need a ServiceAccount, so that could explain the issue. You should be able to handle that by upgrading to the latest version of the chart, since it allows [enabling the ServiceAccount independent of the controller](https://github.com/Kong/charts/blob/1ccde0023443a1096b6fcb4293173a0c2beef2df/charts/kong/values.yaml#L23-L24).

I don't think we'd want a JSON patch to the Deployment because [Helm 3's patch strategy](https://helm.sh/docs/faq/changes_since_helm2/#improved-upgrade-strategy-3-way-strategic-merge-patches) should actually persist that change and keep the controller disabled forever.

didi-at-work:
Thanks! Of course the helm upgrade would be the much cleaner solution, but on 3 installations it has not worked yet :-(
We are not using a service mesh. It's just Kong ingress with a custom Kong image with API-Management plugin included. Deployed on AKS. Pulls the official KIC image from our internal CR.

Was initially installed as 2.3.0, so I tried an upgrade chain as:
```
helm upgrade --wait --set ingressController.enabled=false --version 2.3.0 --reuse-values -n ingress ingress-kong kong/kong

helm upgrade --set ingressController.enabled=true --version 2.4.0 -n ingress ingress-kong kong/kong

kubectl apply -f https://raw.githubusercontent.com/Kong/charts/kong-2.7.0/charts/kong/crds/custom-resource-definitions.yaml

kubectl delete ingressclass kong
helm upgrade -f ingress-kong-values.yml ingress-kong kong/kong
```

But the first command always leads to the service account being deleted and the deployment being stuck in "ReplicaFailure" state with this error message:
```
Error creating: pods "ingress-kong-kong-59f6589554-" is forbidden: error looking up service account ingress/ingress-kong-kong: serviceaccount "ingress-kong-kong" not found
```
I was able to get the failed helm upgrade resolved by executing:
`helm upgrade --wait --set migrations.preUpgrade=false --set migrations.postUpgrade=false --version 2.3.0 --reuse-values -n ingress ingress-kong kong/kong`
but then I'm at the start again.

This is one of the configurations in question (helm get values...):
```
env:
  database: postgres
  headers: "off"
  log_level: info
  nginx_http_include: /usr/local/kong/shared-dict.kong.conf
  oneapi_environment: development
  oneapi_gateway_key: EDITED
  oneapi_gateway_secret: EDITED
  oneapi_redis_host: redis-master
  oneapi_redis_password: EDITED
  oneapi_redis_port: 6379
  pg_database: kong
  pg_host: econpgsqlflexdevm.postgres.database.azure.com
  pg_password: EDITED
  pg_ssl: true
  pg_user: kong
  plugins: bundled, oneapi
image:
  repository: econcrdev.azurecr.io/oneapi/kong-oneapi
  tag: 2.8.1-2.5.0 # of course older image on original 2.3.0 deployment
ingressController:
  image:
    repository: econcrdev.azurecr.io/kong/kubernetes-ingress-controller
  installCRDs: false
podAnnotations:
  prometheus.io/path: /metrics
  prometheus.io/port: "8100"
  prometheus.io/scrape: "true"
proxy:
  annotations:
    service.beta.kubernetes.io/azure-dns-label-name: econ-devm-api
    service.beta.kubernetes.io/azure-load-balancer-resource-group: eConCloud-resources-eternal
  externalTrafficPolicy: Local
  loadBalancerIP: EDITED
replicaCount: 1
securityContext:
  runAsUser: 100
admin:
  tls:
    parameters: []
```

Dieter Rothacker dieter.rothacker@mercedes-benz.com, Mercedes-Benz Mobility AG on behalf of Mercedes-Benz Tech Innovation GmbH.
https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md

rainest:
Ah, looks like we'd actually moved the ServiceAccount config location earlier than I thought, and it was, coincidentally, moved in 2.4: https://github.com/Kong/charts/commit/097ae07d151a9f1ffb0a4006f321c3b6cb5562dd

Normally, changes to the default values.yaml are included in the upgrade, and you'd get `deployment.serviceaccount.create=true` without doing anything. I'm guessing `--reuse-values` bypasses that, and uses exactly the values you previously had, no new defaults included. If you set that manually, do you no longer see the issue?

didi-at-work:
> Ah, looks like we'd actually moved the ServiceAccount config location earlier than I thought, and it was, coincidentally, moved in 2.4: [097ae07](https://github.com/Kong/charts/commit/097ae07d151a9f1ffb0a4006f321c3b6cb5562dd)
> 
> Normally, changes to the default values.yaml are included in the upgrade, and you'd get `deployment.serviceaccount.create=true` without doing anything. I'm guessing `--reuse-values` bypasses that, and uses exactly the values you previously had, no new defaults included. If you set that manually, do you no longer see the issue?

Sorry for the late reply, was on holiday.
No, I'm not getting stuck at the 2.4.0 upgrade, but at the redeployment of 2.3.0 without KIC:
```helm upgrade --wait --set ingressController.enabled=false --version 2.3.0 --reuse-values -n ingress ingress-kong kong/kong```
this leads to the service account being deleted, but the old pod not being replaced by one without KIC because the k8s deployment misses the service account :-(

Same actually also happens when I do not use "reuse-values":
```helm upgrade --wait --set ingressController.enabled=false --set deployment.serviceaccount.create=true --version 2.3.0 -n ingress ingress-kong kong/kong```

rainest:
For some reason the ReplicaSet holds onto the ServiceAccount even though the Deployment no longer has it.

As an alternative path, are you able to upgrade if you set `ingressController.image.tag` to whatever version you're currently using, then upgrade to chart 2.4 or higher, and then run through the upgrade that disables the ingress controller and re-enables it with a 2.x version after?

That should allow you to avoid the ReplicaSet issue since chart 2.4 will actually honor `--set deployment.serviceaccount.create=true` and not delete it.

didi-at-work:
> As an alternative path, are you able to upgrade if you set `ingressController.image.tag` to whatever version you're currently using, then upgrade to chart 2.4 or higher, and then run through the upgrade that disables the ingress controller and re-enables it with a 2.x version after?

That actually worked fine!
So the first step I did (upgrading from 2.3.0 to 2.4.0):
```helm upgrade --set ingressController.image.tag=1.3 --version 2.4.0 -n ingress ingress-kong kong/kong```
which seemed to work well (I just had this up for some seconds, and proceeded with the further updates until 2.8.1)

So, maybe this would be an appropriate suggestion for the upgrade docs?

rainest:
Yep. If you wanted to revise this PR to reflect that, feel free. If you'd prefer I add it or I haven't seen an update here yet, I'll work the updated docs into the next release, which should be in about a week or so.

rainest:
https://github.com/Kong/charts/pull/663 now includes a revised set of instructions.

PR #663, opened 2022-08-31
chore(crds) update CRDs

rainest:
Adds pending 2.6 CRDs (IngressClassParameters, KongPlugin's new ordering section, new upstream fields in KongIngress, various description updates).

Adds a generic upgrade entry about CRD updates. They're basically all the same.

We actually forgot to add IngressClassParameters to the kustomization. This generation _did not_ run from main, and instead ran from:

```
kubectl kustomize github.com/kong/kubernetes-ingress-controller/config/crd?ref=30f06af6b73044f1fd11ff977920713b2d782a79 > crds/custom-resource-definitions.yaml
```
That tip doesn't include changes to the CRDs themselves, it just adds the existing ICP file to the kustomization.

#### Checklist

### DIFF
--- a/charts/kong/crds/custom-resource-definitions.yaml
+++ b/charts/kong/crds/custom-resource-definitions.yaml
@@ -1,15 +1,68 @@
 # generated using: kubectl kustomize github.com/kong/kubernetes-ingress-controller/config/crd?ref=main
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.9.2
+  creationTimestamp: null
+  name: ingressclassparameterses.configuration.konghq.com
+spec:
+  group: configuration.konghq.com
+  names:
+    kind: IngressClassParameters
+    listKind: IngressClassParametersList
+    plural: ingressclassparameterses
+    singular: ingressclassparameters
+  scope: Namespaced
+  versions:
+  - name: v1alpha1
+    schema:
+      openAPIV3Schema:
+        description: IngressClassParameters is the Schema for the IngressClassParameters
+          API
+        properties:
+          apiVersion:
+            description: 'APIVersion defines the versioned schema of this representation
+              of an object. Servers should convert recognized schemas to the latest
+              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+            type: string
+          kind:
+            description: 'Kind is a string value representing the REST resource this
+              object represents. Servers may infer this from the endpoint the client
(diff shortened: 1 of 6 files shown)


charts/kong/crds/custom-resource-definitions.yaml
```diff
@@ -1,15 +1,75 @@
-# generated using: kubectl kustomize github.com/kong/kubernetes-ingress-controller/config/crd?ref=main
```
shaneutt, line 1: We still use this to generate right? I think this was our only breadcrumb leading a reader back to how we actually include these CRDs in the chart, should we leave it?

charts/kong/crds/custom-resource-definitions.yaml
```diff
@@ -1,15 +1,75 @@
...
+apiVersion: apiextensions.k8s.io/v1
```
rainest, line 2:
```suggestion
# generated using: kubectl kustomize github.com/kong/kubernetes-ingress-controller/config/crd?ref=main
apiVersion: apiextensions.k8s.io/v1
```

charts/kong/crds/custom-resource-definitions.yaml
rainest: I forgot how file overwrites work. Whoops.

rainest:
No idea what happened to the Kuma setup, will need to review that.

rainest:
Now includes other assorted stuff in prep for 3.x:
- Upgrade doc fix that I didn't want to make a separate PR for.
- Setting `router_flavor` to match the KIC repo.
- Updating the CRDs again, because there were some more changes.
- Updating RBAC rules, because I forgot those for ICP earlier and we have access to many more Gateway things now.
- CI fixes and workarounds.

charts/kong/values.yaml
```diff
@@ -81,6 +81,11 @@ deployment:
 # is set below. In general, you should not set values here if they are set elsewhere.
 env:
   database: "off"
+  # the chart uses the traditional router (for Kong 3.x+) because the ingress
+  # controller generates traditional routes. if you do not use the controller,
+  # you may set this to "traditional_compatible" or "expression" to use the new
+  # DSL-based router
+  router_flavor: "traditional"
```
pmalek, line 88: Is there a doc in GW docs that we could link to here where the user could explore this option in detail at the source?

pmalek (approved):

charts/kong/values.yaml
rainest: There's https://docs.konghq.com/gateway/latest/key-concepts/routes/expressions/ but it's not really useful for explaining it in the controller context. We probably want something that covers the new router from our end, but there aren't 2.6 docs yet that we can link to from the chart comments.

rainest: Added a task in https://github.com/Kong/charts/pull/666 depending on how much we know about the docs site status for 2.6 docs.

PR #683, opened 2022-10-27
feat(deployment): Add ability to use a custom command within the Kong…

mattholmes15:

This updates the deployment.yaml template to allow a custom command to be supplied.

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -9,6 +9,7 @@
   [#682](https://github.com/Kong/charts/pull/682)
 * Supported `autoscaling/v2` API
   ([#679](https://github.com/Kong/charts/pull/679))
+* Add ability to supply a command to the Deployment.
 
 ## 2.13.1
 
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -12,7 +12,6 @@
 * Add ability to supply a command to the Deployment.
   ([#683](https://github.com/Kong/charts/pull/683))
 
-
 ## 2.13.1
 
 ### Improvements
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -10,6 +10,7 @@
 * Supported `autoscaling/v2` API
   ([#679](https://github.com/Kong/charts/pull/679))
 * Add ability to supply a command to the Deployment.
+  ([#683](https://github.com/Kong/charts/pull/683))
 
 ## 2.13.1
 
--- a/charts/kong/templates/deployment.yaml
+++ b/charts/kong/templates/deployment.yaml
@@ -123,6 +123,9 @@ spec:
       {{- if .Values.deployment.kong.enabled }}
       - name: "proxy"
         image: {{ include "kong.getRepoTag" .Values.image }}
+        {{- if .Values.command }}
+        command: {{ .Values.command }}
+        {{- end }}
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         securityContext:
         {{ toYaml .Values.containerSecurityContext | nindent 10 }}
--- a/charts/kong/templates/deployment.yaml
+++ b/charts/kong/templates/deployment.yaml
@@ -123,8 +123,8 @@ spec:
       {{- if .Values.deployment.kong.enabled }}
       - name: "proxy"
         image: {{ include "kong.getRepoTag" .Values.image }}
-        {{- if .Values.command }}
-        command: {{ .Values.command }}
+        {{- if .Values.deployment.command }}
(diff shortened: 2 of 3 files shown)


charts/kong/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
   [#682](https://github.com/Kong/charts/pull/682)
 * Supported `autoscaling/v2` API
   ([#679](https://github.com/Kong/charts/pull/679))
+* Add ability to supply a command to the Deployment
```
```suggestion
* Add ability to supply a command to the Deployment
  ([#683](https://github.com/Kong/charts/pull/683))
```

charts/kong/templates/deployment.yaml
```diff
@@ -123,6 +123,9 @@ spec:
       {{- if .Values.deployment.kong.enabled }}
       - name: "proxy"
         image: {{ include "kong.getRepoTag" .Values.image }}
+        {{- if .Values.command }}
+              command: {{ .Values.command }}
+        {{- end }}
```
pmalek: - If we go ahead with this change I'd suggest to use a more nested key name, e.g. `.Values.deployment.command`
- Please add this commented out to default `values.yaml`, and comment what this can be used for.

pmalek (commented):
Thanks for the PR 🙇 Just some minor suggestions, overall 👍

charts/kong/CHANGELOG.md
mattholmes15: Thanks for the quick response - updated

charts/kong/templates/deployment.yaml
mattholmes15: Updated to use the more nested key name, and have also added this into the values.yaml file commented out. I've also fixed the indentation issue on L127

charts/kong/CHANGELOG.md
```diff
@@ -9,6 +9,9 @@
   [#682](https://github.com/Kong/charts/pull/682)
 * Supported `autoscaling/v2` API
   ([#679](https://github.com/Kong/charts/pull/679))
+* Add ability to supply a command to the Deployment.
+  ([#683](https://github.com/Kong/charts/pull/683))
+
```
pmalek, line 14:
```suggestion
```

rainest:
@mattholmes15 can you share what you'd need this for? We've rejected [similar requests](https://github.com/Kong/charts/pull/601) in the past on the basis that most users should not need to change the standard command, so it's something that we'd prefer to keep out of the config surface. Unless there's a common use case that requires exposing the entire command, we'll probably continuing to recommend handling that at the image level.

mattholmes15:
> @mattholmes15 can you share what you'd need this for? We've rejected [similar requests](https://github.com/Kong/charts/pull/601) in the past on the basis that most users should not need to change the standard command, so it's something that we'd prefer to keep out of the config surface. Unless there's a common use case that requires exposing the entire command, we'll probably continuing to recommend handling that at the image level.

Hey @rainest,

The requirement behind this is that we want the helm deployment to have different start up actions depending on whether its a control plane, data plane or portal, as we run Kong in Hybrid mode. We also support deployments to different platforms, EC2, ECS and EKS. Having this control would allow us to define a different 'entrypoint' for the Helm deployment, due to the nature of how EKS handles the secrets, as opposed to as an example, writing out certificates to file within EC2 user data or within an ECS Task Definition.
A couple of alternatives are to run with one large entrypoint within the Docker image with all of the logic to workout what platform it's being deployed to, or to run a different image.

rainest:
What are those different startup actions? We'd normally expect all the different hybrid mode settings from the environment, as shown in the [control plane](https://github.com/Kong/charts/blob/main/charts/kong/example-values/minimal-kong-hybrid-control.yaml) and [data plane](https://github.com/Kong/charts/blob/main/charts/kong/example-values/minimal-kong-hybrid-data.yaml) examples.

What are you running to handle the EKS secrets? I expect that's a separate script or application rather than a modified invocation of the `kong` binary. If that's the case, are you able to handle it with a [custom initContainer](https://github.com/Kong/charts/blob/a875d8dddc438303fd8f83d8df8d9ba0fb838999/charts/kong/values.yaml#L41-L43)? You should be able to, for example, write an initContainer that pulls a resource from AWS, writes it to a file in a mounted emptyDir volume, and then mount the same emptyDir volume in the Kong proxy container after.

The Helm chart doesn't target ECS at all, but if deploying the same image to both would require a custom entrypoint anyway, that is probably the approach we'd recommend--either with layers over top of a base entrypoint that modify its behavior (i.e. different images for the different platforms), passing the non-default command in ECS rather than in the chart, or by adding envvar-based switches to the [command processing script](https://github.com/Kong/docker-kong/blob/81a0372da8b3738538ba06d2f7b75d618c50dc69/docker-entrypoint.sh#L31-L57).

PR #684, opened 2022-10-27
Kube service appProtocol field breaking IBM NLB creation

elopsod:
  - fixes #

#### Checklist

### DIFF
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -165,8 +165,10 @@ spec:
   {{- if .http.enabled }}
   - name: kong-{{ .serviceName }}
     port: {{ .http.servicePort }}
-    targetPort: {{ .http.containerPort }}
+    targetPort: {{ .http.containerPort }}-{{ .http.appProtocol }}
+  {{- if .http.appProtocol }}
     appProtocol: http
+  {{- end }}
   {{- if (and (or (eq .type "LoadBalancer") (eq .type "NodePort")) (not (empty .http.nodePort))) }}
     nodePort: {{ .http.nodePort }}
   {{- end }}
@@ -177,7 +179,9 @@ spec:
   - name: kong-{{ .serviceName }}-tls
     port: {{ .tls.servicePort }}
     targetPort: {{ .tls.overrideServiceTargetPort | default .tls.containerPort }}
+  {{- if .tls.appProtocol }}
     appProtocol: https
+  {{- end }}
   {{- if (and (or (eq .type "LoadBalancer") (eq .type "NodePort")) (not (empty .tls.nodePort))) }}
     nodePort: {{ .tls.nodePort }}
   {{- end }}
@@ -422,10 +426,10 @@ The name of the service used for the ingress controller's validation webhook
 
 {{- define "kong.volumes" -}}
 - name: {{ template "kong.fullname" . }}-prefix-dir
-  emptyDir: 
+  emptyDir:
     sizeLimit: {{ .Values.deployment.prefixDir.sizeLimit }}
 - name: {{ template "kong.fullname" . }}-tmp
-  emptyDir: 
+  emptyDir:
     sizeLimit: {{ .Values.deployment.tmpDir.sizeLimit }}
 {{- if and ( .Capabilities.APIVersions.Has "cert-manager.io/v1" ) .Values.certificates.enabled -}}
 {{- if .Values.certificates.cluster.enabled }}
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -166,7 +166,7 @@ spec:
   - name: kong-{{ .serviceName }}
     port: {{ .http.servicePort }}
     targetPort: {{ .http.containerPort }}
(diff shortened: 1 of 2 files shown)


jrsmroz:
Hi. Can you create an issue that describes what happened ? The PR lacks context to understand what it tries to solve.

randmonkey (commented):
Please update `README.md` and comments to show what are added, and update `CHANGELOG.md` to record the update.

elopsod:
> Hi. Can you create an issue that describes what happened ? The PR lacks context to understand what it tries to solve.

hi @jrsmroz 
it is IBM specific problem
to resolve it we need just remove appProtocol from service manifest
details can be found here
https://github.com/kubernetes/ingress-nginx/issues/7885

jrsmroz (commented):
Please take a look at the PR checklist. We also use github issue to track problems. That way we can search for issues we had in past. Please create an issue that describes the problem.

charts/kong/values.yaml
```diff
@@ -282,6 +283,7 @@ proxy:
     enabled: true
     servicePort: 443
     containerPort: 8443
+    appProtocol: true
```
jrsmroz: If it is used for enabling anything, the name should be more descriptive.

charts/kong/templates/_helpers.tpl
```diff
@@ -166,7 +166,9 @@ spec:
   - name: kong-{{ .serviceName }}
     port: {{ .http.servicePort }}
     targetPort: {{ .http.containerPort }}
-    appProtocol: http
+  {{- if .http.appProtocol }}
+    appProtocol: {{ .http.appProtocol }}
+  {{- end }}
```
jrsmroz: Why would we set the `appProtocol` to `true`? It's not a valid value.

elopsod: you are right
fixed typo

charts/kong/values.yaml
randmonkey: So the `appProtocol` here is set to `true` to configure the `appProtocol` field to a certain value (same as actual protocol used), `false` to leave it empty? I think this is not direct at to configure it to be the actual value used, so I think here should use a more descriptive key, or add clear comment for it.

randmonkey (commented):
Don't forget updating README and CHANGELOG.

rainest:
@elops-od do you have additional information from IBM about this? Are they aware of the limitation, and do they have plans to address it either with support for the field, a change to a graceful failure, or some vendor-specific annotation to indicate other behavior?

`appProtocol` is a [standard part of the Kubernetes Service definition](https://kubernetes.io/docs/concepts/services-networking/service/#application-protocol) and has been stable since 1.20, so it's stable in all currently supported versions.

IBM's cloud provider code does look pretty straightforward here, in that it [rejects any Service with that field set outright](https://github.com/IBM-Cloud/cloud-provider-ibm/blob/105846ee9c181c142fce3df6a4b9d6cbba8299c3/ibm/ibm_loadbalancer.go#L2352-L2354) and [confirms this behavior in tests](https://github.com/IBM-Cloud/cloud-provider-ibm/blob/105846ee9c181c142fce3df6a4b9d6cbba8299c3/ibm/ibm_loadbalancer_test.go#L2268-L2279). That's arguably in violation of the spec--we aren't misrepresenting a non-HTTP server as HTTP, and we generally don't want to expose configuration of this because it's correct for cloud providers that handle the field well.

charts/kong/values.yaml
elopsod: @randmonkey hi
what do you think about key name `appProtocolEnable`  ?

randmonkey: > @randmonkey hi what do you think about key name `appProtocolEnable` ?

IMO `setAppProtocol` would be clearer.

elopsod: > > @randmonkey hi what do you think about key name `appProtocolEnable` ?
> 
> IMO `setAppProtocol` would be clearer.

agree
added changes

elopsod:
>@elops-od do you have additional information from IBM about this? Are they aware of the limitation, and do they have plans to address it either with support for the field, a change to a graceful failure, or some vendor-specific annotation to indicate other behavior?

yes, IBM knows about this limitation, they propose to remove appProtocol field as workaround.
IBM do not have any plans to fix this as I know

rainest:
I reached out to them [in the SIG Cloud Provider Slack channel](https://kubernetes.slack.com/archives/C718BPBQ8/p1667249493653939):

> we are planning to change this behavior and ignore the field altogether

with plans to discuss it further at the [SIG meeting](https://github.com/kubernetes/community/blob/master/sig-cloud-provider/README.md#meetings) on the 9th, so there's maybe a disconnect between the front line support and what their engineers have planned.

As an interim solution, are you able to edit the services manually after install? Helm 3 [should keep those edits in place](https://helm.sh/docs/faq/changes_since_helm2/#improved-upgrade-strategy-3-way-strategic-merge-patches) through upgrades.

rainest:
Closing this as IBM plans to address the issue at the provider level, and there hasn't been any discussion for a while.

PR #696, opened 2022-11-30
Dbless toyaml

alice-sawatzky:

Add support for non-string types of `dblessConfig.config` values. This allows users to pass in the config structure instead of having to render it to yaml beforehand; this can be useful as it allows customization of a base config using `--set` or extra values files.

This PR depends on #695 since they're both making changes to the same code, to avoid merge conflicts don't bother looking at this until #695 is merged

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -2,8 +2,12 @@
 
 ## Unreleased
 
+### Improvements
+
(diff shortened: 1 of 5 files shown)


rainest:
To confirm, what would the set in question look like? I'm not sure offhand what sort of non-string you'd provide--it probably makes sense to describe what that alternative actually looks like in the comment, rather than the mechanics (`toYaml`) of how the template achieves it.

PR #712, opened 2023-01-23
feat(values) Support string values for affinity and topologySpreadConstraints

ubergesundheit:
#### What this PR does

This PR adds support for both string and object/array values for `affinity` and `topologySpreadConstraints`. String values will be fed through helms [`tpl` function](https://helm.sh/docs/howto/charts_tips_and_tricks/#using-the-tpl-function) which allows users to use both helpers and chart values. This can be used for example to select a specific version or kong deployment.

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -8,6 +8,10 @@
   ([#706](https://github.com/Kong/charts/pull/706))
 * Prevent supplying duplicate plugin inclusion to `KONG_PLUGINS` env variable.
   ([#711](https://github.com/Kong/charts/pull/711))
+* Allow specifying affinity and topologySpreadConstraints as strings through
+  values.yaml. String values are fed through helms `tpl` function to allow
+  using both values and helpers.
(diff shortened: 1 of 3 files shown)


rainest (changes requested):
Would `toYaml` be sufficient for these instead of `tpl`? Do you expect you'd need full templating for these subsections of the Deployment resource?

ubergesundheit:
Hi, thank you for your comment! The reason tpl is used is because I want to use .Chart.version and the release name for affinity and topology constraints. I have multiple Kong deployments on a single cluster which need to be balanced across nodes

rainest:
Are you able to handle those out-of-band? If you need multiple values.yaml configurations, you can generate them prior to installing the chart.

ubergesundheit:
This change would allow us to handle these things without introducing additional tooling.

rainest:
Understood, but unless there's demonstrated need from a number of users for the same functionality we default ask you put that additional tooling in place.

PR #714, opened 2023-01-23
fix(dbless) remove unnecessary failure condition

rainest:
Community change fails configs with DB-less but no static config, but it shouldn't. Config can be handled outside the chart.


pmalek (commented):
👍

PR #715, opened 2023-01-24
fix(dbless) disable mounts when not in use

rainest:
Also, the chart attempted to mount configs that didn't exist.

Need post-mortem CI review to determine why that didn't fail tests. We likely do not test DB-less installs with neither a controller nor a config.


pmalek (commented):
👍

PR #719, opened 2023-01-25
chore(ci): bump CI dependencies, disable anonymous reporting in tests and some indentation fixes

pmalek:

- bump helm to 3.11
- bump python to 3.11
- bump kind to 0.17.0
- bump Gateway API CRDs to 0.5.1 experimental
- indentation fixes (putting each `--set` on a separate line)
- disable anonymous reporting in CI tests

#### Checklist

### DIFF
--- a/.github/workflows/main-pr.yaml
+++ b/.github/workflows/main-pr.yaml
@@ -17,14 +17,14 @@ jobs:
       - name: Set up Helm
         uses: azure/setup-helm@v3
         with:
-          version: v3.9.0
+          version: v3.11.0
 
       - name: Add Bitnami
         run: helm repo add bitnami https://charts.bitnami.com/bitnami
 
       - uses: actions/setup-python@v4
         with:
-          python-version: "3.10"
+          python-version: "3.11"
 
       - name: Set up chart-testing
         uses: helm/chart-testing-action@v2.3.1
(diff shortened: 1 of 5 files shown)


.github/workflows/main-pr.yaml
```diff
@@ -51,19 +51,17 @@ jobs:
         run: ct install --all
 
   integration-test:
-    # Not a real requirement, but due to worker CPU contention, this can fail if it runs concurrent with lint-test
-    needs: lint-test
```
pmalek, line 55: Since jobs are run an undetermined set of workers (if we're not using self hosted workers, which is the case for now) then we shouldn't worry about this right @rainest ?

czeslavo (approved):

.github/workflows/main-pr.yaml
rainest: Unsure. IIRC GitHub Actions workers can get ornery when running multiple KIND instances simultaneously, which is what I think this was for. On the other hand, GitHub presumably upgrades their own infrastructure, maybe this runs fine now without the block. We'll see if tests start failing in weird ways, but probably fine to try without it.

They aren't failing in this run, so it's not some critical load-bearing requirement, may as well unblock it and see what happens.

PR #720, opened 2023-01-26
Release 2.15.3 to fix volume mount issue

### LINKED ISSUE #717: Version ^2.15.0 breaks mounting of cluster certificate secrets
The change in https://github.com/Kong/charts/pull/695 seems to be breaking for a hybrid setup:
```
Error: cluster_cert: failed loading certificate from /etc/secrets/kong-cluster-cert/tls.crt
```
The cluster secrets are not mounted in the `kong-init-migrations` pod and therefore the deployment fails.

We used the template provided by https://github.com/Kong/charts/blob/main/charts/kong/example-values/minimal-kong-enterprise-hybrid-control.yaml together with the corresponding data plane.

After downgrading to helm charts v2.14.0 everything works fine again.

rainest:

DB-less config stuff accidentally started chopping off volume mounts because YAML template if blocks are pain. This makes the volumes and their mounts not disappear if you don't have a particular configuration, and tries to make the template whitespace a bit clearer.

  - fixes #717 

Significant whitespace within templates that have complex nest if blocks because YAML was a mistake. Supplemental reports from @ahuffman.

#### Checklist
- [x] Changes are documented under the ~"Unreleased"~ (patch release, adds its own section and new unreleased section) header in CHANGELOG.md
- ~New or modified sections of values.yaml are documented in the README.md~ none

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -2,9 +2,13 @@
 
 ## Unreleased
 
+
+## 2.15.3
+
 ### Fixed
 
 * Changed `ingressController.readinessProbe` to use `/readyz` to prevent pods from becoming ready and serving 404s prior to the `ingress-controller` first syncing config to the `proxy` [#716](https://github.com/Kong/charts/pull/716).
+* Fixed incorrect `if` block order in volume mount templates.
 
 ## 2.15.2
 
--- a/charts/kong/Chart.yaml
+++ b/charts/kong/Chart.yaml
@@ -11,7 +11,7 @@ maintainers:
 name: kong
 sources:
 - https://github.com/Kong/charts/tree/main/charts/kong
-version: 2.15.2
+version: 2.15.3
 appVersion: "3.1"
 dependencies:
 - name: postgresql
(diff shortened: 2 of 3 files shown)


pmalek (approved):

ahuffman:
Validated this fixes the secretVolumes mounting issue from 2.15.2.  Thank you @rainest

ghost:
Could it be this issue is not fully fixed?

```
1s          Warning   FailedMount   pod/kong-kong-c7455ddb-4rfzl                           MountVolume.SetUp failed for volume "kong-kong-proxy-cert" : secret "kong-kong-proxy-cert" not found
```

I still get this error, it won't start because proxy-cert is not created.

I'm pretty sure proxy-cert is not created because it can't perform the challenge if the proxy is not running, but I'm not 100% sure. This system used to work, but we recreated our cluster last week and now it doesn't anymore.

rainest:
@TincoAeroscan don't think that's related. This is only for the config mount, not the certificate mounts. Can't say for sure without a values.yaml, but that sounds like a mount you've configured separately, and your cluster refresh deleted the Secret it was previously using.

ghost:
@rainest sorry I misinterpreted the issue. I thought it was not executing the challenge at all. I gave up for the day after not figuring out the problem and making this comment. Then when I came back to work the next morning the request had actually succeeded and the website was online and happily accepting https requests... so yeah :P

PR #728, opened 2023-02-01
feat(template) only allow strings in environment variables

### LINKED ISSUE #701: Validate that env values are strings (and not bools/numbers)
Values with unexpected types render templates may cause unexpected behavior in developing.
For example, users may set `env.proxy_access_log` to `"off"` turn off access logs, but if the value is configured to `off` (no quotes), it will be rendered to `"false"` in the final manifests. This will cause Kong to print logs to `/kong-prefix/false`, which is unexpected.
We could add validations to guarantee that only strings are used in the value. 

## Acceptance criteria
- [ ] `env` and `ingress_controller.env` disallow values of types other than `string`

randmonkey:

only allow strings(not `int` or `bool`) in environment variables to avoid unwanted modifications on the values, like `off` (without quotes, meaning boolean `false`) changed to `"false"` in envs. 

  - fixes #701 

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -8,7 +8,7 @@
   `.Values.ingressController.env.*` (to set environment variables of ingress
   controller container) only allow strings. Specifying values with other types
   (bool, int) will raise error in rendering the template.
-  If you want to set an enviroment varialble to a numerical value or keywords
+  If you want to set an environment variable to a numerical value or keywords
   for boolean values  (like `true`,`yes`,`off`) via helm command, please use 
   `--set-string` flag.
   [#728](https://github.com/Kong/charts/pull/728)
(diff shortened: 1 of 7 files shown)


charts/kong/CHANGELOG.md
```diff
@@ -2,6 +2,17 @@
 
 ## Unreleased
 
+### Breaking Changes
+
+* `.Values.env.*` (to set environment variables of Kong proxy container) and 
+  `.Values.ingressController.env.*` (to set environment variables of ingress
+  controller container) only allow strings. Specifying values with other types
+  (bool, int) will raise error in rendering the template.
+  If you want to set an enviroment varialble to a numerical value or keywords
```
```suggestion
  If you want to set an environment variable to a numerical value or keywords
```

pmalek (approved):

PR #729, opened 2023-02-01
fix(migration-jobs) Respect automountServiceAccountToken setting

ubergesundheit:
Correctly set the value of
.Values.deployment.serviceAccount.automountServiceAccountToken in migrations Jobs.

This was already the case for the Deployment from which the template code has been copied.

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -19,6 +19,8 @@
   [#721](https://github.com/Kong/charts/pull/721)
 * Replaced static secret with projected volume in deployment.
   [#722](https://github.com/Kong/charts/pull/722)
+* Respect setting `.Values.deployment.serviceAccount.automountServiceAccountToken` in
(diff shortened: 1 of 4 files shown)


randmonkey:
@ubergesundheit please update the CHANGELOG again since there are conflicts now.

ubergesundheit:
I've just merged the latest commits from `main` into my branch :+1:

rainest (approved):
This is probably a good case for consolidating the SAT sections into a helper template, since it should be consistent across the containers and AFAIK has no special cases where it needs to vary across them. That should avoid the original problem that prompted this in the future.

However, in the interest of time and wanting to fix the problem, and with the expectation that this area of the template doesn't change much, and seeing that the if blocks aren't complex, copy/pasted blocks suffice for the time being. Good possible area for consolidation in the future, but copy/paste if blocks suffice for now.

PR #735, opened 2023-02-06
fix: make semverCompare handle GKE versions and release 2.16.1

czeslavo:

Fixes `semverCompare` call that decides whether to create a projected volumeMount for the service account. It didn't work on GKE where the expected version format is `vX.Y.Z-gke.N`. It was resulting in:
- using a reference to secret in the deployment volume (`deployment.yaml`)
- not creating a secret (`secret-sa-token.yaml`)

which made deployment fail to become ready because of the missing secret to be mounted.

Regression was introduced in https://github.com/Kong/charts/pull/722.

See related upstream issues: 
- https://github.com/Masterminds/semver#working-with-prerelease-versions
- https://github.com/helm/helm/issues/3810

Fixes bug discovered in KTF CI run: https://github.com/Kong/kubernetes-testing-framework/actions/runs/4102865692/jobs/7076352608

#### Checklist
- ~[ ] New or modified sections of values.yaml are documented in the README.md~

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -1,8 +1,11 @@
 # Changelog
 
-## Unreleased
+## 2.16.1
 
-Nothing yet.
+### Fixed 
+
+* serviceAccount projected volume is properly provisioned for GKE clusters >= 1.20.
+  [#735](https://github.com/Kong/charts/pull/735)
 
 ## 2.16.0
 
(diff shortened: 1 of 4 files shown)


pmalek (dismissed):
Nice. Don't we test on gke to detect such things?

pmalek (approved):

czeslavo:
> Nice. Don't we test on gke to detect such things?

Nah, seems like charts are tested only against KinD, and in KIC we don't use the chart to install the controller (only proxy).

PR #745, opened 2023-02-14
feat: add support for single controller deployments

### LINKED ISSUE #710: Single controller deployments support in kong helm chart
When https://github.com/Kong/kubernetes-ingress-controller/issues/702 gets implemented and released, it should get full support in this helm chart.

This issue tracks the effort of adding support of said feature.

### Proposed Solution

Similarly to https://github.com/Kong/kubernetes-ingress-controller/issues/3423

- Support https://github.com/Kong/kubernetes-ingress-controller/issues/702 as an opt-in mechanism
  - Users will be able to change a setting in helm chart to enable non-sidecar model of deployment for KIC with Gateway(s)
- After sufficient soak time (several minor versions?) change the above mentioned setting to an opt-out: default value will change
  - This will be a **breaking change** hence this has to be done as a major release

## Acceptance criteria

- [ ] As a user I can deploy KIC with multiple Kong Gateways using kong's helm chart using an opt-in setting in the helm chart
- [ ] As a user I can deploy KIC with multiple Kong Gateways using kong's helm chart by default

pmalek:

This PR introduces support for single controller deployments through `deployment.singleController.enabled` setting.

Closes #710

This is expected to not work (i.e. `deployment.singleController.enabled` set to `true` should install successfully but KIC's pods will fail) until KIC 2.9 gets released.

One can test it with the following values:

```yaml
admin:
  enabled: true
  type: ClusterIP
  clusterIP: None

deployment:
  singleController:
    enabled: true
  kong:
    replicaCount: 5

  controller:
    replicaCount: 2

ingressController:
  image:
    repository: kong/nightly-ingress-controller
    tag: nightly
    effectiveSemver: "2.8.1"
```

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## Unreleased
+
+### Improvements
+
+* Added support for single controller deployments.
+  With `deployment.singleControler.enabled` set to `true` you can instruct helm
+  to produce separate deployments for Kong Gateway and Kong Ingress Controller.
+  [#744](https://github.com/Kong/charts/pull/744)
+
 ## 2.16.4
 
 ### Fixed
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -21,6 +21,16 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
 {{- default (printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-") .Values.fullnameOverride -}}
 {{- end -}}
 
+{{- define "kong.deployment.proxy.fullname" -}}
+{{- $name := default .Chart.Name .Values.nameOverride -}}
+{{- default (printf "%s-%s-proxy" .Release.Name $name | trunc 63 | trimSuffix "-") .Values.fullnameOverride -}}
+{{- end -}}
+
+{{- define "kong.deployment.controller.fullname" -}}
+{{- $name := default .Chart.Name .Values.nameOverride -}}
+{{- default (printf "%s-%s-controller" .Release.Name $name | trunc 63 | trimSuffix "-") .Values.fullnameOverride -}}
+{{- end -}}
+
 {{- define "kong.chart" -}}
 {{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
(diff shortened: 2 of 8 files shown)


rainest:
Do you know of a reason we'd definitely _need_ the separate template and dedicated configuration? Having, effectively, a separate `deployment.yaml` template will make future maintenance quite difficult. Historically the chart had a lot of problems stemming from maintainers forgetting to update each of N similar templates: we used to have separate templates for each of the Services, and repeatedly ran into issues where someone would add functionality to (arbitrarily) the admin Service template only, and later discovering that it should have been added to every Service template.

Copying and pasting between those was quite error-prone, so they were ultimately [consolidated into a single helper template with different inputs](https://github.com/Kong/charts/pull/251). That approach is viable, but writing those flexible templates is difficult because template reasons. This _could_ try to use a unified Deployment template, but I expect it'd be quite difficult. Fortunately, it's probably unnecessary!

For any topology more complex than the basic "Kong instance plus KIC instance" Deployment strategy, we typically use [split releases](https://github.com/Kong/charts/blob/main/charts/kong/README.md#separate-admin-and-proxy-nodes). This requires additional configuration, but supports a wide array of topologies (the split admin/proxy setup, hybrid, separate controller and Kong Deployments, etc.) without handling each topology in templates. Avoiding templates handling that complexity is good, because templates are, broadly, hard to modify. Templates are difficult to design effectively on our end, and de facto impossible to modify on the user end--Helm was ostensibly designed with the expectations that end users _would_ modify templates as needed, but in practice almost nobody does (users get hit with both the general difficulty of maintaining templates _and_ needing to handle upstream changes if they make their own modifications).

AFAICT the needs of SCD should be possible to satisfy with separate releases that are loosely-coupled via configuration. That configuration should be a one-off task at install time, so the user cost is pretty low. If we can avoid needing to maintain separate Deployment templates or a complex unified Deployment template by requiring a small amount of initial configuration, we should. Initial configuration is cheap, complicated templates are expensive.

That looks viable using a quick edit to stuff the Kong admin Service into the controller environment and then deploying each separately: https://gist.github.com/rainest/46aa0a2d4a5eb864e7e5f964fa1ce1ce

That probably warrants more thought into how we design the values.yaml controller configuration (I just made some arbitrary edits to get the controller environment I wanted without really thinking about making the config intuitive and/or easily explained), but I don't see anything that'd make it not viable.

I think there's some additional stuff I'm missing on the Kong side. I did need to piggyback off your work to get the controller to recognize the Kong instances, but that should be solvable. I initially tried deploying Kong with the current chart release and just setting `CONTROLLER_PUBLISH_SERVICE` instead of `CONTROLLER_KONG_ADMIN_URL` and got errors about missing Endpoints (even though there were admin Service Endpoints), but I suspect that relies on the selector label changes that are in this PR but not in the current chart release.

pmalek:
>  That approach is viable, but writing those flexible templates is difficult because template reasons

True.

I can agree (without working too much on this chart) that what you're describing (helm template pain) is highly likely to increase with this patch :)

As an alternative I propose #747. PTAL at the description where I describe what would be the usage of this. If that's fine I can continue with documenting this in README.md (and elsewhere, docs.konghq.com?)

pmalek:
Closing, as we decided to use #747.

PR #747, opened 2023-02-15
feat: add support for single controller deployments

### LINKED ISSUE #710: Single controller deployments support in kong helm chart

pmalek:

This PR introduces support for single controller deployments through `ingressController.serviceDiscovery` setting. Added options are as follows:

```yaml
ingressController:
  serviceDiscovery:
    enabled: false
    adminApiService:
      namespace: ""
      name: ""
```

In order to leverage this use case you have to deploy proxy and controller in 2 separate helm releases.

Exemplar values files:

- proxy - deployed with a release name `gw`

  ```yaml
  admin:
    enabled: true
    type: ClusterIP
    clusterIP: None
  
  ingressController:
    enabled: false
  
  # replicaCount: 3
  # or use autoscaling
  autoscaling:
    enabled: true
  ```

- controller - deployed with a release name `kic`

  ```yaml
  deployment:
    kong:
      enabled: false
  
  proxy:
    ## This has to match the full name of the proxy service deployed as part
    ## of the release above. Below matches release called: "gw"
    nameOverride: gw-kong-proxy
  
  ingressController:
    enabled: true
    serviceDiscovery:
      enabled: true
      adminApiService:
        ## optionally specify namespace, otherwise this will use the same
        ## as the controller release here.
        # namespace: kong
        ## This has to match the full name of the admin service deployed as part
        ## of the release above. Below matches release called: "gw"
        name: gw-kong-admin
  
  ## 2 for HA :)
  # replicaCount: 2
  ## or use autoscaling
  autoscaling:
    enabled: true
  ```

Closes #710

This is expected to not work (i.e. `ingressController.serviceDiscovery.enabled` set to `true` will install successfully but KIC's pods will fail) until KIC 2.9 gets released.

#### Checklist


rainest:
https://github.com/Kong/charts/pull/750 to fill out the examples.

I am not quite sure what is going in testing. This appears fine, but I'm weirdly _not_ able to get the example working (after subbing in the nightly image). I just continuously get:

```
time="2023-02-16T01:06:15Z" level=error msg="failed to create kong client(s)" error="no endpoints for kong admin service: \"default/gw-kong-admin\""
time="2023-02-16T01:06:16Z" level=error msg="failed to create kong client(s)" error="no endpoints for kong admin service: \"default/gw-kong-admin\""
time="2023-02-16T01:06:17Z" level=error msg="failed to create kong client(s)" error="no endpoints for kong admin service: \"default/gw-kong-admin\""
```

when actually installing the example in my PR.

I can't really explain this unless we happened to break nightly in a very particular way. Very confusingly, I tried to figure out what was going wrong by running `make debug` with the same `--kong-admin-svc=default/gw-kong-admin`, and it proceeded to find endpoints successfully and sync config as normal.

Thinking maybe the template was broken and somehow mangling the service info (it wasn't--the weird escape quotes there are because the log uses `%q` and maybe shouldn't), I tried forcing the code to always log that line, and compare the output. The output was identical.

I then proceeded to build a Delve variant of the image to ensure running outside a Pod wasn't somehow interfering with the cause of the problem, but nope: running with the Delve image successfully pulled endpoints and synced config as normal.

So I'm _pretty_ sure the example in my PR is valid and that the original changes here are good, but didn't push it directly because apparently it only works with a very atypical image.

pmalek:
As noted on slack: https://github.com/Kong/kubernetes-ingress-controller/pull/3556 is what has been missing from the nightly build. It should be included already (as of 16-02-2023).

charts/kong/README.md
```diff
@@ -725,6 +725,9 @@ section of `values.yaml` file:
 | userDefinedVolumes                      | Create volumes. Please go to Kubernetes doc for the spec of the volumes                                                                                  |                                    |
 | userDefinedVolumeMounts                 | Create volumeMounts. Please go to Kubernetes doc for the spec of the volumeMounts                                                                        |                                    |
 | terminationGracePeriodSeconds           | Sets the [termination grace period](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#hook-handler-execution) for Deployment pod | 30                                 |
+| serviceDiscovery.enabled                | Enables Kong instance service discovery                                                                                                                  |  false                             |
```
mflendrich: > Enables Kong instance service discovery

as a user, after reading this, how can I tell if I want "service discovery" or not? how do I find the place where I will get that information?

suggested reword: "If false: the ingress controller will control a pre-determined set of Gateway instances based on Admin API URLs. If true: the ingress controller will dynamically locate Gateway instances by watching the specified Kubernetes service."

If too long, maybe make it more concise and point at the _The `serviceDiscovery` section_ section of the README.

charts/kong/README.md
```diff
@@ -746,6 +749,19 @@ kong:
       TZ: "Europe/Berlin"
 ```
 
+#### The `serviceDiscovery` section
+
+With Kong Ingress Controller v2.9 a new way of configuration has been introduced
```
mflendrich: this section is heavily based on the assumption that the user knows the "old" way and we're showing the "new" as a diff versus the old.

I think it would be more productive to show the two modes in an equally descriptive fashion, and make a remark that "the new one" is only valid in KIC >= 2.9

charts/kong/example-values/README.md
```diff
@@ -52,5 +52,15 @@ common Kong deployment scenarios on Kubernetes.
   restarts them, so starting control planes second, but around the same time
   will usually work), but is the smoothest option.
 
+* [minimal-kong-scd-controller.yaml](minimal-kong-scd-controller.yaml) and
+  [minimal-kong-scd-gateway.yaml](minimal-kong-scd-gateway.yaml) install a
```
mflendrich: "SCD" means nothing outside of KIC dev. I'd say that this is an internal feature codename more than anything.

suggestion: replace "SCD" with some variation of "service discovery".

BTW I'm inclined to think that "gateway discovery" might be more accurate/helpful than "service discovery". Naming things is hard 🤔

mflendrich (commented):
Looks good, attaching some usability feedback to consider

charts/kong/example-values/README.md
czeslavo: I like "gateway discovery" 👍

pmalek: I've changed `scd` to `sd` as in service discovery. This is already used throughout the documentation that is added. I guess it should be fine?

charts/kong/README.md
pmalek: Rephrased this a bit to now refer to this feature as "the new way". PTAL

charts/kong/README.md
pmalek: I've added similar description to the `serviceDiscovery` section and linked it from each relevant table entry. PTAL

charts/kong/example-values/README.md
czeslavo: That should be fine although I agree that "service discovery" still sounds quite wide and could mean many other things while narrowing it down to "gateway discovery" gives a better grasp of what the feature really does (I've just cross-checked that with a DevOps engineer 😅).

rainest (approved):
Intuitive naming and descriptions for completely new features that aren't tied to other known functionality (integrating with a new Kubernetes feature, for example) is always bit difficult, but I think this is in a good place for release.

We should consider supplementing it with a docs guide that walks through SCD/SD concepts from start to finish, since the chart docs are more reference-style, which isn't the best for explaining new concepts.

charts/kong/example-values/README.md
rainest: To echo my other comment, short abstract names for deployment strategies are always a bit iffy, similar to how "hybrid mode" doesn't really capture what hybrid mode actually entails at a technical level.

I think we largely want the naming to be concise and consistent, and then have a longer docs site article explaining the technical details (controller is de-coupled from Kong instances, uses the admin Service to determine where it needs to send config, which has X/Y/Z benefits over existing strategies, etc.) in depth, so that the name serves less a descriptive purpose than it does a recognizable short name associated with a larger set of concepts.

It's at least much less confusing than the old "Kong for Kubernetes"/"Kong for Kubernetes With Kong Enterprise"/etc. mishmash of near-identical names referring to subtle (but important) distinctions in deployment strategy :laughing:

pmalek:
BTW: I guess we need something to block PRs with labels that explicitly tell you not to merge.

Just like we have it in KIC: https://github.com/Kong/kubernetes-ingress-controller/blob/6d44fb04dec78a66860018edb107397c152a769a/.github/workflows/check_pr_labels.yaml

PR #750, opened 2023-02-16
chore(doc) add SCD example and instructions

rainest:
Adapts the instructions from #747 into an example, adds README stuff.


pmalek:
Alright Github, I didn't expect this 🙄

PR #752, opened 2023-02-16
fix: fix autoscaling version detection

pmalek:

Fix autoscaling version detection.

1.26 doesn't offer `autoscaling/v2beta2` anymore (ref: https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-26) so on 1.26 and up incorrectly detects that we don't have v2 available because of `{{- if not (.Capabilities.APIVersions.Has "autoscaling/v2beta2") }}` where in fact we do.

The introduced check, via `contains` checks if the string returned via `"kong.autoscalingVersion"` contains `autoscaling/v2` (which is true for both `autoscaling/v2` and `autoscaling/v2beta2`).

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -1,7 +1,9 @@
 # Changelog
 
-## Unreleased
+## 2.16.5
 
+* Fix autoscaling version detection.
+  [#744](https://github.com/Kong/charts/pull/744)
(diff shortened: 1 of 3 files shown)


czeslavo (dismissed):

pmalek:
Forgot to bump the version in `Chart.yaml`. PTAL @czeslavo

czeslavo (approved):

PR #770, opened 2023-03-17
feat(values) add default security context

rainest:

Sets a default security context that ~disallows running as root~ and makes the root FS read-only.

We haven't set these in the past, but they are common settings to reduce attack surface and are compatible with the standard chart configuration and standard Kong images.

Originally set non-root, which we technically comply with, but since our images only use a named user and not a specific non-0 UID, Kubernetes can't actually verify that the standard `kong` user isn't root (see the initial failed test run), so removed that.

Additional testing that hacks this in via KTF changes: https://github.com/Kong/kubernetes-ingress-controller/actions/runs/4451318454

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## Unreleased
+
+### Improvements
+
+* Security context enforces read-only root filesystem by default. This is not
+  expected to affect most configurations, but [will afffect custom plugins that
+  write to the container filesystem](https://github.com/Kong/charts/blob/main/charts/kong/UPGRADE.md#2170).
+  [#770](https://github.com/Kong/charts/pull/770)
+
 ## 2.18.0
 
 ### Improvements
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -1,11 +1,11 @@
 # Changelog
 
-## Unreleased
+## 2.19.0
 
 ### Improvements
 
 * Security context enforces read-only root filesystem by default. This is not
-  expected to affect most configurations, but [will afffect custom plugins that
+  expected to affect most configurations, but [will affect custom plugins that
   write to the container filesystem](https://github.com/Kong/charts/blob/main/charts/kong/UPGRADE.md#2170).
   [#770](https://github.com/Kong/charts/pull/770)
 
--- a/charts/kong/UPGRADE.md
+++ b/charts/kong/UPGRADE.md
@@ -17,6 +17,7 @@ upgrading from a previous version.
 ## Table of contents
 
 - [Upgrade considerations for all versions](#upgrade-considerations-for-all-versions)
+- [2.17.0](#2170)
 - [2.13.0](#2130)
 - [2.8.0](#280)
 - [2.7.0](#270)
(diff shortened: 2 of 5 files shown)


charts/kong/UPGRADE.md
```diff
@@ -82,6 +83,26 @@ https://raw.githubusercontent.com/Kong/charts/kong-<version>/charts/kong/crds/cu
...
 
+## 2.17.0
+
+2.17 sets a default [security context](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/)
+that declares a read-only root filesystem for Kong containers. The base Kong and KIC
+images are compatible with this setting. The chart mounts temporary writeable
+emptyDir filesystems for locations that require writeable files (`/tmp` and
+`/kong_prefix/`).
+
+This setting limit attack surface and should be compatible with most
```
pmalek, line 94:
```suggestion
This setting limits attack surface and should be compatible with most
```

pmalek (commented):
One comment that I have about this is that I'd much rather see us go into direction of actually bumping majors for these types of changes. It doesn't break all users but it will break some. Hence it should be a major. We shouldn't really be afraid of bumping that. 

Good example of following such practice to me is e.g. https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack. With a nicely written down (in README) history of major bumps that caused a breakage.

Let's not turn semver upside down 🙏

rainest:
Semver is kinda iffy outside strict APIs. Something like a function signature changing is breaking in strict languages, but complex systems aren't that, and we need to make judgement calls re messaging for the latter.

If we want a versioning scheme that denotes all breaking changes in a complex system, that scheme is a single integer that increments on any change, and considers all changes breaking, since it can't otherwise maintain load-bearing bug-compatibility guarantees.

My assessment here was that this is not expected to break the vast majority of configurations, and as such that messaging otherwise would cause more confusion than it would do good. Telling users that this _is_ breaking is confusing, because our instructions for avoiding issues would be "do nothing" in the vast majority of cases. My assessment of why that is follows.

This does not break the default system. We know where we need writeable FS in the default system and account for them already using non-root FS mounts.

I know of few things that _would_ write to the root FS. Those are:
- Overriding the log location defaults from stdout/stderr to a location on the root FS. Logging to disk is unusual for Kubernetes containers, and even for non Kubernetes environments our default locations are in the Kong prefix directory, which is one of the non-root FSes we create for you.
- Plugins that write to disk. Lua _allows_ this, but we don't recommend that you do it for any purpose, and I would recommend against it in favor of alternatives. It is grey area where we do not explicitly forbid it, but it's also something I would never suggest we officially support unless someone can demonstrate cases where it's clearly better than any alternative.

In either of those cases, you can:
- Create and mount a non-root volume. EmptyDirs are free and, as far as I know, provide the same functionality as writing to a writeable root container FS (it's some space allocated on the Kubelet either way).
- Make a single-value change in values.yaml if you really must write to the root FS for some reason.

Again, while we lack the means to assess the actual percentage of impacted users, if it's ballpark 0.0001%, should we make a major version bump here? We've historically reserved those for cases where we know most users will need to make significant changes, and I think that's actually good practice.

pmalek:
Got it and thanks for the elaborate explanation.

My intention was to take out the "judgement" from the equation. We shouldn't use human judgement to gauge if a breaking change is breaking to a big enough audience so that it's worth a major bump.

A breaking change major bump doesn't necessarily need to be a breaking change for **all** users. It will be breaking for those that are affected by its changes.

I'm not going to push my approach to this in this PR but I'm open to discussion on this topic and creating a common understanding and a rule of thumb for that.

---

Having said that: I'm ok merging this as proposed with a twist that now given the 2.17 is out I'd suggest to target this to 2.18.

charts/kong/CHANGELOG.md
```diff
@@ -20,7 +20,11 @@
   [#762](https://github.com/Kong/charts/pull/762)
 * Added support for loadBalancerClass in LoadBalancer type services.
   [#767](https://github.com/Kong/charts/pull/767)
-* Added support for `GRPCRoute`s.
+* Security context enforces read-only root filesystem by default. This is not
+  expected to affect most configurations, but [will afffect custom plugins that
```
czeslavo:
```suggestion
  expected to affect most configurations, but [will affect custom plugins that
```

czeslavo (commented):
I think we should be able to remove `"no-read-only-root-fs"` exclude from `.kube-linter.yaml` with this 🎉

rainest:
Rebased and added an actual version bump to avoid chasing the UPGRADE/CHANGELOG discrepancy again.

czeslavo (approved):

PR #806, opened 2023-05-24
Add opinionated KIC + Gateway Discovery chart

mheap:

Adds a wrapper chart that will deploy a standalone KIC with Gateway Discovery enabled.

This is a brand new chart. I've tested locally with KIC/Konnect and KIC+Gateway without Konnect. Traffic is proxied as expected + 

#### Checklist

### DIFF
--- a/.github/workflows/main-pr.yaml
+++ b/.github/workflows/main-pr.yaml
@@ -49,8 +49,10 @@ jobs:
         with:
           version: v3.11.0
 
-      - name: Add Bitnami
-        run: helm repo add bitnami https://charts.bitnami.com/bitnami
+      - name: Add Helm repos
+        run: |
+          helm repo add bitnami https://charts.bitnami.com/bitnami
+          helm repo add kong https://charts.konghq.com
 
       - uses: actions/setup-python@v4
         with:
@@ -78,7 +80,7 @@ jobs:
         # always is fine
 
       - name: Run chart-testing (install)
-        run: ct install --all
+        run: ct install --charts charts/kong
 
   integration-test:
     runs-on: ubuntu-latest
@@ -97,6 +99,9 @@ jobs:
         - "1.25.9"
         - "1.26.4"
         - "1.27.1"
+        chart-name:
+          - kong
+          - ingress
     steps:
       - name: Checkout
         uses: actions/checkout@v3
@@ -114,9 +119,13 @@ jobs:
         run: ./scripts/test-env.sh
 
       - name: run integration tests (integration)
+        env:
+          CHART_NAME: ${{ matrix.chart-name }}
         run: ./scripts/test-run.sh
 
       - name: run upgrade integration tests (integration-upgrade)
+        env:
+          CHART_NAME: ${{ matrix.chart-name }}
         run: ./scripts/test-upgrade.sh
 
       - name: cleanup integration tests (cleanup)
--- a/charts/ingress/values.yaml
+++ b/charts/ingress/values.yaml
@@ -1,0 +1,37 @@
+deployment:
+  test:
+    enabled: false
+
+controller:
+  proxy:
+    nameOverride: kong-gateway-proxy
+
+  enabled: true
+
+  deployment:
+    kong:
+      enabled: false
+
+  ingressController:
+    enabled: true
+
+    gatewayDiscovery:
+      enabled: true
+      adminApiService:
+        name: kong-gateway-admin
+
+gateway:
+  enabled: true
+  deployment:
+    kong:
+      enabled: true
+
+  admin:
+    enabled: true
+
+  ingressController:
+    enabled: false
+
+  env:
+    role: traditional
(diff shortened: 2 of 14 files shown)


charts/ingress/values.yaml
```diff
@@ -0,0 +1,36 @@
...
+  ingressController:
+    enabled: true
+    image:
+      repository: kong/kubernetes-ingress-controller
+      tag: 2.9.3
+
+    gatewayDiscovery:
+      enabled: true
+      adminApiService:
+        name: kong-proxy-admin
```
mheap: Ideally we'd build this from the deployment name if not provided by the user, but the underlying chart has it marked as required

charts/ingress/values.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+  deployment:
+    kong:
+      enabled: false
+
+  ingressController:
+    enabled: true
+    image:
+      repository: kong/kubernetes-ingress-controller
+      tag: 2.9.3
```
mheap: Maybe I should leave this to the `kong/kong` `values.yaml`

pmalek: Definitely leave this using defaults from the dependent chart.

charts/ingress/values.yaml
```diff
@@ -0,0 +1,36 @@
+controller:
+  proxy:
+    nameOverride: kong-proxy-proxy
```
pmalek: Can we rename this to `kong-proxy` and `adminApiService` to `kong-admin`? Or some other variation of this to prevent the name stuttering?

pmalek (commented):
Ideally we'd add tests similar to what is done for `kong/kong` chart.

This will require some changes to our testing code as it assumes there's only one chart: e.g. https://github.com/Kong/charts/blob/d5ccb3230e7670aba82444151b91551184b8f844/scripts/test-run.sh#L72

charts/ingress/values.yaml
mheap: We could rename the subchart alias to gateway. That would result in `kong-gateway-proxy` (it’s in the form `deployment-chart-service` and I’m reluctant to change the underlying chart)

pmalek: That could work. Let's not change the underlying chart as the reason for this (opinionated chart) conception was that `kong` chart is hard to change in non breaking way for this purpose.

mheap:
I've added some `helm test` tests, but not `ct install` tests as I couldn't figure out how to get the `controller.ingressController.gatewayDiscovery.adminApiService.name` value. I _think_ I need to update the `kong` chart to dynamically set this as a default based on the release name, but that's a bigger change which I'd like to do in a followup PR.

I'm happy for this to be merged as we have `helm test` validation

scripts/test-run.sh
pmalek, line 1: I like the changes here and I'm glad we can make this work for now. I'm a bit worried that adding more opinionated charts here might make this difficult to reason about. Not a blocker for now but definitely something that can think about adding more charts here

scripts/test-run-ingress.sh
pmalek, line 1: I can't see any call sites of this script. Wasn't that the initial version which then got replaced with `test-run.sh` relying on `CHART_NAME` env?

.github/workflows/main-pr.yaml
```diff
@@ -78,7 +80,7 @@ jobs:
         # always is fine
 
       - name: Run chart-testing (install)
-        run: ct install --all
+        run: ct install --charts charts/kong
```
pmalek, line 83: What is the reason for specifying only the `kong` chart here 🤔 ?

I can see `.Values.deployment.test.enabled` dependent manifests in `ingress` chart as well but OTOH I can see those used in the `test-run.sh` so this has to be used in a different way. It's been a while I looked at `ct` testing.

scripts/test-run-ingress.sh
mheap: This was an old file from when I copy/pasted to customise the test run. Removing

.github/workflows/main-pr.yaml
mheap: The `.Values.deployment.test.enabled` check is for deploying resources used by `helm test` rather than `ct install`.

The `ct install` tests run the chart for each values file in `charts/<name>/ci`. Reading through I'm not sure how they work, or if I'll be able to add one for `ingress` without changing the underlying `kong` chart as `adminApiService.name` currently needs to be known in advance.

Given we have `helm test` tests for the new `ingress` chart I'd like to merge this and we can revisit `ct install` when @rainest returns

scripts/test-run.sh
mheap: I've tidied up `test-run.sh` a bit more. Understood on the complexity as we add more charts. We can generalise a bit more once we add one more chart as it then becomes more complex than an `if/else`

pmalek (approved):

PR #808, opened 2023-05-25
fix(release): Fix release workflow

mheap:

The existing release workflow failed with the following error:

```
Run ./scripts/test-run.sh
./scripts/test-run.sh: line 36: CHART_NAME: unbound variable
Error: Process completed with exit code 1.
```

This PR sets `CHART_NAME` as expected in `main-push.yaml`

#### Checklist

### DIFF
--- a/.github/workflows/main-push.yaml
+++ b/.github/workflows/main-push.yaml
@@ -8,6 +8,11 @@ on:
 jobs:
   lint-test:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        chart-name:
+          - kong
+          - ingress
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/main-push.yaml
```diff
@@ -27,17 +32,20 @@ jobs:
         uses: helm/chart-testing-action@v2.4.0
 
       - name: Run chart-testing (lint)
-        run: ct lint
+        run: ct lint --charts ${{ matrix.chart-name }}
```
pmalek, line 35: This will pass for `ingress` given that it doesn't have any ct related manifests?

pmalek (approved):

.github/workflows/main-push.yaml
mheap: It does. `ct lint` and `ct install` are different tasks. `ct install` won't pass [which is why it's got an `if:` block](https://github.com/Kong/charts/pull/808/files#diff-f6084f2529e142d031439a61076ecb3f1e30b47b81c812937178e872f9fb8331R44)

PR #813, opened 2023-05-30
feat(conf): support multiple hosts and tls certificates when kong is…

bergemalm:
… exposed via ingress

An Ingress support multiple hosts and it's common to use CNAME using "simpler" dns records pointing to the same ingress. We need to support this.

#### Checklist


czeslavo:
@bergemalm Thank you for the contribution. Have you considered making this change non-breaking by keeping the old keys and merging paths and TLS config from both? 

As I can see the only clash we'd have would be the `tls` key name, but we could dispatch how we handle it by its type using helm's [`kindIs`](https://docs.helm.sh/docs/chart_template_guide/function_list/#kind-functions) (e.g. `kindIs "slice" .tls` / `kindIs "string" .tls`).

bergemalm:
Ok, if that is preferred I can have a look at it @czeslavo .

bergemalm:
Hi again @czeslavo, how about solving it like what's in this PR now? (edited the description above removing the breaking change related text)
It's possible to mix and combine. I also added a requirement for actually setting a hostname. The host can't be empty in an ingress configuration.

charts/kong/templates/_helpers.tpl
```diff
@@ -116,12 +120,49 @@ spec:
...
+              name: {{ $.fullName }}-{{ $.serviceName }}
+              port:
+                number: {{ $servicePort }}
+            {{- end }}
+          {{- range .paths }}
+          path: {{ .path }}
+          {{- if (not (eq $.ingressVersion "extensions/v1beta1")) }}
+          pathType: {{ .pathType }}
+          {{- end }}
+          {{- end }}
```
czeslavo: Shouldn't we iterate over `{{- range .paths }}` not here, but rather just before `http.paths` to generate one path per `hosts[].paths` entry?

czeslavo:
Hey @bergemalm, thank you for adjusting the PR 🙂 I think we're on a good track, but I still find a couple of issues we may have here:

1. I think that we should keep the behavior that allowed using just `proxy.ingress.enabled` without `proxy.ingress.hostname` specified:
```yaml
proxy:
  ingress:
    enabled: true
```

Before the changes, it was generating a valid Ingress (it has an empty `host` which is valid for K8s API): 

```yaml
# Source: kong/templates/service-kong-proxy.yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: kong-ingress-controller-kong-proxy
  namespace: kong
  labels:
    app.kubernetes.io/name: kong
    helm.sh/chart: kong-2.22.0
    app.kubernetes.io/instance: "kong-ingress-controller"
    app.kubernetes.io/managed-by: "Helm"
    app.kubernetes.io/version: "3.2"
spec:
  rules:
  - host:
    http:
      paths:
        - backend:
            service:
              name: kong-ingress-controller-kong-proxy
              port:
                number: 443
          path: /
          pathType: ImplementationSpecific
```

After the changes, it generates no rules:
```yaml
# Source: kong/templates/service-kong-proxy.yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: kong-ingress-controller-kong-proxy
  namespace: kong
  labels:
    app.kubernetes.io/name: kong
    helm.sh/chart: kong-2.22.0
    app.kubernetes.io/instance: "kong-ingress-controller"
    app.kubernetes.io/managed-by: "Helm"
    app.kubernetes.io/version: "3.2"
spec:
  rules:
```

I'd suggest not requiring the `proxy.ingress.hostname` to be specified and in case there are no `proxy.ingress.hostname` or `proxy.ingress.hosts`, generate the default rule with no `host` (as in the output above).

2. We don't have any `charts/kong/ci/*-values.yaml` that would include both `proxy.ingress.hostname` and `proxy.ingress.hosts` to ensure they do not conflict and render correctly.

3. We do not have `charts/kong/ci/*-values.yaml`s that would include `proxy.ingress.tls` as a slice and another one as a string to verify they both render properly.

For 2 and 3 I suggest adding [the following](https://github.com/Kong/charts/files/11653332/values.patch) `*-values.yaml` files to the `charts/kong/ci` directory so they're linted in the CI. It's also worth running `helm install --dry-run kong ./charts/kong -n kong --values ./charts/kong/ci/<name>-values.yaml` against each of them to manually verify that everything renders as expected. Unfortunately, we have no automatic testing machinery in place that would allow specifying an expected output somehow for a given test case so a manual check is required. 😞

bergemalm:
Oh, I've never used an empty host and assumed it was an error. Sorry, I'll have another look then...

charts/kong/templates/_helpers.tpl
bergemalm: Correct. Will fix.

bergemalm:
I have made changes now @czeslavo, sorry for the delay. I used your patch for tests but with some modifications, for example the "combined" hostname and hosts test since there would otherwise be a duplication of the `/` path.
What do you think?

bergemalm:
I guess we would need a tls secret... Or multiple actually. Keep those in tests or not or how to solve for it @czeslavo ?

bergemalm:
I see now that there are some certificate generation/cert-manager support on other places. The question is if we want similar support for the ingress?

bergemalm:
Perhaps skip adding ingress tls in tests for now and get this merged? More comments? Ping @czeslavo

charts/kong/templates/_helpers.tpl
```diff
@@ -116,12 +117,74 @@ spec:
...
+            serviceName: {{ .fullName }}-{{ .serviceName }}
+            servicePort: {{ $servicePort }}
+          {{- else }}
+            service:
+              name: {{ .fullName }}-{{ .serviceName }}
+              port:
+                number: {{ $servicePort }}
+            {{- end }}
+          path: {{ $path }}
+          {{- if (not (eq .ingressVersion "extensions/v1beta1")) }}
```
pmalek: You might want to drop the v1beta1 branch here since support for v1beta1 ingresses has been dropped in KIC 2.10: https://github.com/Kong/kubernetes-ingress-controller/blob/v2.10.0/CHANGELOG.md#deprecated

charts/kong/templates/_helpers.tpl
```diff
@@ -116,12 +117,74 @@ spec:
...
+              name: {{ $.fullName }}-{{ $.serviceName }}
+              port:
+                number: {{ $servicePort }}
+          {{- end }}
+          {{- end }}
+          {{- if (and $hostname (and (eq $path .path))) }}
+          {{- fail "duplication of specified ingress path" }}
+          {{- end }}
+          path: {{ .path }}
+          {{- if (not (eq $.ingressVersion "extensions/v1beta1")) }}
```
pmalek: Ditto

charts/kong/templates/_helpers.tpl
```diff
@@ -116,12 +117,74 @@ spec:
...
+        - backend:
+          {{- if (not (eq .ingressVersion "networking.k8s.io/v1")) }}
+            serviceName: {{ .fullName }}-{{ .serviceName }}
+            servicePort: {{ $servicePort }}
+          {{- else }}
+            service:
+              name: {{ .fullName }}-{{ .serviceName }}
+              port:
+                number: {{ $servicePort }}
+            {{- end }}
```
pmalek: This seems to not be indented properly, right?


```suggestion
          {{- end }}
```

pmalek:
> I guess we would need a tls secret... Or multiple actually. Keep those in tests or not or how to solve for it @czeslavo ?

It's perfectly fine to define the secrets in test ( to rely on concrete, the same values across test runs). Since that's a non trivial change I'd suggest we try to pursue the route of adding some basic tests for this change.

charts/kong/templates/_helpers.tpl
bergemalm: Sure, I can remove the ingress version related things.

bergemalm: Done.

charts/kong/templates/_helpers.tpl
bergemalm: Done.

bergemalm:
> > I guess we would need a tls secret... Or multiple actually. Keep those in tests or not or how to solve for it @czeslavo ?
> 
> It's perfectly fine to define the secrets in test ( to rely on concrete, the same values across test runs). Since that's a non trivial change I'd suggest we try to pursue the route of adding some basic tests for this change.

Ok, so could you point me in the direction of where to create and apply the tls secret(s) before the helm installs?

czeslavo:
@bergemalm You can leverage the `extraObjects` and just append the below to every `*-values.yaml` you've added:
```yaml
extraObjects:
  - apiVersion: v1
    kind: Secret
    metadata:
      name: kong.proxy.example.secret
```

bergemalm:
Changes has been pushed, checks are passing. Ready for another review.

charts/kong/templates/_helpers.tpl
```diff
@@ -95,33 +95,75 @@ metadata:
     {{- end }}
   {{- end }}
 spec:
-{{- if (and (not (eq .ingressVersion "extensions/v1beta1")) .ingress.ingressClassName) }}
+{{- if .ingress.ingressClassName }}
   ingressClassName: {{ .ingress.ingressClassName }}
 {{- end }}
   rules:
+  {{- if ( not (or $hostname .ingress.hosts)) }}
+  - host: {{ $hostname | quote }}
```
czeslavo: Nit: as here we already know that `$hostname` is empty, we might as well omit defining `host` field.

czeslavo (commented):
Beside one tiny nit I commented, LGTM now. 👍

charts/kong/templates/_helpers.tpl
bergemalm: Ofc! Will fix...

bergemalm:
Updated now @czeslavo.

czeslavo (approved):
✅

bergemalm:
I cannot merge, please do when appropriate.

PR #823, opened 2023-06-20
feat(kong): fail to render templates when PodSecurityPolicy is requested but cluster doesn't serve it

pmalek:

When PSP is requested via `.Values.podSecurityPolicy` but cluster cannot serve it then fail to render instead of silently ignoring.

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 ### Improvements
+
 * Running `tpl` against user-supplied labels and annotations used in Deployment
   #### example:
   ```yaml
@@ -10,6 +11,9 @@
     version: "{{ .Values.image.tag }}"  # Will render dynamically when overridden downstream
   ```
   [#814](https://github.com/Kong/charts/pull/814)
+* Fail to render templates when PodSecurityPolicy was requested but cluster doesn't
+  serve its API.
+  [#823](https://github.com/Kong/charts/pull/823)
 
 ## 2.23.0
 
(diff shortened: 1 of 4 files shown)


rainest:
We don't have a way to make CI expect failures like these on cluster versions where we expect them, correct? Do we simply have to remove coverage for this to make chart-testing not complain "hey! this failed to install!" (as it should)?

pmalek:
> We don't have a way to make CI expect failures like these on cluster versions where we expect them, correct? Do we simply have to remove coverage for this to make chart-testing not complain "hey! this failed to install!" (as it should)?

There might be a way but I don't know it off the top of my head

rainest (changes requested):
I don't see anything in chart-testing that suggests it would, so go ahead and just strip the PSP setting out of ci/test1-values.yaml

Realistically, PSP hasn't changed in a while, isn't likely to change going forward (since it no longer exists in current versions), and isn't likely to be broken by changes elsewhere (or rather, if we break something like the namespace or SA templates, that will break other resources also).

pmalek:
> I don't see anything in chart-testing that suggests it would, so go ahead and just strip the PSP setting out of ci/test1-values.yaml
> 
> Realistically, PSP hasn't changed in a while, isn't likely to change going forward (since it no longer exists in current versions), and isn't likely to be broken by changes elsewhere (or rather, if we break something like the namespace or SA templates, that will break other resources also).

👍 Done.

It seems that if we ever wanted to do something like this we'd need to provide a matrix similarly as [it's done for "integration" tests ](https://github.com/Kong/charts/blob/21ec72e9e2b716dd119bc093dba36a403334191a/.github/workflows/main-pr.yaml#L85-L104) and somehow map test values against those versions.

rainest (approved):

Issue #587, opened 2022-04-26
IngressClass is not deployed with ArgoCD


spoukke:
When I deploy kong with ArgoCD using the chart, the IngressClass is not deployed by default.

I think this is because of [this line](https://github.com/Kong/charts/blob/main/charts/kong/templates/ingress-class.yaml#L7). Indeed, since ArgoCD does not do any helm release, this annotation is never added.

It would be nice to change this condition with another annotation, which will exist whether or not the chart is installed through a Helm release.

For information, I use the 2.7.0 version.

rainest:
Does ArgoCD provide any means to add arguments to the Helm executable? I suspect what you actually need is this:

```
--api-versions "networking.k8s.io/v1/Ingress" `--api-versions "networking.k8s.io/v1/IngressClass"
```

The IngressClass template has two types of conditionals, one set to [determine whether the cluster supports IngressClass at all](https://github.com/Kong/charts/blob/b3e6644a27824d5b177caf515aabc7f2e96c7777/charts/kong/templates/ingress-class.yaml#L3-L4) and another to [check if there's already an IngressClass in place](https://github.com/Kong/charts/blob/b3e6644a27824d5b177caf515aabc7f2e96c7777/charts/kong/templates/ingress-class.yaml#L5-L7).

If the resource isn't present _at all_ it should fall into the [final else condition](https://github.com/Kong/charts/blob/b3e6644a27824d5b177caf515aabc7f2e96c7777/charts/kong/templates/ingress-class.yaml#L14), which does create it. That should work on the initial install--the resource lookup/name comparison should only matter for upgrades of existing installs. However, `helm template` has a wide variety of limitations, one of which being that it doesn't consult the cluster to determine the API versions available, which will make our chart fall back to using an older version of Ingress prior to the introduction of IngressClass.

https://github.com/argoproj/argo-cd/issues/3631 indicates that Argo should set those automatically, but that there's apparently some issue with Helm 2 charts. Upgrading to 2.8 may fix this, since we just recently converted to the Helm 3 chart format.

I'm not quite sure what would happen after on upgrades--it depends on how ArgoCD is handling the release management/resource deletion/etc. that'd normally be handled by Helm when using `install` instead of `template`. If `template` can't perform the lookup (which I think is the case), it should just treat this as a fresh install with no existing IngressClass and continue to render it.

PR #827, opened 2023-06-28
Fix Ingress and HPA API versions during capabilities checking

hnajib-sym:

Helm templating for Ingress and HPA since apiversions ending with /Ingress and /HorizontalPodAutoscaler doesn't exist 

  - fixes: 

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -15,7 +15,7 @@
   serve its API.
   [#823](https://github.com/Kong/charts/pull/823)
 * Fix Ingress and HPA API versions during capabilities checking 
-  [#823](https://github.com/Kong/charts/pull/827)
+  [#827](https://github.com/Kong/charts/pull/827)
 
 ## 2.23.0
 
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -14,6 +14,8 @@
 * Fail to render templates when PodSecurityPolicy was requested but cluster doesn't
   serve its API.
   [#823](https://github.com/Kong/charts/pull/823)
+* Fix Ingress and HPA API versions during capabilities checking 
+  [#823](https://github.com/Kong/charts/pull/827)
 
 ## 2.23.0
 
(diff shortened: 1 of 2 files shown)


rainest (commented):
Did you see an issue with the current templates' output? As best I recall we did conduct tests against older versions when we introduced this and got the expected output.

The [Helm documentation](https://helm.sh/docs/chart_template_guide/function_list/#capabilitiesapiversionshas) for this built-in indicates that it _does_ support including the Kind.

hnajib-sym:
@rainest Indeed , for ingress we had to add         --api-versions networking.k8s.io/v1/Ingress to helm template, else it was rendering extensions/v1beta1 for the ingress despite having networking.k8s.io/v1 capabilities , removing /Ingress solved the issue

This block always fall to extensions/v1beta1

{{- define "kong.ingressVersion" -}} 
 {{- if (.Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress") -}} 
 networking.k8s.io/v1 
 {{- else if (.Capabilities.APIVersions.Has "networking.k8s.io/v1beta1/Ingress") -}} 
 networking.k8s.io/v1beta1 
 {{- else -}} 
 extensions/v1beta1 
 {{- end -}} 
 {{- end -}}


Same goes for HPA.

Tested against embedded helm in argocd 2.4.15 as well 2.7.6 ( helm 3.11)

rainest:
To clarify, I think you're saying that Argo automatically populates capability info, but only does the API versions, not the individual resources? That would explain why we got a [similar report](https://github.com/Kong/charts/issues/587#issuecomment-1115288514) despite upstream indicating it did set those when it runs `helm template`.

hnajib-sym:
You can ignore Argocd for now , doing a helm template . while listening all the capabilities will list networking.k8s.io/v1 as one of the capabilities, but the condition  {{- if (.Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress") -}}  templates/_helpers.tpl fails , removing /Ingress resource fix the issue and the condition succeed.

rainest (approved):

PR #833, opened 2023-07-06
fix: re-add missing kong.ingressVersion template

pmalek:

This adds `kong.ingressVersion` template which was removed in #813.

Additionally it adds a tests that uses that ( by exposing Admin API service ).

Without this patch the added test values.yaml file, yields:

```
Error: template: kong/templates/service-kong-admin.yaml:5:47: executing "kong/templates/service-kong-admin.yaml" at <include "kong.ingressVersion" .>: error calling include: template: no template "kong.ingressVersion" associated with template "gotpl"
```

### DIFF
--- a/charts/kong/ci/admin-api-service-clusterip.yaml
+++ b/charts/kong/ci/admin-api-service-clusterip.yaml
@@ -1,0 +1,6 @@
+admin:
+  enabled: true
+  type: ClusterIP
+
+ingressController:
+  enabled: false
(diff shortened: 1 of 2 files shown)


czeslavo:
Just for context: why do we need to bring this back?

Edit: Ah, we removed the function while it was still used in several places, right?

pmalek:
> Edit: Ah, we removed the function while it was still used in several places, right?

yes

pmalek:
To verify that this breaks indeed you can try using 

```yaml
admin:
  enabled: true
  type: ClusterIP

ingressController:
  enabled: false
```

as values for current dev version of the chart.

czeslavo (approved):
🎖️ for catching this before the release.

PR #834, opened 2023-07-06
chore: bump postgres to 13.11

pmalek:

This PR bumps default default postgres tag to 13.11.

Why? To include most recent 13.X fixes and most importantly to by default use a version that has arm64 images: https://hub.docker.com/layers/bitnami/postgresql/13.11.0-debian-11-r20/images/sha256-2f4e684e911831a62cdabc214298b1dcb9a1c797bfdb8993dc28350c79b8a637?context=explore.

The reason for this is to allow people running on arm64 to have a working postgres setup with default values and also, transitively to have that in ktf where we don't (yet!) have an option to specify this.

Also release 2.24 chart version.

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -1,23 +1,30 @@
 # Changelog
 
-## Unreleased
+## 2.24.0
 
 ### Improvements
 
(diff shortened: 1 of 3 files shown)


pmalek:
Wait for  #833 to include it in the release

programmer04 (approved):

Issue #836, opened 2023-07-07
Kong pod is getting crashloopback after abrupt shutdown 


soorajsomans:
Im using kong gateway in a single node kubernetes cluster. The kong pod is getting CrashLoopBackOff after a abrupt shutdown, 
eg. container killed forcefully or system shutdown due to power failure or system crash
On the next system start kubernets brings the pod but the state is always CrashLoopBackoff. The below error where shown in the proxy container logs

nginx: [emerg] bind() to unix:/kong_prefix/worker_events.sock failed (98: Address in use)
[notice] 1#0: try again to bind() after 500ms

In the kong container /kong_prefix/worker_events.sock file is mounted using empty dir and this not get cleaned up during abrupt shutdown. 

I have solved this issue by cleaning this file in clear-stale-pid init container. I have updated the command in like below

Please find the below pull request for the solution i have tried
https://github.com/Kong/charts/pull/837

Please suggest me if there is a better approach to handle this problem. Can you please address this issue in your upcoming releases?

rainest:
Which Kong images are you using? This looks like something we addressed in the image earlier: https://github.com/Kong/docker-kong/pull/620

You would need Kong image 3.2.2+ for that fix.

soorajsomans:
I'm using kong 3.0 image. I can't update the image right now,  Is there any issue with the above mentioned fix?. I have tested multiple times and its working. Please suggest if any changes required or any possible errors with this fix. Thanks in advance.

rainest:
We'd prefer to have only the one fix rather than multiple overlapping ones. The chart-level workaround should go away once versions older than 3.2 are EOL (after the end of 2023).

https://github.com/Kong/charts/pull/837#pullrequestreview-1533334571 for changes.

soorajsomans:
Thank you for your support. I'm planning for the image upgrade.

PR #837, opened 2023-07-07
Cleaning $KONG_PREFIX/worker_events.sock file

soorajsomans:
Fix for Kong pod is getting crashloopback after abrupt shutdown #836

Kong pod is getting crashloopback after abrupt shutdown #836 

#### Checklist

### DIFF
--- a/charts/kong/templates/deployment.yaml
+++ b/charts/kong/templates/deployment.yaml
@@ -93,9 +93,14 @@ spec:
         resources:
 {{ toYaml .Values.resources | indent 10 }}
         command:
-        - "rm"
-        - "-vrf"
-        - "$KONG_PREFIX/pids"
+        - sh
+        - -c
+        - 'if test -S $KONG_PREFIX/worker_events.sock; then
+              rm $KONG_PREFIX/worker_events.sock;
+              rm -vrf $KONG_PREFIX/pids;
+           else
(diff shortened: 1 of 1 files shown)


charts/kong/templates/deployment.yaml
```diff
@@ -93,9 +93,14 @@ spec:
         resources:
 {{ toYaml .Values.resources | indent 10 }}
         command:
-        - "rm"
-        - "-vrf"
-        - "$KONG_PREFIX/pids"
+        - sh
+        - -c
+        - 'if test -S $KONG_PREFIX/worker_events.sock; then
```
rainest, line 98: Is there a reason to use the `test -S` rather than having this also use an `rm -vf` to not fail if the file isn't present? It would technically remove non-socket files also, but there's no reason that this path should ever be anything else.

Removing the if case and running the same set of commands always would make this simpler.

rainest (changes requested):
This needs a changelog entry as well as the script simplification.

charts/kong/templates/deployment.yaml
soorajsomans: I had tried that initially but it was failing (maybe because of some other issue in my system). Since this issue has already addressed, I'm planning to upgrade the images. I'm closing this request, Thank You for your support.

PR #839, opened 2023-07-14
feat(kong): Generate Admin API service name if requested

mheap:

Adds a `generateAdminApiService` option to use `.Release.name` rather than hardcoding to `kong`. We specify the release name as `kong` in demos, but don't always make it clear that there's an implicit dependency here. Instead if documenting the implicitness, let's generate an accurate service name in the chart.

N/A

#### Checklist
- [x] Changes are documented under the "Unreleased" header in CHANGELOG.md <- Added as 0.4.0 as when the PR is merged it'll auto-release

### DIFF
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -805,8 +805,8 @@ You'll be able to configure this feature through configuration section under
   - `ingressController.gatewayDiscovery.adminApiService.name`
   - `ingressController.gatewayDiscovery.adminApiService.namespace`
 
-  If you set `ingressController.gatewayDiscovery.generateAdminApiService` to `true`
-  and the chart will generate values for `name` and `namespace` based on the current release name and
+  If you set `ingressController.gatewayDiscovery.generateAdminApiService` to `true`,
+  the chart will generate values for `name` and `namespace` based on the current release name and
   namespace. This is useful when consuming the `kong` chart as a subchart.
 
 Using this feature requires a split release installation of Gateways and Ingress Controller.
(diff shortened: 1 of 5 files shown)


charts/ingress/CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## 0.4.0
+
+### Improvements
+
+- Generate the `adminApiService.name` value from `.Release.Name` rather than 
+  hardcoding to `kong`
+  [#839](https://github.com/Kong/charts/pull/839)
```
pmalek: This would also need to be mentioned in `kong` chart right? And that would need to be bumped as well.

I believe a better course of action for this would be to first introduce this in `kong` get that released and only after use it in `ingress` chart. WDYT?

mheap: This PR changes the `kong` chart (I forgot to bump the version, doing so now) _and_ updates the `values.yaml` default in `ingress`. Both should go out together

mheap: Ah, once `kong` is released we need to update `Chart.lock` in `Ingress`. I'll split the work

pmalek (dismissed):

charts/kong/README.md
```diff
@@ -798,12 +799,16 @@ You'll be able to configure this feature through configuration section under
...
-  The following admin API Service flags have to be provided in order for gateway
+  The following admin API Service flags have to be present in order for gateway
   discovery to work:
 
   - `ingressController.gatewayDiscovery.adminApiService.name`
   - `ingressController.gatewayDiscovery.adminApiService.namespace`
 
+  If you set `ingressController.gatewayDiscovery.generateAdminApiService` to `true`
+  and the chart will generate values for `name` and `namespace` based on the current release name and
+  namespace. This is useful when consuming the `kong` chart as a subchart.
```
czeslavo:
```suggestion
  If you set `ingressController.gatewayDiscovery.generateAdminApiService` to `true`,
  the chart will generate values for `name` and `namespace` based on the current release name and
  namespace. This is useful when consuming the `kong` chart as a subchart.
```

czeslavo (dismissed):

czeslavo (approved):

PR #843, opened 2023-07-19
fix(ci) update test to agree with new defaults

rainest:
Do not set the admin API Service name for tests against the ingress chart.

5802d40a78eb9fb6eabcdf9acddf4c3f8e81a1fd added a new default for gatewayDiscovery.generateAdminApiService. This setting is incompatible with specifying an admin Service name. The test was originally setting that name and failing, and the commit was merged to main despite in https://github.com/Kong/charts/pull/840.


pmalek (commented):
We could remove this `if` statement altogether but this will also do for now 👍

pmalek (approved):

PR #844, opened 2023-07-19
Update the readiness path default with backwards compatibility

### LINKED ISSUE #830: Use `/status/ready` as Gateway's containers readiness probe
### Context 

As Kong 3.3 has introduced a new `/status/ready` status endpoint that returns 200 only when a Gateway is configured with a valid configuration, we'd like to use it as a default readiness probe of Gateway containers. 

### Proposed solution

1. Change the default path of the readiness probe from `/status` to `/status/ready` in [values.yaml](https://github.com/Kong/charts/blob/9d1cb442b82b2198f2b30d78e157d3c26abb2515/charts/kong/values.yaml#L774).
2. In case a Kong image version is < 3.3 and the `readinessProbe.httpGet.path == /status/ready`, fall back to the old `/status`. Otherwise, accept whatever the user had supplied. 

### Acceptance criteria
1. When Kong image version >= 3.3 is used, the default readiness probe used for Gateway container is `/status/ready`.
2. When Kong image version < 3.3, the default readiness probe used for Gateway container is `/status`.
3. None of the existing configurations are broken due to the change: 
   - Kong+Ingress Controller in a single Pod
   - Kong with no Ingress Controller (just using `/status/ready` might break it as the Admin API service will never get ready and `deck` won't be able to configure the Gateway, which may require special handling)
   - Kong and Ingress controller in separate Pods (aka Gateway Discovery mode - KIC has to be able to discover Gateways despite them being not ready, and configure them: see https://github.com/Kong/kubernetes-ingress-controller/issues/3499)

rainest:
Use the 3.3 readiness endpoint by default, unless the image versions are not compatible with it. While the readinessProbe is still fully configurable in values.yaml, the template will edit it if necessary.

Fix #830 

This needs the nightly KIC commit removed before merging. It's in place for now to demonstrate behavior on that version. test1 still uses 2.10 to validate the backwards-compatible path.

We can release this prior to 2.11 as it will not use the new default unless it's compatible.

This includes a copy of https://github.com/Kong/charts/pull/843. If we merge that separately we need to rebase this.

#### Checklist


charts/kong/templates/_helpers.tpl
```diff
@@ -1600,3 +1600,13 @@ networking.k8s.io/v1beta1
 extensions/v1beta1
 {{- end -}}
 {{- end -}}
+
+{{- define "kong.proxy.compatibleReadiness" -}}
+{{- $proxyReadiness := .Values.readinessProbe -}}
+{{- if (or (semverCompare "< 3.3.0" (include "kong.effectiveVersion" .Values.image)) (semverCompare "< 2.11.0" (include "kong.effectiveVersion" .Values.ingressController.image))) -}}
```
czeslavo: I think we should check KIC's version only if `.Values.ingressController.enabled` to make it work when the controller is not used, shouldn't we?
```suggestion
{{- if (or (semverCompare "< 3.3.0" (include "kong.effectiveVersion" .Values.image)) (and .Values.ingressController.enabled (semverCompare "< 2.11.0" (include "kong.effectiveVersion" .Values.ingressController.image)))) -}}
```

charts/kong/values.yaml
```diff
@@ -513,14 +513,14 @@ dblessConfig:
...
-    tag: "2.10"
+    repository: kong/nightly-ingress-controller
+    tag: "2023-07-18"
     # Optionally set a semantic version for version-gated features. This can normally
     # be left unset. You only need to set this if your tag is not a semver string,
     # such as when you are using a "next" tag. Set this to the effective semantic
     # version of your tag: for example if using a "next" image for an unreleased 3.1.0
     # version, set this to "3.1.0".
-    effectiveSemver:
+    effectiveSemver: 2.11.0
```
czeslavo: If we discard these changes we could merge it already, even before KIC 2.11 release, right?

rainest: Yeah, that was why I'd marked 58024d5888bc7c5c4595d918e588057fb86ba700 `tmp` and said:

> This needs the nightly KIC commit removed before merging. It's in place for now to demonstrate behavior on that version. test1 still uses 2.10 to validate the backwards-compatible path.

IDK a great way in GitHub to handle stuff that needs future changes elsewhere to properly demonstrate/test for review without some manual commit juggling. Have removed it and the hold label. https://github.com/Kong/charts/actions/runs/5684355422 is the latest run that checked both new and old with that temp commit.

czeslavo (approved):

PR #851, opened 2023-08-04
feat(migrations) add Argo CD lifecycle policies

### LINKED ISSUE #763: Support Argocd
We are deploying Kong using ArgoCD.
It's working ok but the only problem is with the migration jobs.
Since they do not get automatically deleted, the next time that I upgrade the chart, ArgoCD fails because the job fields are immutable

I tried to configure already ArgoCD  hooks: argocd.argoproj.io/hook: PreSync and   argocd.argoproj.io/hook-delete-policy: HookSucceeded but I had to revert it because I had to add it for all the dependency and as well on my custom resources deployed alongside Kong helm chart.
So far I prefer to keep the helm chart as it is and delete the job manually before upgrading it.
It would be great if somehow we could reshape the helm chart in a way to solve this issue.

Here an example of the logs (I upgraded the chart from 2.14 to 2.16)

`
Failed sync attempt to dad5bdcdd039d6e8585fe55297453aad4fd650d8: one or more objects failed to apply, reason: Job.batch "controlplane-dev-kong-init-migrations" is invalid: spec.template: Invalid value: ...: field is immutable (retried 5 times).`

rainest:

Set both the init-migrations and pre-upgrade-migrations Jobs to the Argo CD Sync policy. This appears to be the best option for working around the lack of distinct install and upgrade lifecycle phases in Argo CD.

Add a blurb to docs about Argo-specific chart features.

  - fixes #763

I have [some notes and test resources](https://gist.github.com/rainest/ee9c75c901de6f101ffd171cd5fa35d9) I used with a KTF build that includes https://github.com/Kong/kubernetes-testing-framework/commit/c097dbd2008cb5fea7c6269ca7ddb3af0aa003ec (not released yet) via `./build/ktf.linux.amd64 env create --addon argocd --addon metallb`

The policies in this PR successfully deployed and upgraded a database instance 

The short version is:

PreSync is more appropriate for upgrade migrations, but in Argo land this is functionally equivalent to pre-install (first install has no special status) and comes with its usual chicken before egg problems, where everything fails because the Job relies on resources that won't exist until the main sync phase.

Sync for `migrations up` may cause the main Deployment to enter crashloop if the migrations take a very long time, but Kubernetes can sort that out. It will eventually recover.

A minor edge case temporary failure feels preferable to the alternatives:

- We use documentation to instruct users to manually toggle jobs on or off, or add Argo policies to them. This will avoid potential temporary backoff, but requires that users read docs and manually apply configuration, which is unreliable.
- We apply sync wave values to all resources, so that Argo will first apply prereq resources, then run migrations, then start the main Deployment. This may avoid temporary bad states (I didn't try it and can't say for certain if it won't come with issues of its own) but requires a lot more work.

Forcibly applying the override Argo hook policies in templates seems like a good enough solution that doesn't require additional user configuration, so I'd like to try it in the wild and see if there are unexpected problems. If there are, we can work on a more complex solution.

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -7,6 +7,8 @@
 * Use the Kong 3.3 `/status/ready` endpoint for readiness probes by default if
   available. If not available, use the old `/status` default.
   [#844](https://github.com/Kong/charts/pull/844)
+* Add ArgoCD `Sync` and `BeforeHookCreation` [hook policies](https://argo-cd.readthedocs.io/en/stable/user-guide/resource_hooks/)
+  to the the init and pre-upgrade migrations Jobs.
 
 ## 2.25.0
 
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -71,6 +71,7 @@ $ helm install kong/kong --generate-name
   - [Sessions](#sessions)
   - [Email/SMTP](#emailsmtp)
 - [Prometheus Operator integration](#prometheus-operator-integration)
+- [Argo CD considerations](#argo-cd-considerations)
 - [Changelog](https://github.com/Kong/charts/blob/main/charts/kong/CHANGELOG.md)
 - [Upgrading](https://github.com/Kong/charts/blob/main/charts/kong/UPGRADE.md)
 - [Seeking help](#seeking-help)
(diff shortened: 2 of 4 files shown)


czeslavo:
Maybe not strictly related to the problem that is being solved in this PR (adding explicit Argo hooks makes sense as the default translation of Helm's `pre-upgrade` to Argo's `PreSync` won't work for us as it's described), I have a general question regarding our approach to migration jobs: is there any blocking reason why we cannot merge both init and pre-upgrade migrations into one? As I recall `kong migrations bootstrap` running before every `kong migrations up` would at worst be a no-op. That would at least leave us with no problem with the special case of the pre-install that is not supported by Argo at all.

czeslavo (approved):

PR #853, opened 2023-08-08
feat(ports): add extra ports feature in kong chart

pereiratps:

This PR adds the possibility of including new ports for the proxy and mapping it to the service.
We need to be able to expose another port to the proxy with different parameter settings, for example, one for http1.1 and another for http2.

It's the first time I contribute, if there's something wrong with the patterns, I'm happy if you can help me.

#### Checklist

### DIFF
--- a/charts/kong/Chart.yaml
+++ b/charts/kong/Chart.yaml
@@ -11,7 +11,7 @@ maintainers:
 name: kong
 sources:
 - https://github.com/Kong/charts/tree/main/charts/kong
-version: 2.25.0
+version: 2.26.0
 appVersion: "3.3"
 dependencies:
 - name: postgresql
--- a/charts/kong/README.md
+++ b/charts/kong/README.md
@@ -670,6 +670,12 @@ or `ingress` sections, as it is used only for stream listens.
 | SVC.tls.hostPort                   | Host port to use for TLS                                                              |                          |
 | SVC.tls.overrideServiceTargetPort  | Override service port to use for TLS without touching Kong containerPort              |                          |
 | SVC.tls.parameters                 | Array of additional listen parameters                                                 | `["http2"]`              |
+| SVC.extraPorts                     | Array of additional listen ports                                                      | `[]`                     |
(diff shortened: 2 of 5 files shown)


rainest:
This level of flexibility is something we expect to handle [in our implementation of the Gateway API spec](https://incubator.konghq.com/p/gateway-operator/), where [listener configuration is part of the specification](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1beta1.Gateway).

As a workaround, Helm [will respect changes to generated resources across upgrades](https://helm.sh/docs/faq/changes_since_helm2/#improved-upgrade-strategy-3-way-strategic-merge-patches) so long as they don't conflict with something Helm wants to change. An additional Service port won't conflict with the existing ones, so you can modify it independent of the chart configuration.

If you edit the Service to add:

```
  - name: kong-proxy-extra
    port: 9999
    protocol: TCP
    targetPort: 9999
```

and set:

```
env:
  proxy_listen: "0.0.0.0:8000, 0.0.0.0:8443 http2 ssl, 0.0.0.0:9999"
```

It'd be equivalent.

pereiratps:
Thanks for the answer, @rainest.

I don't know if I understand the first option, I still don't know the model with gateway well.
I use kong with database and admin API for KongA. Is the option compatible with this scenario?

I found the workaround option interesting, but I don't feel safe using it in a productive environment without some extra automation. I thing It would be much better if it was on the helm chart.

Thank you anyway. If there's anything I can do that would make it possible to include this feature, just let me know.

rainest:
The Ingress spec covers routing configuration only. The Gateway spec expands on that to add configuration for the application (in our case, Kong instances) ingesting that configuration. Its [Gateway resource](https://gateway-api.sigs.k8s.io/api-types/gateway/) lets you specify proxy ports via its Listeners section. It also adds standardized resources for L4 routing.

Helm does provide [an automation hook](https://helm.sh/docs/topics/advanced/#post-rendering) to patch generated resources. You can use it to execute [kustomize](https://kubectl.docs.kubernetes.io/references/kustomize/) or similar.

PR #857, opened 2023-08-10
fix: fix for comparing semver for enterprise versions

### LINKED ISSUE #856: Chart Image Not Valid Semantic Version
We are using the most recent version of the kong helm chart (2.26) and this morning we are unable to update or install kong on our clusters. 

When we attempt to deploy, we are getting the following error (we have not changed anything with our values file)
image

"error calling semverCompare: Invalid Semantic Version"

pmalek:

This fixes an issue where an enterprise version (like e.g. `3.4.0.0`) wouldn't be allowed to be provided.

Fixes #856 

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -4,6 +4,13 @@
 
 Nothing yet.
 
+## 2.26.1
+
+### Fixed
+
+* Fix parsing enterprise tags (like e.g. `3.4.0.0`)
+  [#857](https://github.com/Kong/charts/pull/857)
+
 ## 2.26.0
 
 ### Improvements
--- a/charts/kong/Chart.yaml
+++ b/charts/kong/Chart.yaml
@@ -11,7 +11,7 @@ maintainers:
 name: kong
 sources:
 - https://github.com/Kong/charts/tree/main/charts/kong
-version: 2.26.0
+version: 2.26.1
 appVersion: "3.3"
 dependencies:
 - name: postgresql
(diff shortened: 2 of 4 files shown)


rainest (commented):
My kingdom for Enterprise using standard semver build tags exclusively for its extra stuff so it at least parses and sticks its mystery values in the designated mystery value box.

What's the intended effect of the change? Regex inscrutability on top of the normal template mess inscrutability will be hard to read in the future, so we should have an explanatory comment inline explaining how this intends to parse input.

We presumably need `^` anchors, correct? What happens if we get a `4.3.4.1` where the tail is a valid semver version we'd compare against?

randmonkey:
What if the `latest` tag is used (as used in KIC nightly integration tests)?

pmalek:
> We presumably need `^` anchors, correct? What happens if we get a `4.3.4.1` where the tail is a valid semver version we'd compare against?

Yes, added that.

I've simplified the change to just extract the first 3 segments if those are detected, otherwise fallback to old behaviour. This will fix the 4 segment versions and leave the old behaviour otherwise.

pmalek:
Not sure why the tests are failing 😠 https://github.com/Kong/charts/actions/runs/5831706278/job/15815656070?pr=857#step:10:2018 the pods deploy successfully 🤔

rainest (approved):
Problem diagnosis: chart-testing is terrible at reporting issues. `--helm-extra-args "--debug"` will make Helm explain its reasoning, because for some reason it doesn't actually explain them on failure.

Do not use chart-testing's `--debug` in most cases, since it _suppresses_ Helm debug output for some reason.

Deleting the other CI files temporarily makes the output less of a mess.

Problem cause: chart-testing's KIND does not use any LB provider, and Helm will block installs if a spawned LB never gets addresses. All the CI entries need `proxy.type: NodePort`.

PR #858, opened 2023-08-10
tests(kong): add -values suffixes to ct test values.yaml files

pmalek:

It seems that we have missed some of the `-values.yaml` suffixes in `kong` chart ct tests.

As per `ct install --help`:

```
$ ct install --help
Run 'helm install', 'helm test', and optionally 'helm upgrade' on

...

Charts may have multiple custom values files matching the glob pattern
'*-values.yaml' in a directory named 'ci' in the root of the chart's
directory. The chart is installed and tested for each of these files.
If no custom values file is present, the chart is installed and
tested with defaults.

```

This PR adds those so that the tests are run.

It also sets the readiness probe period to 1 second to speed up the tests.

### DIFF
--- a/charts/kong/ci/custom-labels.yaml
+++ b/charts/kong/ci/custom-labels.yaml
@@ -1,6 +1,3 @@
-
 # install chart with some extra labels
-
 extraLabels:
   acme.com/some-key: some-value
-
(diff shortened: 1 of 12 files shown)


pmalek:
This has already been done in #865

PR #855, opened 2023-08-10
chore(ingress): bump chart deps and release 0.5.0

czeslavo:
#### What this PR does / why we need it:

Bumps `kong/ingress` dependencies to latest `kong/kong` and triggers `kong/ingress` 0.5.0 release.

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [x] PR is based off the current tip of the `main` branch.
- [x] Changes are documented under the "Unreleased" header in CHANGELOG.md
- [x] New or modified sections of values.yaml are documented in the README.md
- [x] Commits follow the [Kong commit message guidelines](https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#commit-message-format)

PR #859, opened 2023-08-10
deps(ingress): bump ingress kong/kong dep to 2.26.0

pmalek:

Bump ingress kong/kong dependency to 2.26.0 and up.

### DIFF
--- a/charts/ingress/CHANGELOG.md
+++ b/charts/ingress/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## 0.4.1
+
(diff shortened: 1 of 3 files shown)


rainest (approved):
Rebased to main to get the Kubernetes matrix changes from https://github.com/Kong/charts/pull/860/files

rainest:
On second thought, this probably should be a minor bump (as in https://github.com/Kong/charts/pull/855/) since it's a minor bump in the dependency, and the dependency is most of this chart.

PR #862, opened 2023-08-15
chore(ingress): bump chart deps and release 0.6.0

programmer04:

Enforce using the newest [kong/kong 2.26.2](https://github.com/Kong/charts/blob/main/charts/kong/CHANGELOG.md#2262) that includes required fixes - `KongConsumerGroup` CRD and extended subresource `Status` 

continuation of https://github.com/Kong/charts/pull/861

#### Checklist

### DIFF
--- a/charts/ingress/CHANGELOG.md
+++ b/charts/ingress/CHANGELOG.md
@@ -1,11 +1,18 @@
 # Changelog
 
+## 0.6.0
+
+### Improvements
+
+- Bumped dependency `kong/kong` to `2.26.2`.
+  [#862](https://github.com/Kong/charts/pull/862)
+
 ## 0.5.0
 
 ### Improvements
 
 - Bumped dependencies on `kong/kong` chart to `>=2.26.0`.
-  [#855](https://github.com/Kong/charts/pull/855
+  [#855](https://github.com/Kong/charts/pull/855)
 
 ## 0.4.0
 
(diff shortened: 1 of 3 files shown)


rainest (changes requested):
We want to get patches into `ingress` automatically, correct? https://helm.sh/docs/chart_best_practices/dependencies/#versions indicates we can use ranges here, though I don't know what we'd want to use on an ongoing basis.

Here we'd probably floor to 2.26.2 (we definitely want those CRD fixes), so `~2.26.2` but we need some process around how we handle the charts relationship going forward.

programmer04:
Currently, we have a dichotomy. On the one hand, we're liberal in https://github.com/Kong/charts/blob/98079a0677b2ac248074e0c470c32925e692e807/charts/ingress/Chart.yaml#L14-L24
but on the other, we pin the version in

https://github.com/Kong/charts/blob/98079a0677b2ac248074e0c470c32925e692e807/charts/ingress/Chart.lock#L1-L9
hence version is explicitly specified and what you mentioned as far as I know works only for `Chart.yaml` not for `Chart.lock`. To have auto-update we shouldn't include `Chart.lock` file at all. But still new release of `kong/ingress` has to be made to include the newest `kong/kong` in the package.

[kong/kong 2.26.2](https://github.com/Kong/charts/blob/main/charts/kong/CHANGELOG.md#2262) includes an important fix that makes consumer group works, thus I'm following the path it's been done always, the explicit setting version in `Chart.lock`.

We should have a broader discussion about how we want to handle it because doing it manually is tedious

rainest (approved):

PR #864, opened 2023-08-17
chore(docs): Update charts.konghq.com to mention kong/ingress

### LINKED ISSUE #845: Improve homepage of Helm Chart
Currently, page https://charts.konghq.com contains the below
image

It doesn't include information about [kong/ingress](https://github.com/Kong/charts/blob/main/charts/ingress/README.md). It should contain clear guidelines for when to use which one.

Other minor issues
- comment `# Helm 3` seems to be redundant since [previous versions are deprecated](https://helm.sh/blog/helm-v2-deprecation-timeline/) for a long time
- including `$` with commands make them harder to copy, because you have to omit `$`

mheap:

Add `kong/ingress` to the charts.konghq.com homepage

Resolves #845 

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,18 +4,31 @@ This is the official Helm Charts repository for installing Kong on Kubernetes.
 
 ## Setup
 
-```bash
-$ helm repo add kong https://charts.konghq.com
-$ helm repo update
+Add the repo on your machine:
 
-# Helm 3
-$ helm install kong/kong --generate-name
+```bash
+helm repo add kong https://charts.konghq.com
+helm repo update
+```
+
+If you plan to use the Kong Ingress Controller, use the `kong/ingress` chart:
+
+```bash
+helm install kong/ingress --generate-name
+```
+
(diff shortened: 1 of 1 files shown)


rainest (commented):
We probably need to cover these in more detail. The default installs of either will install a controller-managed Kong instance, but will install either the split deployment discovery sort or the older single-deployment DB-less install.

Something to the effect of:

> `ingress` is an opinionated controller-managed DB-less configuration, designed for <general use case whatever> and <general use case whatever, but with Konnect>.

> `kong` is a flexible chart that can deploy various Kong configurations (hybrid, database-backed, etc.) depending on chart configuration. The _Deployment Options_ section of [its README](https://github.com/Kong/charts/blob/main/charts/kong/README.md) covers these in more detail.

We shouldn't show the default install command and discuss the non-default configurations absent other info, since it won't actually do any of those.

rainest:
Superseded by https://github.com/Kong/charts/pull/874

PR #865, opened 2023-08-17
ci: use ktf for ct related tests

pmalek:

Use ktf for ct related tests.

This will install metallb which will assign IPs to LoadBalancer services and this will allow testing without usage of `proxy.type` set to `NodePort`.

#### Checklist


rainest (approved):
Swapped the non-standard readiness for loading a config in the KIC-less instance, since that feels more natural.

Issue #869, opened 2023-08-19
Kong manager unable to connect to admin api


elhackeado:
Kong manager is unable to locate the right URI for the admin api. As per the [chart readme](https://github.com/Kong/charts/tree/main/charts/kong#service-location-hints) we can use Service location hints to make the Kong aware where admin api service is running using `admin_api_uri ` under `.env` key in values.yaml of the chart.

But according to [kong official documentation](https://docs.konghq.com/gateway/3.4.x/reference/configuration/#admin_api_uri) `admin_api_uri` is renamed to `admin_gui_api_url`. We might need an update in the chart readme.

PR #870, opened 2023-08-19
fix(manager): updated kong service location hint var

elhackeado:
As per the latest [Kong documentation](https://docs.konghq.com/gateway/3.4.x/reference/configuration/#admin_api_uri) `admin_api_uri` has been updated to `admin_gui_api_url`. This PR contains the updates related to the renaming of this variable wherever required.

- This PR will fix issue #869 where Kong Manager is unable to connect to admin api.

#### Checklist

### DIFF
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -965,6 +965,7 @@ the template that it itself is using form the above sections.
 
 {{- if .Values.admin.ingress.enabled }}
   {{- $_ := set $autoEnv "KONG_ADMIN_GUI_API_URL" (include "kong.ingress.serviceUrl" .Values.admin.ingress) -}}
+  {{- $_ := set $autoEnv "KONG_ADMIN_API_URI" (include "kong.ingress.serviceUrl" .Values.admin.ingress) -}}
 {{- end -}}
 
 {{- $_ := set $autoEnv "KONG_PROXY_LISTEN" (include "kong.listen" .Values.proxy) -}}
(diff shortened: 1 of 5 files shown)


charts/kong/templates/_helpers.tpl
```diff
@@ -964,7 +964,7 @@ the template that it itself is using form the above sections.
 {{- end -}}
 
 {{- if .Values.admin.ingress.enabled }}
-  {{- $_ := set $autoEnv "KONG_ADMIN_API_URI" (include "kong.ingress.serviceUrl" .Values.admin.ingress) -}}
+  {{- $_ := set $autoEnv "KONG_ADMIN_GUI_API_URL" (include "kong.ingress.serviceUrl" .Values.admin.ingress) -}}
```
rainest, line 967: There's no harm in still setting the old variable, correct? The chart is intended for multiple Kong versions, so we should set both unless the old value breaks things. A brief test indicates you can set both; it just warns about the deprecated value on 3.4+

```suggestion
  {{- $_ := set $autoEnv "KONG_ADMIN_GUI_API_URL" (include "kong.ingress.serviceUrl" .Values.admin.ingress) -}}
  {{- $_ := set $autoEnv "KONG_ADMIN_API_URI" (include "kong.ingress.serviceUrl" .Values.admin.ingress) -}}
```

rainest (changes requested):
The config load should still honor the deprecated value for now AFAICT. You may have an unrelated issue causing #869.

Either way we'd need to retain support for older versions.

charts/kong/templates/_helpers.tpl
amandeep-s-maersk: yes that makes sense. I am using current version (2.26.3) of chart with both variables set and its running fine. We can keep both.

elhackeado: yes that makes sense. I am using current version (2.26.3) of chart with both variables set and its running fine. We can keep both.

rainest (approved):

PR #871, opened 2023-08-21
chore(ci) test against older KIC/Kong versions

rainest:
Adds a second integration tests matrix for testing older versions.

Rather stuck because this is already broken due to the ingress chart incompatibility with KIC 2.10 and Kong 3.3 unless you override the readiness endpoint. Unless we can figure out a clever fix to that, this will just fail CI.

Ideally, this would avoid needing to set these versions manually and the duplicated integration KIND version setting. For the former two, it's unfortunately not fun to reach into the values.yaml and extract the current value and subtract from it, though sort of possible with YAML parsing and some sort of semver math tool (ignoring going below minor `x.0`).

Also, condensing the test list shown on the PR Actions preview would be nice. There are a lot of them now.

#### Checklist

### DIFF
--- a/.github/workflows/main-pr.yaml
+++ b/.github/workflows/main-pr.yaml
@@ -130,7 +130,6 @@ jobs:
       matrix:
         kic-version:
         - "2.10"
-        - "2.9"
         kong-version:
         - "3.3"
         - "3.2"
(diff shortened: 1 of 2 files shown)


rainest:
2.10 now has backports that handle the 3.3 endpoint gracefully; I don't plan to add the same for 2.9. This thus only includes one old KIC version for now, but we can add N versions back going forward depending on what we need to catch.

randmonkey (approved):

PR #872, opened 2023-08-22
Update README with info about both charts

rainest:
It's like https://github.com/Kong/charts/pull/864 with the additional context mentioned in https://github.com/Kong/charts/pull/864#pullrequestreview-1583405496. Left out the Konnect bit since I figure that's covered in the Konnect documentation. Also gets rid of the Helm 2 stuff and explicit installCRDs, since that's ancient.

I don't know if we want to mention the umbrella chart stuff, since that's (maybe) kinda intentionally omitted for `kong/ingress`. However, I expect many users will need to use some of the additional values.yaml settings, and as-is I'm not sure we mention that the `kong/kong` values work in those subsections.

This PR goes to main, which will update the main repo README. Apparently the `gh-pages` branch does _not_ get updated by chart-releaser, which is annoying. Attempting to cherry-pick those commits directly on top of main broke, since they were based on a very old version.

I don't actually know if we expect many people to hit https://charts.konghq.com's landing page, but if so we need to mirror that manually for now. I don't know offhand how to mirror a single file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,11 @@ environment. It is the recommended chart for new installations. To use it:
 helm install kong/ingress --generate-name
 ```
 
+`kong/ingress` is an umbrella chart using two instances of the `kong/kong`
+chart with some pre-configured values.yaml settings. The `controller` and
+`gateway` subsections support additional settings available in the `kong/kong`
+values.yaml.
+
 `kong/kong` is a flexible building block for supporting a wide variety of
 environment configurations not supported by `kong/ingress`, such as hybrid mode
 or unmanaged (no controller) Kong instances. To use it:
--- a/README.md
+++ b/README.md
@@ -4,21 +4,32 @@ This is the official Helm Charts repository for installing Kong on Kubernetes.
 
 ## Setup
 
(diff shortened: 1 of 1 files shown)


mheap (approved):

README.md
```diff
@@ -4,21 +4,46 @@ This is the official Helm Charts repository for installing Kong on Kubernetes.
...
+documentation's table of contents.
+
 ## Documentation
 
-The documentation for Kong's Helm Chart is available at
-[here](https://github.com/Kong/charts/blob/main/charts/kong/README.md).
+The documentation for Kong's Helm Charts is available on GitHub:
+
+* [kong/ingress](https://github.com/Kong/charts/blob/main/charts/ingress/README.md)
+* [kong/kong](https://github.com/Kong/charts/blob/main/charts/kong/README.md)
```
pmalek, line 45: Nit can we make those relative links?

rainest: No, this will get copied to https://charts.konghq.com/ after, and those files won't exist there.

pmalek (approved):

PR #886, opened 2023-09-26
Update admission webhook config, remove Chart.lock, and release kong 2.29.1

rainest:

See https://github.com/Kong/kubernetes-ingress-controller/pull/4733. Adds that; releases 2.28.1.

Removes Chart.lock and adds it to gitignore.

Technically we had GWAPI checks available for earlier versions also, but they never made it into the chart webhook config. I'm gating them to 2.12+ because backporting the checks isn't that crucial.

I don't think we should be including Chart.lock in our distributed chart, though unfortunately docs are kinda vague about this. https://helm.sh/docs/helm/helm_dependency_update/ and https://helm.sh/docs/helm/helm_dependency_build/ suggest the intended use of this file is so that installs from a local copy of the chart repo can be restored to their original installed version in some sort of data loss scenario.

We did not include a Chart.lock until [2.8.1](https://github.com/Kong/charts/commit/2e89647d86ec7c8fa2a2a1697a325e6679b18d08) and had no issues. Chart.yaml includes its own version rules that look more appropriate for distribution (you can specify ranges rather than a pinned version+hash, though we don't actually use them for the Postgres dep).

AFAIK this got included because the `helm dependency` commands you use to get dependency updates always create it, and it's not obvious whether you should include it or not.

#### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 charts/kong/charts
 charts/ingress/charts
+**/Chart.lock
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -4,18 +4,18 @@
 
 Nothing yet.
 
-## 2.29.0
-
-### Improvements
-* Make it possible to set the admission webhook's `timeoutSeconds`.
-
-## 2.28.1
+## 2.29.1
 
 ### Fixed
 
 * The admission webhook now includes Gateway API resources and Ingress
   resources for controller versions 2.12+. This version introduces new
   validations for Kong's regex path implementation.
+
+## 2.29.0
+
+### Improvements
+* Make it possible to set the admission webhook's `timeoutSeconds`.
 
 ## 2.28.0
 
--- a/charts/kong/Chart.lock
+++ b/charts/kong/Chart.lock
@@ -1,6 +1,0 @@
-dependencies:
-- name: postgresql
-  repository: https://charts.bitnami.com/bitnami
-  version: 11.9.13
-digest: sha256:051285066cef2799e39e2953c4abd405c36510a09e9e1bd1833a29224daffddb
-generated: "2022-12-19T11:56:46.951582785-08:00"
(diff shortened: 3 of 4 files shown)


pmalek:
> I don't think we should be including Chart.lock in our distributed chart, though unfortunately docs are kinda vague about this. https://helm.sh/docs/helm/helm_dependency_update/ and https://helm.sh/docs/helm/helm_dependency_build/ suggest the intended use of this file is so that installs from a local copy of the chart repo can be restored to their original installed version in some sort of data loss scenario.

It's also for the reproducible builds. When the dependency changes you'd probably get a warning/error because the digests don't match. Not sure how much we'd like to pay attention to that though.

rainest:
> It's also for the reproducible builds. When the dependency changes you'd probably get a warning/error because the digests don't match. Not sure how much we'd like to pay attention to that though.

For the intended purpose of our distributables I think no--if we're declaring an allowed dependency range, we expect that the exact dependency will vary within that range. What Chart.yaml and Chart.lock express are at odds.

I think we'd have more use for ensuring that a given tag is unchanged (e.g. that the contents of ingress 3.4.5, including its allowed dependency ranges, are always the same even if someone tries to rewrite history). Helm doesn't directly provide that (Chart.lock tracks dependency hashes only, it does not hash the chart itself), but the newer OCI option for distributing charts [does](https://github.com/opencontainers/image-spec/blob/main/descriptor.md#digests) AFAIK.

It makes sense once you've installed a particular state and want to get back to exactly that, e.g.
1. Chart.yaml says dependency must be `~3`
1. Your release pulled dependency 3.4.1 because that was the latest 3.4 available at the time.
1. Environment asplode and state is lost.
1. You want to return to exactly what you installed originally.

Chart.lock can then ensure that _you_ get 3.4.1 still even if 3.5 has since been released, but we have no reason for unrelated new installs to get 3.4.1--we do want them to get 3.5.0.

This is somewhat moot because [chart-releaser behavior effectively forces you to a specific version anyway](https://github.com/Kong/charts/pull/887) (albeit minus hash guarantees), but ignoring that, AFAIK exact reproducible dependencies aren't something we want to enforce, e.g. here [we indicate compatibility with a wide range of Postgres major versions](https://docs.konghq.com/gateway/latest/support/third-party/) that you may choose to pin yourself.

PR #896, opened 2023-10-11
Disable PodDisruptionBudget for replicaCount: 1

ubergesundheit:

Check if replicaCount is above 1 before including the PodDisruptionBudget when enabled.

Installations leaving `replicaCount: 1` as default and specifying `podDisruptionBudget.enabled: true` will prevent kong Deployments from progressing during upgrades.

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Improvements
 
-* Prevent installing PodDisruptionBudget for `replicaCount: 1`.
+* Prevent installing PodDisruptionBudget for `replicaCount: 1` or `autoscaling.minReplicas: 1`.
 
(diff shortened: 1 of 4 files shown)


rainest (changes requested):
Can you make this a [template fail](https://helm.sh/docs/chart_template_guide/function_list/#fail) condition instead of silently deleting the resource?

AFAICT this should only happen on accidental misconfiguration (there's no way to reconcile a single-replica release with a disruption budget), so this should probably immediately fail in the client, to alert you to choose either multiple replicas or no PDB.

ubergesundheit:
Sure :) I've changed the template to fail on both `replicaCount: 1` and `autoscaling.minReplicas: 1`

rainest (approved):

PR #914, opened 2023-10-26
feat(kong): allow using templates when specifying `controller.proxy.nameOverride`

pmalek:

When using the [`ingress` chart 
](https://github.com/Kong/charts/tree/main/charts/ingress) we have to specify the proxy name override 

https://github.com/Kong/charts/blob/824d02276e92d4fea084f603b6833fbd0a38ee92/charts/ingress/values.yaml#L7

in order to get controller's `CONTROLLER_PUBLISH_SERVICE` set properly.

Currently the override is set to `kong-gateway-proxy` which works only when the release name (set by the user) is `kong`. When a different release name is chosen users have to know what to change and how to change it, e.g. when using a release name `kong123` `controller.proxy.nameOverride` has to be set to `kong123-gateway-proxy`.

Without this change installing the `ingress` chart using `kong123` release name yields the following controller env variables:

```
- name: CONTROLLER_KONG_ADMIN_SVC
  value: kong/kong123-gateway-admin
- name: CONTROLLER_PUBLISH_SERVICE
  value: kong/kong-gateway-proxy
```

With this change:

```
- name: CONTROLLER_KONG_ADMIN_SVC
  value: kong/kong123-gateway-admin
- name: CONTROLLER_PUBLISH_SERVICE
  value: kong/kong123-gateway-proxy
```

without the user specifying the override themselves.

Release `kong/kong` chart 2.30.0.

### DIFF
--- a/charts/ingress/values.yaml
+++ b/charts/ingress/values.yaml
@@ -4,7 +4,7 @@ deployment:
 
 controller:
   proxy:
-    nameOverride: kong-gateway-proxy
+    nameOverride: "{{ .Release.Name }}-gateway-proxy"
 
   enabled: true
 
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## Unreleased
+## 2.30.0
 
 ### Improvements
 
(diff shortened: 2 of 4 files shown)


czeslavo (approved):
👍

pmalek:
Holding this for @mheap's review

mheap:
LGTM. Should we deprecate `gatewayDiscovery.generateAdminApiService` to use `tpl` for admin service discovery too?

pmalek:
> LGTM. Should we deprecate `gatewayDiscovery.generateAdminApiService` to use `tpl` for admin service discovery too?

I'm not sure but even if that should be a separate PR.

PR #923, opened 2023-11-02
Split the Deployment

rainest:
#### What this PR does / why we need it:

Splits the Deployment into separate proxy and controller Deployments. Adds settings to configure the Deployments settings separately where settings previously configured the one unified Deployment. Rearranges some settings.

#### Which issue this PR fixes
*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*
  - fixes #921

#### Special notes for your reviewer:

Currently this is just a bunch of TODOs explaining the tasks needed to handle this, for easier understanding of what the end result would look like.

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [ ] PR is based off the current tip of the `main` branch.
- [ ] Changes are documented under the "Unreleased" header in CHANGELOG.md
- [ ] New or modified sections of values.yaml are documented in the README.md
- [ ] Commits follow the [Kong commit message guidelines](https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#commit-message-format)

PR #935, opened 2023-11-09
feat: set default router flavor to expressions if KIC >=3.0 and traditional_compatible otherwide

### LINKED ISSUE #1188: Chart uses the expression router by default
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem Statement

The Helm chart installs KIC+Kong with the expression router enabled by default IF KIC is >= 3.0.

### Proposed Solution

ditto

### Additional information

_No response_

### Acceptance Criteria

- [ ] If `helm install`ing with KIC 3.0 using the official instructions in the docs, the helm chart sets the router flavor to `expression` unless overridden

randmonkey:

Add `kong.router_flavor` to set router flavor to:
- `values.env.router_flavor` if specified
- `expressions` if KIC is enabled and image version >= 3.0.0
- `traditional_compatible` otherwise (REVIEW: is it better to leave it empty to let Kong gateway to use the default value)?

  - fixes https://github.com/Kong/charts/issues/1188

#### Checklist

### DIFF
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -1144,6 +1144,7 @@ the template that it itself is using form the above sections.
 {{- end }}
 
 {{- $_ := set $autoEnv "KONG_PLUGINS" (include "kong.plugins" .) -}}
+{{- $_ := set $autoEnv "KONG_ROUTER_FLAVOR" (include "kong.router_flavor" .) -}}
 
 {{/*
     ====== USER-SET ENVIRONMENT VARIABLES ======
@@ -1709,4 +1710,16 @@ extensions/v1beta1
     {{- end -}}
 {{- end -}}
 {{- (toYaml $proxyReadiness) -}}
+{{- end -}}
+
+{{- define "kong.router_flavor" -}}
+{{- if .Values.env.router_flavor -}}
+.Values.env.router_flavor
+{{- else -}}
+    {{- if (and  .Values.ingressController.enabled  (semverCompare ">= 3.0.0" (include "kong.effectiveVersion" .Values.ingressController.image))) -}}
+        expressions
+    {{- else -}}
+        traditional_compatible
+    {{- end -}}
+{{- end -}}
 {{- end -}}
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -8,9 +8,11 @@
   [#917](https://github.com/Kong/charts/pull/917)
 * Added services resource to admission webhook config for KIC >= 3.0.0.
   [#919](https://github.com/Kong/charts/pull/919)
-* Update default ingress controller version to v3.0
+* Update default ingress controller version to v3.0.
   [#929](https://github.com/Kong/charts/pull/929)
   [#930](https://github.com/Kong/charts/pull/930)
+* Set default router flavor to `expressions` for KIC >= 3.0.0.
+  [#935](https://github.com/Kong/charts/pull/935)
 
 ### Fixed
 
--- a/charts/kong/values.yaml
+++ b/charts/kong/values.yaml
@@ -84,11 +84,6 @@ deployment:
 # is set below. In general, you should not set values here if they are set elsewhere.
 env:
   database: "off"
-  # the chart uses the traditional router (for Kong 3.x+) because the ingress
-  # controller generates traditional routes. if you do not use the controller,
-  # you may set this to "traditional_compatible" or "expressions" to use the new
-  # DSL-based router
-  router_flavor: "traditional"
   nginx_worker_processes: "2"
   proxy_access_log: /dev/stdout
   admin_access_log: /dev/stdout


charts/kong/templates/_helpers.tpl
```diff
@@ -1710,3 +1711,15 @@ extensions/v1beta1
...
 {{- end -}}
+
+{{- define "kong.router_flavor" -}}
+{{- if .Values.env.router_flavor -}}
+.Values.env.router_flavor
+{{- else -}}
+    {{- if (and  .Values.ingressController.enabled  (semverCompare ">= 3.0.0" (include "kong.effectiveVersion" .Values.ingressController.image))) -}}
+        expressions
+    {{- else -}}
+        traditional_compatible
```
czeslavo, line 1722: As I understand this also changes the default when KIC is < 3.0.0 from `traditional` to `traditional_compatible`, is this intended?

pmalek:
> REVIEW: is it better to leave it empty to let Kong gateway to use the default value

On one hand setting this in chart makes this predictable for us. Regardless of Kong Gateway version we use as default, we'll always get the expected behavior. On the other hand though, users might be expecting the Gateway defaults to translate into corresponding chart settings.

We've had a long history of **setting** the router flavor so I'd say we retain this and keep setting it in the chart. If there'a debate we'd like to have on this I suggest we consider pros and cons of that in a separate issue.

charts/kong/templates/_helpers.tpl
pmalek: 👍 on keeping whatever was the default - `traditional` for KIC <3.0.

czeslavo (changes requested):
As mentioned on the team sync, we have to keep in mind this won't work as expected with `kong/ingress` chart. `gateway` part of the release doesn't know about the `controller`'s `ingressController.image`. The condition will be executed based on the `gateway`'s `ingressController.image` (which will be the default from `values.yaml` - so we accidentally get KIC 3.0 there _always_ if a user for some weird reason didn't change `gateway.ingressController.image` to something else). 

As an example, this change will break Konnect `values.yaml` where we still want to keep using `traditional_compatible` router with KIC 2.11/2.12:

```
controller:
  ingressController:
    image:
      tag: "2.11"
...

gateway:
  image:
    repository: kong/kong-gateway
    tag: "3.4"
...
 ```

~As this will pick `expressions` based on the default value of `gateway.ingressController.image` which is 3.0 now.~

EDIT after checking actual results: As in the template there's `.Values.ingressController.enabled` condition required to set router flavor to `expressions`, for `kong/ingress` chart this will always default to `traditional_compatible`. The problem persists though because the router flavor is not changed based on the KIC version as intended. So, if we set `controller.ingressController.image.tag` to `3.0` to start using KIC 3.0 in `kong/ingress`, Gateway's router flavor will still default to `traditional_compatible`.

I think we have to either:
- Wait for @rainest's https://github.com/Kong/charts/pull/923
- Add some templating to `kong/ingress` that will somehow override `gateway.ingressController.image` to what is set in `controller.ingressController.image`.

czeslavo:
I was trying to come up with some workaround that would make this work with `kong/ingress`, but I couldn't achieve anything that would satisfy the requirement of `helm install kong kong/ingress` just setting the expression router based on controller image tag. 😞 

I was trying to use `.Values.global` or templates to override values of the `gateway` subchart using values of the `controller`, but that is not possible (there was a proposal in Helm charts repo implementing that, but it was rejected: https://github.com/helm/helm/pull/6876). 

The only more-or-less reasonable way I came up with was adding a new variable in `kong/kong` values (`deployment.kong.alignWithControllerVersion`) that could allow users to explicitly tell to auto-align gateway's config based on a given controller version. https://github.com/Kong/charts/pull/936 But of course, it has drawbacks, e.g. requires setting an additional value that also should match `ingress.ingressController.image.tag` to not blow up (that could be validated).

I think I just assured myself that https://github.com/Kong/charts/pull/923 is the only way we can make it work as we expected.

pmalek:
I couldn't find the helm doc page on this but `.Subcharts` seems to be the thing that we're looking for (at least as a workaround until/if we get to the split Deployment in `kong` chart).

https://github.com/Kong/charts/pull/937 shows how it could be used to enforce the router flavor in `ingress` chart using subcharts values.

czeslavo:
> I couldn't find the helm doc page on this but `.Subcharts` seems to be the thing that we're looking for (at least as a workaround until/if we get to the split Deployment in `kong` chart).
> 
> Kong/kubernetes-ingress-controller#937 shows how it could be used to enforce the router flavor in `ingress` chart using subcharts values.

As I understand that would prevent using anything else but `expressions` with KIC >= 3.0 which might be too rigorous 🤔 

Sidenote: If I'm not mistaken, `.Subcharts` is an equivalent of getting these values from `.Values.controller` or `.Values.gateway`.

pmalek:
> As I understand that would prevent using anything else but `expressions` with KIC >= 3.0 which might be too rigorous 🤔

Hmm, yes. We'd like to set this for the user as in perform the defaulting in template code rather than limit the possibilities as https://github.com/Kong/charts/pull/937 is trying to do.

In this case I'm not sure this is possible this way (or at least without massive rewrite of `kong` chart into something like a gigantic template which would be rendered in `ingress` using provided subchart values 🤯 )

> Sidenote: If I'm not mistaken, `.Subcharts` is an equivalent of getting these values from `.Values.controller` or `.Values.gateway`.

`.Subcharts` is an object holding fields like `.Files`, `.Values` etc.

`.Subcharts.controller.Values` and `.Subcharts.gateway.Values` can be used in `ingress` chart to get values from subcharts.

czeslavo:
@pmalek I came up with another alternative: https://github.com/Kong/charts/pull/939. It sets the router flavor to expressions in `kong/ingress` and prevents using this kind of flavor with KIC < 3.0. It has a drawback that would make it break when somebody:
- has `kong/ingress` release installed with KIC version pinned to one lower than 3.0
- has router flavor set to `expressions` already 
- wants to keep using the pinned KIC with the new chart version

pmalek:
cc: @randmonkey 

In the effort of cleaning up the stale branches I'm closing this and removing the branch ( it's possible to recover that via "restore branch" button).

PR #936, opened 2023-11-09
add `alignWithControllerVersion` to allow auto-adjustment of GW settings

czeslavo:

It's an attempt to provide a workaround for the inability to propagate values between two subcharts (`gateway` and `controller`) in `kong/ingress` chart. 

Branches off https://github.com/Kong/charts/pull/935. 

This is a bit explicit way, but I couldn't come up with anything else that could allow us to pick a `controller.ingressController.image` value and somehow inject it into the other subchart.

Example `values.yaml` that demonstrates how this could be used in Konnect installation instructions:

```yaml
controller:
  ingressController:
    image:
      tag: &ingressControllerVersion "2.11"
    env:
      feature_gates: "FillIDs=true"
    konnect:
      license:
        enabled: true
      enabled: true
      runtimeGroupID: "adea34c5-90c6-4e23-bc93-3c2df8554f72"
      tlsClientSecretName: konnect-client-tls
      apiHostname: "us.kic.api.konghq.tech"

gateway:
  deployment:
    kong:
      # Align Gateway's settings with the controller version capabilities (e.g. router flavor).
      alignWithControllerVersion: *ingressControllerVersion
  image:
    repository: kong/kong-gateway
    tag: "3.3"
  env:
    konnect_mode: "on"
    vitals: "off"
    cluster_mtls: pki
    cluster_telemetry_endpoint: "3e07ee37d9.tp0.konghq.tech:443"
    cluster_telemetry_server_name: "3e07ee37d9.tp0.konghq.tech"
    cluster_cert: /etc/secrets/konnect-client-tls/tls.crt
    cluster_cert_key: /etc/secrets/konnect-client-tls/tls.key
    lua_ssl_trusted_certificate: system

  secretVolumes:
    - konnect-client-tls
```

This should be safe, as when `gateway.deployment.kong.alignWithControllerVersion` will be left empty, the old default will be used, so we're not breaking anything. It would be opt-in and instructions would have to be updated to use that.

An obvious drawback is that it won't work with just `helm install kong kong/ingress`. Installation command would have to be more complex, i.e.: 
```bash
helm install kong kong/ingress --set gateway.deployment.kong.alignWithControllerVersion=3.0 --set controller.ingressController.image.tag=3.0
```

So it's on a similar level of user-friendliness as just telling the user to set the router flavor explicitly if they're using KIC >= 3.0:
```bash
helm install kong kong/ingress --set gateway.env.router_flavor=expressions
```

### DIFF
--- a/charts/ingress/Chart.lock
+++ b/charts/ingress/Chart.lock
@@ -1,9 +1,9 @@
 dependencies:
 - name: kong
-  repository: https://charts.konghq.com
+  repository: file://../kong/
   version: 2.31.0
 - name: kong
-  repository: https://charts.konghq.com
+  repository: file://../kong/
   version: 2.31.0
-digest: sha256:2897c91fb6e37c04f99a413dac85f991fed223591b8de327a04795e27b90a617
-generated: "2023-11-06T14:36:11.845865+01:00"
+digest: sha256:43b29b6944b4f933dac8cea47c076a151e793df14464d52b4029c9d5689c5c6d
+generated: "2023-11-09T13:56:18.166194+01:00"
--- a/charts/ingress/Chart.yaml
+++ b/charts/ingress/Chart.yaml
@@ -13,11 +13,11 @@ appVersion: "3.4"
 dependencies:
   - name: kong
     version: ">=2.31.0"
-    repository: https://charts.konghq.com
+    repository: file://../kong/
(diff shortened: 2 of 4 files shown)


czeslavo:
This won't be a desired solution, closing.

PR #937, opened 2023-11-09
ingress: enforce expressions router flavor when using KIC v3.0+

pmalek:

Use an obscure `.Subcharts` variable to get subcharts values and based on that enforce the `expressions` router flavor in Kong Gateway when KIC v3.0+ is used

### DIFF
--- a/charts/ingress/templates/_helpers.tpl
+++ b/charts/ingress/templates/_helpers.tpl
@@ -1,0 +1,9 @@
+{{- define "ingressController.enforceGatewayRouterFlavorFor3_0AndUp" -}}
+{{ $gatewayValues := .Subcharts.gateway.Values }}
+{{ $controllerValues := .Subcharts.controller.Values }}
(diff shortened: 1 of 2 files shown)


czeslavo:
Nothing really important, but I tested both `Subcharts.gateway.Values` and `.Values.gateway` and they give the same data (validation fails as expected). I have no idea which way is the good way as I couldn't find documentation for any. 😬 
```diff
--- a/charts/ingress/templates/_helpers.tpl
+++ b/charts/ingress/templates/_helpers.tpl
@@ -1,6 +1,6 @@
 {{- define "ingressController.enforceGatewayRouterFlavorFor3_0AndUp" -}}
-{{ $gatewayValues := .Subcharts.gateway.Values }}
-{{ $controllerValues := .Subcharts.controller.Values }}
+{{ $gatewayValues := .Values.gateway }}
+{{ $controllerValues := .Values.controller }}
```

PR #939, opened 2023-11-09
feat: make expressions router default

czeslavo:

Sets the expression router as default in `kong/kong`. Adds validations that ensure that:
- `expressions` router flavor is not used with KIC < 2.10 as it's not supported
- when `expressions` router flavor is used with KIC < 3.0, the feature gate must be set.

to both `kong/kong` and `kong/ingress` charts.

Validations will output instructions on how to fix the issue if it is encountered.

It also changes the way we refer to `kong/kong` in `kong/ingress`'s `Chart.yaml`: instead of using the repository URL that forces us to bump charts in two steps, use `file://` to make it possible in one step.

Should solve https://github.com/Kong/charts/issues/1188.

#### Checklist

### DIFF
--- a/charts/ingress/values.yaml
+++ b/charts/ingress/values.yaml
@@ -44,3 +44,4 @@ gateway:
   env:
     role: traditional
     database: "off"
+    router_flavor: "expressions"
--- a/charts/ingress/values.yaml
+++ b/charts/ingress/values.yaml
@@ -44,4 +44,3 @@ gateway:
   env:
     role: traditional
     database: "off"
-    router_flavor: "expressions"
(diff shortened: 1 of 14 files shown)


charts/ingress/values.yaml
```diff
@@ -44,3 +44,4 @@ gateway:
   env:
     role: traditional
     database: "off"
+    router_flavor: "expressions"
```
randmonkey: Could we add a template to check image of KIC?

czeslavo: Can you please elaborate on what you mean by this? There are two validation rules added in `validations.tpl` file: one for KIC < 2.10 and one for KIC < 3.0.

czeslavo:
Per @mheap, I'm labeling this with `HOLD/DO NOT MERGE` until Konnect is compatible with expressions.

pmalek:
Could potentially be put in [`kong` v3 milestone](https://github.com/Kong/charts/milestone/2)

PR #943, opened 2023-11-15
feat: allow hostname override

nicklasfrahm:

Our API management uses Kong and we have dataplane deployments in multiple regions. At the moment our lovely uber-API guy REMOG creates a hardcopy of the helm chart, which drastically increases maintenance effort on his side.

This is the only feature he is missing, so the platform guy (me) is helping out.

#### Checklist


charts/kong/README.md
```diff
@@ -853,6 +853,7 @@ On the Gateway release side, set either `admin.tls.client.secretName` to the nam
 | deployment.minReadySeconds         | Minimum number of seconds for which newly created pods should be ready without any of its container crashing, for it to be considered available. |                     |
 | deployment.initContainers          | Create initContainers. Please go to Kubernetes doc for the spec of the initContainers |                     |
 | deployment.daemonset               | Use a DaemonSet instead of a Deployment                                               | `false`             |
+| deployment.hostname                | Set a custom hostname. Very useful in controlplane/dataplane configurations.          |                     |
```
rainest:
```suggestion
| deployment.hostname                | Set the Deployment's `.spec.template.hostname`        |                     |
```

charts/kong/values.yaml
```diff
@@ -60,6 +60,9 @@ deployment:
   # Use a DaemonSet controller instead of a Deployment controller
   daemonset: false
   hostNetwork: false
+  # Set a custom hostname. This is very useful when running a 
+  # controlplane/dataplane setup with multiple dataplanes.
```
```suggestion
  # Set the Deployment's spec.template.hostname field.
```

rainest (changes requested):
Suggested changed comments to refer to the upstream field, in a somewhat futile effort to explain how our values.yaml structure maps to underlying resources (we've been very bad about flattening different levels of resources).

Do you have more information about how this is actually used in hybrid installs in practice? It'd be better to explain that in detail under https://github.com/Kong/charts/blob/main/charts/kong/README.md#hybrid-mode. From the original comments it's not clear how the custom hostname would affect that or what values you should use for the control and data plane releases vis a vis their other settings.

Is there any reason to include `subdomain` along with this? https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-hostname-and-subdomain-fields is a bit confusing as to the exact effects--I'm not sure if this has any effect on the cluster DNS entries or just the local hostname (and whether the former matters).

remogvestas:
we use the hostname variable to have a saying hostname in the response of the  /clustering/data-planes endpoint, like:

data: [ [-]
     { [-]
       config_hash: b4d10b38ea9c510b0d5590c2c04a4fdb
       hostname: **dataplane-git-weu**
       id: 010be8cf-c5f9-42ea-b94f-76a3a05bda2c
       ip: xx.xx.xx.xx
       last_seen: 1700132083
       sync_status: normal
       ttl: 1209581
       version: 3.1.1.1
       
      so we easily can check the status of all dataplanes across all regions and instances

rainest (approved):

Issue #921, opened 2023-11-01
Separate proxy and controller Deployments


rainest:
https://github.com/Kong/kubernetes-ingress-controller/issues/4671 talks about chart stuff but isn't in this repo (presumably for milestone reasons). This is a stub to provide charts tracking of the same.

rainest:
## tl;dr

Using an umbrella chart results in a number of limitations on template
development (in general, we cannot share information between the two releases,
and thus cannot, for example, automatically configure the controller admin API
target based on the proxy's admin API Service configuration) and challenges to
release processes. If we no longer wish to support deploying the controller as
a sidecar, it makes sense to use a single chart that deploys the controller and
Kong each in their own Deployment.

This requires some redesign of the values.yaml to support configuring the
Deployments independently. There are several options for this:

- Do not actually require changes or add new settings, and simply apply the
  same settings to both Deployments equally. This is a bad option.
- Apply existing settings to the proxy Deployment and create new settings for
  the controller Deployment as needed. This requires fewer changes, but results
  in confusing configuration design, since like settings are configured in very
  different places.
- Reorganize settings for consistency and clearing some debt where we poorly
  structured them in the past. This involves breaking changes. Most of these
  changes are just moving a key to a new location, however, so a tool to
  automate it shouldn't be difficult to write.

https://github.com/Kong/charts/pull/923 is the companion to this analysis.
It annotates keys of interest in place within values.yaml.

I do not think we should attempt to support both models, as designing good
configuration and templates for that would be quite difficult. If we want to
keep support for the sidecar in any fashion, I'd recommend doing that in a
continued 2.x line while we use 3.x for split Deployments, and work with users
to figure out what they'd need to migrate if it's difficult.

## Background

The current design of the `kong` chart uses a single Deployment, which can
contain a Kong container, controller container, or both. values.yaml only has a
single set of Deployment-level (and Pod-level) settings, as these are shared
equally among all containers.

Migrations Jobs often also share many of these settings even though they're a
separate workload. Since migrations use a Kong container, settings applicable
to the Kong Deployment are often applicable to the migrations, and sharing them
allows us to economize on settings.

To split Kong and the controller into separate Deployments, we thus need to
convert these settings into two settings in most cases. We may opt to fall back
to the original shared setting for both--it's not ideal and limits some
functionality (e.g. you probably _don't_ want the same replica count and
scaling rules for a controller Deployment as a Kong Deployment), but could
allow for a simpler transition for existing users. Defaults for the new
[...]

PR #748, opened 2023-02-15
tests: add golden tests

czeslavo:
#### What this PR does / why we need it:

Adds golden test that renders the chart using values stored under `testdata/values/*.yaml` and compares the result against a golden manifest generated beforehand and stored under `testdata/golden/*.yaml`. 

If the resulting golden manifest is expected to be changed, one can simply run `make generate` to update the golden files.

The rationale for adding this kind of test is we recently had some cases in which we released the chart with unexpected changes. Having such tests could allow us to visually see if there's any change happening in the rendered manifests, allowing us to consciously decide whether the change is expected or not.

#### Which issue this PR fixes

Part of #740.

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [ ] PR is based off the current tip of the `main` branch.
- [ ] Changes are documented under the "Unreleased" header in CHANGELOG.md
- [ ] New or modified sections of values.yaml are documented in the README.md
- [ ] Commits follow the [Kong commit message guidelines](https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#commit-message-format)

PR #947, opened 2023-11-17
Split the Deployment without values reorg

rainest:

A much simpler alternative to https://github.com/Kong/charts/pull/923.

Only creates a new `ingressController.deployment` section for the new controller Deployment. The existing proxy Deployment configuration is left as-is in values.yaml. The Deployment configurations are not homogenous as a result, and some "global" (root-level) configuration (e.g. `podAnnotations`, `containerSecurityContext`) confusingly only apply to the proxy Deployment.

The contents of this PR are ready for review, but I want to figure out how we can maintain the 2.x branch. The chart repo hasn't ever released two major release branches before. Merge blocked pending that.

A much quicker implementation of the minimum in #921 needed to satisfy the separate Deployment, single chart criterion.

Although not strictly required, I'd like https://github.com/Kong/charts/pull/952 to be in place before actually adding the 3.0.0-rc1 bump here.

##### Differences between Deployment features

The ingress controller Deployment does not support using an HPA or DaemonSet mode. These feel misleading to offer given the controller's leader/inactive model. Scaling will not distribute load across instances and the controller does not have an obvious use case for running across all kubelets (whereas the proxy does if being used as a bare metal proxy).

The chart generates a separate ServiceAccount for each Deployment, but only assigns permissions to the controller account. I do not know of any reason the proxy Deployment would need API server permissions--on its own I don't think it needs a ServiceAccount at all, but I think it does need one when used in a mesh, even if the proxy container doesn't usually mount the SAT.

##### Migration from existing `kong/kong` installs

See https://gist.github.com/rainest/9a6961d43858e833decc8976856c757c#file-moved-csv for keys that have _moved_ as part of this change. There are also a number of new keys that aren't listed outside the diff.

##### Differences from `kong/ingress`

See https://gist.github.com/rainest/9a6961d43858e833decc8976856c757c#file-kong-ingress-diff for a diff between default configuration `kong/ingress` and kong/kong` (with these changes).

The admin Service is NodePort instead of headless ClusterIP. NodePort isn't really necessary (it was just the `kong` default), but headless Services are a annoying to check for diagnostic purposes. Since discovery works fine when the Service has a ClusterIP, I'd prefer to keep it in place for human usage.

`KONG_ROLE=traditional` is not explicitly set (from defaults). This is the default in the proxy, we shouldn't need it.

`KONG_KIC` is set again. This doesn't work in `ingress` due to the cross-subchart border.

Volumes differ for the controller and Kong. This is expected; there's no reason the proxy needs, for example, the webhook cert.

#### Checklist


rainest:
Draft still needs the ServiceAccount sorted out, test/linter issues fixed, and manual migration testing.

https://github.com/Kong/charts/pull/923 was, to put it lightly, overly ambitious. The vast majority of the chart, including the Deployment template and most templates it calls, was written with the expectation that it'd be called at the global scope only. Converting these to scoped templates has proven difficult because most pull inputs from multiple sections of values.yaml. Helm's lack of static analysis and unit testing makes this particularly difficult, as the only way to confirm you haven't broken anything is to render and hope you have a values.yaml configuration that triggers any bugs you've introduced.

This simpler approach isn't particularly clean, but it's much less likely to break. Given that I'd like to resolve the kong/ingress split as soon as possible, it makes sense to go with this approach (more or less complete after about 4h) than to try and use the opportunity to revamp everything (still mired in issues after at least 36h). I don't want to use globals still in the new controller Deployment and don't want the Deployment templates to not share future updates, but I think it makes sense to do so at present in the interest of time.

pmalek:
Adding this to a dedicated [`kong` v3 milestone](https://github.com/Kong/charts/milestone/2) since that will probably be a breaking change, right?

charts/kong/values.yaml
```diff
@@ -146,7 +146,7 @@ admin:
   # Enable creating a Kubernetes service for the admin API
   # Disabling this is recommended for most ingress controller configurations
   # Enterprise users that wish to use Kong Manager with the controller should enable this
-  enabled: false
+  enabled: true
   type: NodePort
```
czeslavo: Given that Service Discovery is now to become a default, I think we should use `ClusterIP` type with `spec.clusterIP` set to `None`. `NodePort` would unnecessarily allocate ports while they're not needed from the KIC perspective. Using `NodePort` would also make the attack surface unnecessarily bigger by default.

>  headless Services are a annoying to check for diagnostic purposes.

Could you elaborate on this more? As I understand there should not be a problem to have a port-forward set up for a headless Service. 🤔

charts/kong/values.yaml
```diff
@@ -146,7 +146,7 @@ admin:
   # Enable creating a Kubernetes service for the admin API
   # Disabling this is recommended for most ingress controller configurations
```
czeslavo:
```suggestion
```
It no longer holds with this config.

charts/kong/values.yaml
```diff
@@ -518,24 +518,142 @@ dblessConfig:
 # created in k8s. It uses CRDs for more fine grained control over routing and
 # for Kong specific configuration.
 ingressController:
-  enabled: true
```
czeslavo, line 521: As I agree with moving stuff strictly related to Deployment under `ingressController.deployment` (it avoids building unnecessary abstractions over vanilla K8s API), I'm not convinced we should move this toggle there though. It's not strictly related to the deployment only, but I sense it's more of a feature toggle from the Helm chart perspective. Rendering a controller Deployment is one of its effects, but it also determines whether to generate ServiceAccount,  RBACs, etc. 

Is there any advantage in moving it that I missed?

czeslavo (commented):
👍 I like this less-destructive approach, it should allow smoother transition for our users. 

Side note: regarding detecting changes that we make by such huge refactorings, I think it would be useful to have golden tests in this repository that would allow us to inspect and compare diffs our changes are causing to the configurations rendered from test values files (something that I approached once in https://github.com/Kong/charts/pull/748, but it didn't get enough traction).

charts/kong/values.yaml
rainest: I generally use LoadBalancer Services with metallb because they're easier to use than port-forward (you can always send to the same thing, regardless of whether you've restarted the Pod that was handling your port-forward). `None` ClusterIPs can't be converted to an assigned ClusterIP AFAICT, so you can't easily convert them to something else unless you started with the other configuration.

I also tend to be lazy, install with defaults, and edit my resources in place rather than setting them up in advance. There's no particular reason for this other than preference (I like working with the complete resource YAML instead of building a values.yaml), but unless we need `None` I'd prefer to avoid it.

charts/kong/values.yaml
rainest: Maybe more an artifact of the original design (#923) than anything, which had `deployment.kong.enabled` and `deploment.controller.enabled`. Given that this won't have like configuration paths anyway, there's probably no good reason to move it.

charts/kong/values.yaml
czeslavo: Yeah, I think we can stay with `ClusterIP` without `None` - that's not required. KIC uses its `EndpointSlices` anyway which are populated in both cases.

czeslavo (approved):

czeslavo:
@rainest Please consider reviewing and merging https://github.com/Kong/charts/pull/953 first so we can review generated differences in this PR.

czeslavo:
@rainest Looking at the golden [diff](https://gist.github.com/czeslavo/e043a841de6d63db0b2654566e6e07c3) I generated locally, I cannot see anything controversial. If you're ready, let's merge it. 🚀

rainest:
Into the currently dormant `release/kong-3.x` branch pending an okay of an RC or release. Once we're ready to do that we cut `release/kong-2.x` from main, send future PRs for 2.x there, merge `release/kong-3.x` into main, and delete the separate `release/kong-3.x`

PR #953, opened 2023-11-22
tests: introduce golden tests in kong chart

czeslavo:

Introduces golden test case that renders `kong/kong` chart templates using test `ci/*-values.yaml` and compares them with golden outputs stored in `charts/kong/test/testdata/golden` directory. The golden files can be regenerated with `make test.golden.update`. 

It should allow us to easily inspect changes in the output manifests that we make by introducing changes to templating, default values, etc., and catch potential mistakes.

Follow up to https://github.com/Kong/charts/pull/947.

#### Checklist


rainest:
Is there a way to have it run automatically and add the diff set (either as an actual commit or as a PR comment)? Or trigger it on demand?

KIC changes often don't result in a change to the end output Kong configuration, but changes to the templates almost always do unless they're outside test coverage. Refactors that aren't expected to result in changes (such as the consolidation of Services into a shared function) are more the exception than the rule.

czeslavo:
> Is there a way to have it run automatically and add the diff set (either as an actual commit or as a PR comment)? Or trigger it on demand?

Do you mean some workflow that would automatically commit changes when golden tests fail so that no one has to run `make test.golden.update` locally and push the changes? Probably it's doable, but will require some work.

I wouldn't like to block https://github.com/Kong/charts/pull/947 on it. ~If you agree I could try improving the process by adding the automatic commits in a separate PR.~ Let's just take a look at the [diff](https://gist.github.com/czeslavo/e043a841de6d63db0b2654566e6e07c3) I generated on this branch for #947 and not depend on it. I'll try to come up with something more developer-friendly.

pmalek:
> Do you mean some workflow that would automatically commit changes when golden tests fail so that no one has to run make test.golden.update locally and push the changes?

I would suggest us not try to automate this to an extent where no developer action is necessary  to update the golden files because that IMHO defeats the purpose of those. Having a manual target like `test.golden.update` would hit the spot IMO because then you have to run it after performing changes that did change the test templates.

czeslavo:
> I would suggest us not try to automate this to an extent where no developer action is necessary to update the golden files because that IMHO defeats the purpose of those. Having a manual target like test.golden.update would hit the spot IMO because then you have to run it after performing changes that did change the test templates.

Yeah, that was the initial idea to make it work the same way it works in KIC. That is basically what this PR introduces. I understand that Travis' point is that it will be too much to remember to run `make test.golden.update` after every template change. I have invested no time in implementing automation around that so if we could agree it's fine to have it as it is I'd be more than happy. 😉

tao12345666333:
I agree with this approach, which can indeed help us move forward more safely.

I am considering another possibility, whether introducing tools like helm-unittest can achieve the same effect.
https://github.com/helm-unittest/helm-unittest

pmalek:
> I agree with this approach, which can indeed help us move forward more safely.
> 
> I am considering another possibility, whether introducing tools like helm-unittest can achieve the same effect. https://github.com/helm-unittest/helm-unittest

This looks interesting 🤔 One major difference that I can see immediately between helm-unittest and this PR is that here we'd be diffing and asserting on the full contents of generated templates where in the former we'd only be testing what we define in our test cases. This has both pros and cons.

I'd love to hear @czeslavo's thoughts on this as he's the one that invested his time in proposing this framework.

czeslavo:
I didn't consider `helm-unittest` back when implementing this. Indeed its [Snapshot testing](https://github.com/helm-unittest/helm-unittest?tab=readme-ov-file#snapshot-testing) looks like something that we could use to achieve similar results. I'll give it a go and report back here.

czeslavo:
I tried using `helm-unittest` to achieve the same mechanics as I proposed in this PR (test all `ci/` values files by rendering templates for all of them and keeping golden files/snapshots), but it's not capable of doing it as of now. The main limitation I bumped into was the lack of ability to ignore dynamically generated content when comparing/storing snapshots which simply made the tests always fail because of generated passwords/certificates. Besides that, it also requires creating a YAML file with test definitions for every test suite while I think we would prefer to simply render all templates for a given set of values files like in the proposed solution. 

I also discovered we're not the first to bump into the same limitations - there's [`helm-chartsnap`](https://github.com/jlandowner/helm-chartsnap) tool that was created because of similar reasons (see [motivations](https://github.com/jlandowner/helm-chartsnap?tab=readme-ov-file#motivation)). I tried using that as well and it _almost_  hits the spot for us. It allows ignoring dynamically generated data, but this has to be defined for each values file separately, which in our case would mean lots of duplication again. I created an issue that could potentially bridge this gap for us (https://github.com/jlandowner/helm-chartsnap/issues/30), but we either have to rely on the maintainer to implement this for us, or contribute ourselves. 

I'd suggest going with the solution we have in this PR (it's not that much code in the end) and considering switching to `helm-chartsnap` once the issue I created is resolved, WDYT?

pmalek:
That makes sense. Thanks for checking those @czeslavo! 🙇 

In that case let's go with the approach proposed in this PR and see how it works for us in the long run.

Makefile
```diff
@@ -22,3 +22,16 @@ lint.charts.kong:
 
 lint.shellcheck:
 	shellcheck ./scripts/*
+
+KONG_CHART_GO_TEST_MODULE=charts/kong/tests
+TEST_GOLDEN_TESTCASE_NAME=TestGolden
```
pmalek: We might leave the room to specify this when calling `make`

```suggestion
KONG_CHART_GO_TEST_MODULE ?= charts/kong/tests
TEST_GOLDEN_TESTCASE_NAME ?= TestGolden
```

charts/kong/tests/golden_test.go
```diff
@@ -0,0 +1,97 @@
...
+			goldenFile := goldenFilePath(valuesFile)
+
+			if *update {
+				t.Logf("updating golden file for %q", valuesFile)
+				err := os.WriteFile(goldenFile, rendered, 0644)
+				require.NoError(t, err, "could not write the golden file")
+			}
+
+			expected, err := os.ReadFile(goldenFile)
+			require.NoError(t, err, "could not read the golden file")
```
pmalek:
```suggestion
			require.NoErrorf(t, err, "could not read the golden file: %s", goldenFile)
```

charts/kong/tests/golden_test.go
```diff
@@ -0,0 +1,97 @@
...
+var update = flag.Bool("update-golden", false, "update golden test output files")
+
+// TestGolden goes through all test *-values.yaml (helm values) files in ci/ directory and `helm template` kong chart with those
+// against various Kubernetes versions. The output for each is expected to be equal to the generated golden file stored
+// under `./testdata/golden/<valuesFileName>_golden.yaml`.
+// When a difference in the output is expected, you can regenerate golden files by passing `-update-golden` flag while
+// invoking this test (or using `make test.golden.update`).
+func TestGolden(t *testing.T) {
+	helm.AddRepo(t, &helm.Options{}, "bitnami", "https://charts.bitnami.com/bitnami")
+	_, err := helm.RunHelmCommandAndGetOutputE(t, &helm.Options{}, "dependencies", "build", "../", "--skip-refresh")
```
pmalek, line 31: It would be nice to include the mention of `helm dependencies build` being called and the reason for that.

czeslavo: Yes, good point. Added a comment. 👍

czeslavo:
In the meantime, our issue in the `helm-chartsnap` (https://github.com/jlandowner/helm-chartsnap/issues/30) was resolved. I posted an alternative PR that uses it instead of writing the Go-based test. PTAL: https://github.com/Kong/charts/pull/978

czeslavo:
Closed in favor of https://github.com/Kong/charts/pull/978.

PR #959, opened 2023-11-29
Use mTLS by default for 3.x

rainest:

Enables mTLS on the controller and admin API by default.

Changes the default gateway discovery strategy to `pod` for compatibility with generated certificates. This soft breaks compatibility with controllers older than 3.0 (you can disable mTLS but need to come up with your own approach to admin API security).

Uses the CA generated for the controller client certificate by default if no other CA is configured for the admin API.

Part of the split chart work. Since we need to expose the admin API outside the Pod network, we need to protect it, and mTLS is a readily and universally available option for this.

I don't know of a great way to 

Going to 3.x for staging instead of main.

The original implementation and UX [has some issues](https://github.com/Kong/charts/issues/921#issuecomment-1830952573) that make this change a bit kludgy, but I'm on the fence about completely refactoring it and breaking existing config.

mTLS does have a one significant drawback: client certificates are generally annoying for human-use clients, and using them complicates using Manager and performing diagnostic admin API requests with curl. The generated certificate is particularly bad for this, since there's no mechanism to issue and distribute multiple certificates for the controller and other clients. You'd likely want to operate your own CA if you use Manager or the admin API regularly. I have not tested to see if mTLS introduces additional complications in Manager (beyond browser client certificates being generally annoying).

I don't have any great alternatives here. RBAC is a viable alternative that is simpler to use for non-controller requests, but is not universally available. An additional localhost-only HTTP listen would work for direct admin API requests with a port-forward, but would be a poor option for Manager use.

#### Checklist

### DIFF
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -439,6 +439,7 @@ The name of the Service which will be used by the controller to update the Ingre
 
   {{- if .Values.ingressController.gatewayDiscovery.enabled -}}
     {{- $_ := set $autoEnv "CONTROLLER_KONG_ADMIN_SVC" (printf "%s/%s-%s" (include "kong.namespace" .) (include "kong.fullname" .) "admin") -}}
+    {{- $_ := set $autoEnv "CONTROLLER_GATEWAY_DISCOVERY_DNS_STRATEGY" "pod" -}}
   {{- else -}}
     {{- $_ := set $autoEnv "CONTROLLER_KONG_ADMIN_URL" (include "kong.adminLocalURL" .) -}}
   {{- end -}}
@@ -639,6 +640,13 @@ The name of the Service which will be used by the controller to update the Ingre
 - name: admin-client-ca
   configMap:
     name: {{ template "kong.fullname" . }}-admin-client-ca
+{{- else if (and $.Values.ingressController.adminApi.tls.client.enabled (not $.Values.ingressController.adminApi.tls.client.certProvided)) }}
+- name: admin-client-ca
+  secret:
+    secretName: {{ template "adminApiService.caSecretName" . }}
+    items:
+    - key: tls.crt
+      path: tls.crt
 {{- end -}}
 {{- range $secretVolume := .Values.secretVolumes }}
 - name: {{ . }}
@@ -708,7 +716,7 @@ The name of the Service which will be used by the controller to update the Ingre
   mountPath: /kong_dbless/
     {{- end }}
   {{- end }}
-{{- if or $.Values.admin.tls.client.caBundle $.Values.admin.tls.client.secretName }}
+{{- if or $.Values.admin.tls.client.caBundle $.Values.admin.tls.client.secretName (and $.Values.ingressController.adminApi.tls.client.enabled (not $.Values.ingressController.adminApi.tls.client.certProvided)) }}
 - name: admin-client-ca
   mountPath: /etc/admin-client-ca/
   readOnly: true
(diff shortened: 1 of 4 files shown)


charts/kong/values.yaml
```diff
@@ -182,6 +182,9 @@ admin:
     # - secretName - the secret must contain a key named "tls.crt" with the PEM-encoded certificate.
     # - caBundle (PEM-encoded certificate string).
     # If both are set, caBundle takes precedence.
+    # If neither are set, and the controller uses generated client certificates
+    # ($.Values.ingressController.adminApi.tls.client.enabled=true and $.Values.ingressController.adminApi.tls.client.certProvided=false),
```
czeslavo: I think it's fine to skip the `$.Values.` in the documentation for conciseness.
```suggestion
    # (ingressController.adminApi.tls.client.enabled=true and ingressController.adminApi.tls.client.certProvided=false),
```

charts/kong/templates/_helpers.tpl
```diff
@@ -439,6 +439,7 @@ The name of the Service which will be used by the controller to update the Ingre
 
   {{- if .Values.ingressController.gatewayDiscovery.enabled -}}
     {{- $_ := set $autoEnv "CONTROLLER_KONG_ADMIN_SVC" (printf "%s/%s-%s" (include "kong.namespace" .) (include "kong.fullname" .) "admin") -}}
+    {{- $_ := set $autoEnv "CONTROLLER_GATEWAY_DISCOVERY_DNS_STRATEGY" "pod" -}}
```
czeslavo, line 442: Given it's almost impossible to configure mTLS in a reasonable way when using `ip` strategy, what do you think about changing the default on the KIC side as well?

rainest: My initial response was "I don't know, I don't even know what our use cases were for adding the others; Patryk's doc just said to use `pod` and `pod` does work". Reviewing the actual history there reveals a problem: `pod` apparently [does not work on anything that uses kube-dns (e.g. GKE)](https://github.com/Kong/kubernetes-ingress-controller/issues/4065).

The cert can't list IP SANs because we don't know what they are beforehand, ant KIC#4065 indicates that we can't use `service` either as I read it. It looks like there are are [alternate hostnames](https://github.com/Kong/kubernetes-ingress-controller/issues/4065#issuecomment-1560958169) available under kube-dns, but we didn't implement support for multiple Pod hostname schemes in favor of the simpler IP option, so we'd probably need to revisit that. Would be interested in hearing @pmalek's thoughts when he returns.

rainest: `pod` actually works fine; we had some incorrect assumptions and blamed kube-dns incorrectly: https://github.com/Kong/kubernetes-ingress-controller/issues/3934

Certificates valid for the entire namespace are still a problem, but arguably better than the status quo in `ingress`, where there's no TLS.

Dunno if there should be a default change because it seems like we maybe shouldn't have a switch at all--there should arguably be a single implementation that can validate arbitrary hostnames. Given that 3.0 is out with the IP default, I think we're stuck setting it here--I don't really want to make 3.1 or later the minimum version.

czeslavo: I think I was wrong that the CN in the client certificate matters at all - they will work no matter what we put in the CN entry. 

As I understand, what's still to be done (probably in the future) is not skipping the verification on the controller side and making sure that server certs are valid at least for the `*.<namespace>.pod` wildcard in the default config. Currently, Kong generates self-signed certs and we skip the verification.

czeslavo:
I posted some context that could help understand some design decisions made in the past: https://github.com/Kong/charts/issues/921#issuecomment-1847053879

Given that we now are going to have two separate Deployments in one Helm release, I think we could consider merging the `ingressController.adminApi.tls.client` and `admin.tls.client` settings because now it's technically possible. I'll leave it up to you to decide if we want to do that in `3.x`.

czeslavo (approved):

PR #961, opened 2023-12-05
kong: only set `clusterIP` when `type` is `ClusterIP`

### LINKED ISSUE #960: Cannot specify admin.type as NodePort
When specifing `admin.type` as `NodePort`, the following error occurred:

```
Error: UPGRADE FAILED: failed to create resource: Service "kong-gateway-admin" is invalid: spec.clusterIPs[0]: Invalid value: "None": may not be set to 'None' for NodePort services
```

Example `values.yaml` for `kong/ingress`

```yaml
 gateway:
   image:
     repository: kong/kong-gateway
     tag: "3.4.2.0"
   env:
     license_data:
       valueFrom:
         secretKeyRef:
           name: kong-enterprise-license
           key: license
     # So that we can access the Admin GUI from browser using NodePort without any CORS issue
     admin_gui_url: http://kong-manager.k8s.orb.local:31883 
     admin_gui_api_url: https://kong-admin-api.k8s.orb.local:31884

   manager:
    enabled: true
    type: NodePort
    http:
      # Hardcode the nodePort so that we don't have to change admin_gui_url
      nodePort: 31883 

   admin:
    enabled: true
    type: NodePort
    http:
      # Hardcode the nodePort so that we don't have to change admin_gui_api_url
      nodePort: 31884
```

pmalek:

Only set `clusterIP` when `type` is `ClusterIP`. This will allow using e.g. `NodePort` in `ingress` chart.

  - fixes #960

After this gets merged we can add a test case to `ingress` chart like so:

```yaml
controller:
  ingressController:
    env:
      anonymous_reports: "false"
    enabled: true
    image:
      repository: kong/kubernetes-ingress-controller

    gatewayDiscovery:
      enabled: true
    adminApi:
      tls:
        client:
          enabled: true

gateway:
  env:
    anonymous_reports: "off"
  admin:
    type: NodePort
  readinessProbe:
    initialDelaySeconds: 1
    timeoutSeconds: 1
    periodSeconds: 1
```

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Improvements
 
+* Only set `Service` `clusterIP` when `type` is `ClusterIP`
 * Bumped Kong version to 3.5.
   [#957](https://github.com/Kong/charts/pull/957)
 * Support for `affinity` configuration has been added to migration job templates.
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -251,7 +251,7 @@ spec:
   {{- if .externalTrafficPolicy }}
   externalTrafficPolicy: {{ .externalTrafficPolicy }}
   {{- end }}
-  {{- if .clusterIP }}
+  {{- if (and (eq .type "ClusterIP") .clusterIP )}}
   clusterIP: {{ .clusterIP }}
   {{- end }}
   selector:


czeslavo (approved):

rainest (changes requested):
This should only omit if the value is `None` for a non-ClusterIP type. Choosing a specific ClusterIP with NodePort or LoadBalancer type Services is fine if you're choosing an actual IP, it's only `None` that's incompatible with them.

Ideally we wouldn't need such a check at all, but the current default in `ingress` necessitates it (IIRC you technically can set it to `null` in your values.yaml to override the default, but that's kind of annoying).

pmalek:
Ah yes, `null` does work 🤦 Approved #962

PR #974, opened 2023-12-12
fix(kong): add RBAC rules for listing namespaces when gateway API is detected

### LINKED ISSUE #790: Gateway ClusterRole needs namespace permission
## Problem
* if we specify namespace selector in gateway API, we get an error about RBAC.
```yaml
apiVersion: gateway.networking.k8s.io/v1beta1
kind: Gateway
metadata:
  name: clous-search
  namespace: clous-search
spec:
  gatewayClassName: clous-search
  listeners:
    - name: proxy
      port: 80
      protocol: HTTP
      allowedRoutes:
        kinds:
          - kind: HTTPRoute
        namespaces:
          from: Selector
          selector:
            matchLabels:
              cluster.namespace: "ctower"
```

```
W0418 18:01:06.029833       1 reflector.go:424] pkg/mod/k8s.io/client-go@v0.26.3/tools/cache/reflector.go:169: failed to list *v1.Namespace: namespaces is forbidden: User "system:serviceaccount:clous-search:clous-search-gateway-kong" cannot list resource "namespaces" in API group "" at the cluster scope
E0418 18:01:06.029871       1 reflector.go:140] pkg/mod/k8s.io/client-go@v0.26.3/tools/cache/reflector.go:169: Failed to watch *v1.Namespace: failed to list *v1.Namespace: namespaces is forbidden: User "system:serviceaccount:clous-search:clous-search-gateway-kong" cannot list resource "namespaces" in API group "" at the cluster scope
```

## Solution
* Add namespace RBAC to cluster roles
```yaml
  - verbs:
      - get
      - list
      - watch
    apiGroups:
      - ''
    resources:
      - namespaces
```
* But we cannot override this values. it should be fixed in chart.

https://github.com/Kong/charts/blob/dd8bae5f59be843653ded530cdf2bf988f2ab8c2/charts/kong/templates/_helpers.tpl#L1466-L1513

pmalek:

This adds RBAC rules for listing namespaces. There's not way to detect if user has objects from particular API group present in the cluster so this can only be conditional on the presence of Gateway API CRDs in the cluster.

This could potentially be added (by refactoring) to https://github.com/Kong/charts/blob/72650f5768dde1867812fab42e863de7db60f80f/charts/kong/templates/controller-rbac-resources.yaml#L10-L18 but KIC doesn't need to list namespaces when users do not use Gateway API (and specifically namespace selectors).

Related controller-runtime issue describing why we need those permissions: https://github.com/kubernetes-sigs/controller-runtime/issues/1156

Fixes #790

#### Checklist

### DIFF
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -1255,7 +1255,6 @@ Kubernetes namespace-scoped resources it uses to build Kong configuration.
 
 Collectively, these are built from:
 kubectl kustomize github.com/kong/kubernetes-ingress-controller/config/rbac?ref=main
-kubectl kustomize github.com/kong/kubernetes-ingress-controller/config/rbac/knative?ref=main
 kubectl kustomize github.com/kong/kubernetes-ingress-controller/config/rbac/gateway?ref=main
 
 However, there is no way to generate the split between cluster and namespaced
@@ -1675,6 +1674,13 @@ Kubernetes Cluster-scoped resources it uses to build Kong configuration.
   verbs:
   - get
   - update
+- apiGroups:
+  - ""
+  resources:
+  - namespaces
+  verbs:
+  - get
+  - list
 {{- end }}
 - apiGroups:
   - networking.k8s.io
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -1681,6 +1681,7 @@ Kubernetes Cluster-scoped resources it uses to build Kong configuration.
   verbs:
   - get
   - list
+  - watch
 {{- end }}
 - apiGroups:
   - networking.k8s.io
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
+
+### Fixed
+
+* Add RBAC rules for get and list operations on namespaces so that Gateway API
+  controllers in KIC can access using a cached controller-runtime client.
+  [#974](https://github.com/Kong/charts/pull/974)
+
 ## 2.33.2
 
 * Fix a template bug related to the `affinity` field for migrations Pods.
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixed
 
-* Add RBAC rules for get and list operations on namespaces so that Gateway API
+* Add RBAC rules for get, list and watch operations on namespaces so that Gateway API
   controllers in KIC can access using a cached controller-runtime client.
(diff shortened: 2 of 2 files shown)


charts/kong/templates/_helpers.tpl
```diff
@@ -1583,6 +1582,12 @@ resource roles into their separate templates.
   - get
   - patch
   - update
+- apiGroups:
+    - ""
+  resources:
+    - namespaces
+  verbs:
+    - list
```
czeslavo:
```suggestion
    - get
    - list
    - watch
```

pmalek: Actually I believe we should add that in the `CluterRole` rules rather than in `Role`. PTAL [f412d1e](https://github.com/Kong/charts/pull/974/commits/f412d1ef40fcc7cf396c7b50bd9be1b4f689a6df)

charts/kong/templates/_helpers.tpl
```diff
@@ -1675,6 +1674,13 @@ Kubernetes Cluster-scoped resources it uses to build Kong configuration.
   verbs:
   - get
   - update
+- apiGroups:
+  - ""
+  resources:
+  - namespaces
+  verbs:
+  - get
+  - list
```
czeslavo, line 1683:
```suggestion
  - list
  - watch
```

pmalek: Why would we need the `watch` 🤔 ? This is to address the lookup in  https://github.com/Kong/kubernetes-ingress-controller/blob/2889365e4908b7aa49fbb07ab783cfdf8ff72ff3/internal/controllers/gateway/route_utils.go#L454-L458 (which uses the cached client hence the need for `list`).

czeslavo: It's because the underlying client uses `watch` to keep the cache in sync apparently according to this: https://github.com/kubernetes-sigs/controller-runtime/issues/1156

pmalek: But that's only for the resources that we watch using `For()` or `Owns()`, no? Just like the linked issue suggests

czeslavo: I understand `For()` and `Owns()` is just another example when `watch` is required. The issue itself is purely about `watch` being required when `client.Get()` is used. 🤔

pmalek: So I tried to reproduce the original issue and it seems that we do need the `watch`.

Without it I keep getting 

```
E1212 16:51:09.392184       1 reflector.go:147] pkg/mod/k8s.io/client-go@v0.28.3/tools/cache/reflector.go:229: Failed to watch *v1.Namespace: unknown (get namespaces)
```

Let me add that and mention it in the changelog.

Good call!

czeslavo (approved):

PR #976, opened 2023-12-12
Release consolidated chart as kong 3.0.0-rc1

rainest:

Releases the consolidated chart as kong 3.0.0-rc1.

Removes most existing `ingress` chart content and consolidates some `ingress` documentation into the `kong` files. Leaves a brief migration message and historical under `ingress`.

See [the other PR](https://github.com/Kong/charts/pull/975). The name for the consolidated chart isn't particularly important, but since it is built from the `kong` templates, releasing as `kong` requires much less trickery behind the scenes. There's also much less docs burden if existing `kong` locations continue to exist.


charts/ingress/README.md
```diff
@@ -1,19 +1,7 @@
-## Kong Ingress Controller
+# Migrate to consolidated chart
 
-[Kong for Kubernetes](https://github.com/Kong/kubernetes-ingress-controller)
-is an open-source Ingress Controller for Kubernetes that offers
-API management capabilities with a plugin architecture.
+The 3.x `kong` chart consolidates the functionality provided by the `ingress`
+:q
```
mflendrich: too much vim :)

```suggestion
```

charts/ingress/README.md
```diff
@@ -1,19 +1,7 @@
...
-```bash
-helm repo add kong https://charts.konghq.com
-helm repo update
-
-helm install kong kong/ingress -n kong
-```
-
-If you need more control over what is deployed, see the [kong/kong chart](https://github.com/Kong/charts/blob/main/charts/kong/README.md). Any `values.yaml` setting can be specified in the `controller` or `gateway` section of your `values.yaml` using this chart.
+and 2.x `kong` charts into a single chart. `ingress` users should upgrade to
+`kong` 3.x following the [upgrade guide](https://github.com/Kong/charts/blob/main/charts/kong/UPGRADE.md#300).
```
mflendrich: These instructions will almost certainly change before the final 3.0 release - we've fixed the problematic logic of separate deployments, but, as we talked, we still need to figure out the end-user interface.

There are two options on the table
1. have a single `kong` chart with various targeted `values.yaml` flavors
2. have the `kong` chart as a library with an understanding of it never being used directly by end users, and targeted charts named after use cases

Option 2 is very likely - we'll have to sort it out before we release the chart 3.0.

scripts/test-env.sh
```diff
@@ -92,5 +92,9 @@ kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd/experimenta
 
 echo "INFO: Updating helm dependencies"
 for i in charts/*; do
+  if [ "$i" = "charts/ingress" ]
+  then
+    continue
+  fi
```
mflendrich, line 106: Can we have a more generic mechanism (for example: if `Chart.yaml` is not present) here?

Based on my previous comment: it's likely that the `ingress` chart will be reinstated in another PR very soon and we'll almost certainly forget to update this place.

charts/ingress/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+The 3.x `kong` chart consolidates the functionality provided by the `ingress`
+and 2.x `kong` charts into a single chart. `ingress` users should upgrade to
+`kong` 3.x following the [upgrade guide](https://github.com/Kong/charts/blob/main/charts/kong/UPGRADE.md#300).
```
mflendrich, line 5: As we don't know for sure what the shape will be in 3.0 - I'd rather not imply that the chart name will change for Ingress.

```suggestion
The 3.x `kong` chart consolidates the functionality provided by the `ingress`
and 2.x `kong` charts into a single chart. `ingress` users should upgrade to
3.x following the [upgrade guide](https://github.com/Kong/charts/blob/main/charts/kong/UPGRADE.md#300).
```

mflendrich (dismissed):
I approve this change with a big caveat: we understand that the 3.0 release must support various named use cases and this will take one of two forms, and we understand that the former is more likely
    - differently named charts (`kong/ingress`, `kong/konnect-dp`, `kong/operator`, etc) with "batteries included" using default values.yaml
    - same chart but with different values.yaml examples

Therefore
- I've asked that we remove the definitive language that might otherwise be forgotten to update before we reach a conclusive decision about the 3.0 shape
- we'll continue the march towards 3.0 in https://github.com/Kong/charts/issues/977

PR #978, opened 2023-12-21
tests: introduce golden tests in kong chart using chartsnap

czeslavo:

An alternative to https://github.com/Kong/charts/pull/953 using the existing `helm-chartsnap` helm plugin that became useful for us thanks to its maintainer @jlandowner quickly fixing an issue that was a blocker for us: https://github.com/jlandowner/helm-chartsnap/issues/30.

#### Checklist

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+charts/kong/ci/__snapshots__/*.snap linguist-generated=true
--- a/.github/workflows/release-pr.yaml
+++ b/.github/workflows/release-pr.yaml
@@ -159,6 +159,26 @@ jobs:
       - name: cleanup integration tests (cleanup)
         run: ./scripts/test-env.sh cleanup
 
+  golden-tests:
+    runs-on: ubuntu-latest
+    name: golden-tests
+    steps:
+      - name: checkout
+        uses: actions/checkout@v4
+
+      - name: setup helm
+        uses: azure/setup-helm@v3
+        with:
+          version: v3.11.0
+
+      - name: build helm chart dependency
+        run: |
+          helm repo add bitnami https://charts.bitnami.com/bitnami
+          helm dependency build charts/kong
+
+      - name: run golden tests
+        run: make test.golden
+
   # Workaround to allow checking the matrix tests as required tests without adding the individual cases
   # Ref: https://github.com/orgs/community/discussions/26822#discussioncomment-3305794
   passed:
@@ -168,6 +188,7 @@ jobs:
       - lint-test
       - integration-test
       - oldversion-integration-test
+      - golden-tests
     if: always()
     steps:
       - if: ${{ contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled') }}
--- a/Makefile
+++ b/Makefile
@@ -32,10 +32,17 @@ GOLDEN_TEST_CHART ?= kong
 GOLDEN_TEST_CHART_VALUES_DIR ?= ./charts/kong/ci/
 CHARTSNAP_COMMAND = helm chartsnap -c ./charts/$(GOLDEN_TEST_CHART) -f $(GOLDEN_TEST_CHART_VALUES_DIR)
 
+# Defining multi-line strings to echo: https://stackoverflow.com/a/649462/7958339
+define GOLDEN_TEST_FAILURE_MSG
+>> Golden tests have failed.
+>> Please run 'make test.golden.update' to update golden files and commit the changes if they're expected.
+endef
+export GOLDEN_TEST_FAILURE_MSG
+
 .PHONY: test.golden
 test.golden: chartsnap
-	$(CHARTSNAP_COMMAND) || \
-	(echo ">> Golden tests have failed.\n>> Please run 'make test.golden.update' to update golden files and commit the changes if they're expected." && \
+	@ $(CHARTSNAP_COMMAND) || \
+	(echo "$$GOLDEN_TEST_FAILURE_MSG" && \
 	exit 1)
 
 .PHONY: test.golden.update
--- a/Makefile
+++ b/Makefile
@@ -6,12 +6,17 @@ _download_tool:
 		GOBIN=$(PROJECT_DIR)/bin go generate -tags=third_party ./$(TOOL).go )
 
 .PHONY: tools
-tools: kube-linter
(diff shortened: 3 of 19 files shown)


rainest (commented):
Make target seems a bit off? It looks like this was intended to only print the "failed" message if the command fails, but it's printing always, even though all the individual checks have `PASSED` and the update target results in no changes:

```
$ make test.golden       
helm chartsnap -c ./charts/kong -f ./charts/kong/ci/ || \
(echo ">> Golden tests have failed.\n>> Please run 'make test.golden.update' to update golden files and commit the changes if they're expected." && \
exit 1)
```

What all should trigger a failure? Are there certain things that have blanket exemptions even if we haven't explicitly ignored them?

From poking around a bit, it seems that some changes don't actually trigger one even though they're not in a dynamic field. For example, adding an annotation to the main Deployment's Pod template:

```
$ git diff
diff --git a/charts/kong/templates/deployment.yaml b/charts/kong/templates/deployment.yaml
index 28f9b06..a861edf 100644
--- a/charts/kong/templates/deployment.yaml
+++ b/charts/kong/templates/deployment.yaml
@@ -41,6 +41,7 @@ spec:
   template:
     metadata:
       annotations:
+        kuma.io/service-account-token-volume: "change!"
         {{- if (and (not .Values.deployment.serviceAccount.automountServiceAccountToken) (or .Values.deployment.serviceAccount.create .Values.deployment.serviceAccount.name)) }}
         kuma.io/service-account-token-volume: {{ template "kong.serviceAccountTokenName" . }}
         {{- end }}
```
and doing a basic default values `helm template ana /tmp/symkong/` to compare output does yield a diff in the output:

```
$ diff -u /tmp/old /tmp/new

@@ -514,6 +514,7 @@
   template:
     metadata:
       annotations:
+        kuma.io/service-account-token-volume: "change!"
         kuma.io/service-account-token-volume: ana-kong-token
         kuma.io/gateway: "enabled"
         traffic.sidecar.istio.io/includeInboundPorts: ""
```
but still gets an `All snapshots matched` result. That annotation block [is present in the  snapshot](https://github.com/Kong/charts/blob/d8c436db2e7c33df06711a60b4d10ddbe4d31e05/charts/kong/ci/__snapshots__/default-values.snap#L96-L101), so I'm not quite sure of what to make of that.

Comparatively, flipping the Deployment/DaemonSet if condition does fail:

```
$ git diff
diff --git a/charts/kong/templates/deployment.yaml b/charts/kong/templates/deployment.yaml
index 28f9b06..091fbb4 100644
--- a/charts/kong/templates/deployment.yaml
+++ b/charts/kong/templates/deployment.yaml
@@ -1,6 +1,6 @@
 {{- if or .Values.deployment.kong.enabled .Values.ingressController.enabled }}
 apiVersion: apps/v1
-{{- if .Values.deployment.daemonset }}
+{{- if not .Values.deployment.daemonset }}
 kind: DaemonSet
 {{- else }}
 kind: Deployment
```

```
$ make test.golden       
helm chartsnap -c ./charts/kong -f ./charts/kong/ci/ || \
(echo ">> Golden tests have failed.\n>> Please run 'make test.golden.update' to update golden files and commit the changes if they're expected." && \
exit 1)
 RUNS  Snapshot testing chart=./charts/kong values=charts/kong/ci/admin-api-service-clusterip-values.yaml
 RUNS  Snapshot testing chart=./charts/kong values=charts/kong/ci/custom-labels-values.yaml
 RUNS  Snapshot testing chart=./charts/kong values=charts/kong/ci/default-values.yaml
 RUNS  Snapshot testing chart=./charts/kong values=charts/kong/ci/kong-ingress-1-values.yaml
 RUNS  Snapshot testing chart=./charts/kong values=charts/kong/ci/kong-ingress-2-values.yaml
 RUNS  Snapshot testing chart=./charts/kong values=charts/kong/ci/kong-ingress-3-values.yaml
 RUNS  Snapshot testing chart=./charts/kong values=charts/kong/ci/kong-ingress-4-values.yaml
 RUNS  Snapshot testing chart=./charts/kong values=charts/kong/ci/service-account.yaml
 RUNS  Snapshot testing chart=./charts/kong values=charts/kong/ci/single-image-default-values.yaml
 RUNS  Snapshot testing chart=./charts/kong values=charts/kong/ci/test-enterprise-version-3.4.0.0-values.yaml
 RUNS  Snapshot testing chart=./charts/kong values=charts/kong/ci/test1-values.yaml
 RUNS  Snapshot testing chart=./charts/kong values=charts/kong/ci/test2-values.yaml
 RUNS  Snapshot testing chart=./charts/kong values=charts/kong/ci/test3-values.yaml
 RUNS  Snapshot testing chart=./charts/kong values=charts/kong/ci/test4-values.yaml
 RUNS  Snapshot testing chart=./charts/kong values=charts/kong/ci/test5-values.yaml
 FAIL  Snapshot does not match chart=./charts/kong values=charts/kong/ci/admin-api-service-clusterip-values.yaml
Expected to match
--- kind=ConfigMap name=chartsnap-kong line=2
  - object:
      apiVersion: apps/v1
-     kind: Deployment
+     kind: DaemonSet
      metadata:
          labels:
```

rainest (changes requested):
Okay, the annotation not detecting was because I had a duplicate annotation name, which is not really allowed. Not sure what chartsnap's doing under the hood but I guess it runs it through some processing that processes the collision and chooses the last value (this is how the Kubernetes API handles them) before checking rather than just using raw template output. Kinda weird but whatever, I at least understand it now.

The instruction print is also not processing the newline properly. The output at the end (not the initial command echo) is still a single line: 

```
>> Golden tests have failed.\n>> Please run 'make test.golden.update' to update golden files and commit the changes if they're expected.
```

So if we can sort that and suppress the earlier print we're good. I thought we could use something like

```
false
if [[ $? ]]
then
		echo "good"
fi
```
but alas, the wonders of screwy Make syntax to do something like that bash elude me.

czeslavo:
The `test.golden` makefile target was printing the command itself as there was no `@` before it to suppress the output. With that added, it should no longer misguide with the "Golden tests have failed" in the positive scenario. Also, the newline should work now (it's been working for me without the changes, but apparently that might be an issue in some shells 🤷): 0dccf83

rainest (approved):

PR #987, opened 2024-01-19
feat(env) add support for envFrom

rainest:
Add envFrom and ingressController.envFrom values. These populate the standard container envFrom field, which contains a list of ConfigMap or Secret refs. Keys and values from the referenced resources are set as environment variables in the containers.

Aha idea KIC-I-52 from @felderi 

The root-level `envFrom` violates the modern rule about placing config for the Deployment under the `deployment` key. I put it there for consistency with the existing `env` and `customEnv` values.

This uses the `envFrom` key to populate `envFrom` in the Deployment Kong container as well as the init containers and migrations containers. Several of these only appear in DB mode. test2-values uses Postgres and the the controller, so it hits every use of this value in templates.

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## Unreleased
+
+### Added
+
+* The `envFrom` and `ingressController.envFrom` values.yaml keys now populate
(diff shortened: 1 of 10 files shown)


tao12345666333 (approved):
LGTM

PR #1004, opened 2024-02-07
#1003 Removal of default security context UID setting

### LINKED ISSUE #1003: runAsUser: 1000 in securityContext causes error in Openshift
Openshift does dynamically attribute defined user ranges. Therefore the user 1000 is in general not a valid user.
In version 1.6 it was removed already (### Removal of default security context UID setting) , but was reintroduced with #909.
The securityContext can be defined without the runAsUser and everything works fine.

MarkusFlorian79:
Does reapply a fix done in 1.6.0 This is the text form ./kong/UPGRADE.md:
### Removal of default security context UID setting

Versions of Kong prior to 2.0 and Kong Enterprise prior to 1.3 use Docker
images that required setting a UID via Kubernetes in some environments
(primarily OpenShift). This is no longer necessary with modern Docker images
and can cause issues depending on other environment settings, so it was
removed.

Most users should not need to take any action, but if you encounter permissions
errors when upgrading (`kubectl describe pod PODNAME` should contain any), you
can restore it by adding the following to your values.yaml:

```
securityContext:
  runAsUser: 1000


  - fixes #1003 

#### Checklist

### DIFF
--- a/charts/kong/values.yaml
+++ b/charts/kong/values.yaml
@@ -626,7 +626,6 @@ ingressController:
         securityContext:
           readOnlyRootFilesystem: true
           allowPrivilegeEscalation: false
-          runAsUser: 1000
           runAsNonRoot: true
           seccompProfile:
             type: RuntimeDefault


rainest:
Closing pending further investigation into whether we can satisfy both OpenShift's and base Kubernetes' desires here: https://github.com/Kong/charts/issues/1003#issuecomment-1933042323

PR #1008, opened 2024-02-15
fix: #1005: escale controller status port from being scraped by Prometheus

### LINKED ISSUE #1005: Service monitor scraping both status port of ingress pod
Hello,

since the version 2.35.0 of the helm chart, we see that the service monitor tries to scrape 2 ports per pod. 

The release of the version 2.35.0 renamed the port `10254` of the `ingressController` `status`. The service monitor looks for that port name to scrape metrics, making the prometheus scrape job effectively scrape 2 metrics port. 

However, the ingress-controller does not expose any metrics on the port `10254` making half of the job target `unavailable`.

A simple proposition would be to rename the `status` port of the ingress-controller to `status-ingress` so the service monitor does not discover that port and does not try to scrape metrics that are unavailable.

sgrzemski:
…erviceMonitor

  - fixes #1005

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-* Escape ingress-controller status port from being scraped by Prometheus and its ServiceMonitor
+* Rename the controller status port. This fixes a collision with the proxy status port in the Prometheus ServiceMonitor.
 
 ## 2.36.0
 
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## Unreleased
+## 2.37.1
 
 * Rename the controller status port. This fixes a collision with the proxy status port in the Prometheus ServiceMonitor.
   [#1008](https://github.com/Kong/charts/pull/1008)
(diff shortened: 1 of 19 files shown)


rainest (approved):
Dunno if we should bother making the probe configuration that prompted the problem change use the named port rather than its number, but I figure it doesn't matter much. We don't let you configure it and have no intention of changing it away from 10254.

charts/kong/templates/_helpers.tpl
```diff
@@ -890,7 +890,7 @@ The name of the Service which will be used by the controller to update the Ingre
     containerPort: 10255
     protocol: TCP
   {{- end }}
-  - name: status
+  - name: controller-status
```
rainest:
```suggestion
  - name: cstatus
```

charts/kong/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+* Escape ingress-controller status port from being scraped by Prometheus and its ServiceMonitor
```
rainest:
```suggestion
* Rename the controller status port. This fixes a collision with the proxy status port in the Prometheus ServiceMonitor.
```

rainest (dismissed):

czeslavo (dismissed):

czeslavo (dismissed):

czeslavo (approved):

PR #1014, opened 2024-02-21
tests: add ingress golden tests

czeslavo:

Adds golden tests to the ingress chart.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 charts/kong/ci/__snapshots__/*.snap linguist-generated=true
+charts/ingress/ci/__snapshots__/*.snap linguist-generated=true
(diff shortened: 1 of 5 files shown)


rainest (approved):
Dunno whether it'd be more of a pain to maintain the dynamic field config twice or add additional values.yamls to the `kong` inputs to cover anything we're not testing in `kong`'s own CI.

czeslavo:
> Dunno whether it'd be more of a pain to maintain the dynamic field config twice or add additional values.yamls to the kong inputs to cover anything we're not testing in kong's own CI.

I think that the dynamic fields config won't be updated so often to become painful. This way we also test "real" interaction between two subcharts. It will all depend anyway on how we proceed further with charts anyway so we will adapt later.

PR #1020, opened 2024-02-28
feat: add gateway-operator helm chart

tao12345666333:
Add initial functionality to the gateway-operator Helm chart.

#### Checklist


.github/workflows/release-pr.yaml
```diff
@@ -83,14 +84,15 @@ jobs:
...
-        - "1.23.13"
-        - "1.24.7"
-        - "1.25.9"
-        - "1.26.4"
-        - "1.27.1"
+          - "1.23.13"
+          - "1.24.7"
+          - "1.25.9"
+          - "1.26.4"
+          - "1.27.1"
```
rainest, line 93: Why the heck is (what I presume is) a YAML auto-formatter using the indented array item syntax variant for this but then changing `chart-name` to the same indent as the parent field syntax variant.

YAML :frowning_face:

rainest (commented):
Did an initial scan of this but it looks like most of the resource bits that will be templated (container env, labels, resources) are still using static placeholders.

One note up front is that, for the initial release, we probably want to merge into a prep branch instead of main, so we don't actually release the chart on merge and can split base functionality into multiple PRs for more focused reviews, then merge the prep branch to main once that's complete.

tao12345666333:
How about `prepare-kgo-chart`?

rainest:
Yeah, anything is fine, you'll just need to make a PR to main that adds that branch name to https://github.com/Kong/charts/blob/main/.github/workflows/release-pr.yaml#L11-L13

We should arguably rename that file too. It runs tests, not releases. IDK what I was thinking when I renamed it to that.

.github/workflows/release-pr.yaml
tao12345666333: The modifications I made here are just to keep them consistent.

tao12345666333:
As an initial submission, I want it to work properly first and then continue adding configuration options.

charts/gateway-operator/templates/rbac-resources.yaml
czeslavo, line 1: Can you please leave a comment at the top of this file what was the command to assemble it? It could be useful for someone who will need to sync it with KGO.

charts/gateway-operator/templates/rbac-resources.yaml
```diff
@@ -0,0 +1,815 @@
...
+  resources:
+  - kongclusterplugins/status
+  verbs:
+  - get
+  - patch
+  - update
+- apiGroups:
+  - configuration.konghq.com
+  resources:
+  - kongconsumergroups
```
czeslavo, line 149: 🤔 I don't think KGO's manager's role should have access to KIC resources (like `kongconsumergroups`, `kongconsumers`, etc.). I posted a PR that removes them https://github.com/Kong/gateway-operator/pull/1546.

charts/gateway-operator/templates/NOTES.txt
czeslavo, line 1: Please provide some content or remove the file.

charts/gateway-operator/templates/_helpers.tpl
czeslavo, line 1: How about naming files with helper functions a little bit less generic? I feel this pattern of having just a single generic `_helpers.tpl` didn't end up well in `kong/kong` chart where we have all kind of helpers in a single "bag". I'd suggest going with underscore prefixed files for helper functions, but segregating them. For starters, we could rename this one to `_names.tpl`.

charts/gateway-operator/values.yaml
```diff
@@ -0,0 +1,18 @@
+image:
+  repository: docker.io/kong/gateway-operator-oss
```
czeslavo: We should default to the `docker.io.kong/gateway-operator` (enterprise) image.

charts/gateway-operator/templates/NOTES.txt
rainest: As a basic stub entry, I guess print the namespace and name of the Deployment?

Links to the next "Getting Started" docs section would also make sense. There are two, for Konnect- and KIC-managed configuration:

https://docs.konghq.com/gateway-operator/latest/get-started/kic/create-gateway/
https://docs.konghq.com/gateway-operator/latest/get-started/konnect/deploy-data-plane/

charts/gateway-operator/templates/_helpers.tpl
```diff
@@ -0,0 +1,29 @@
+{{/* vim: set filetype=mustache: */}}
+{{/*
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+*/}}
+
+{{- define "kong.namespace" -}}
```
rainest, line 7: Template helper names are not inherently namespaced by chart, so if we are defining these separately they should use a different prefix than the existing `kong` chart. There's no obvious reason you would use both this and `kong` in an umbrella chart, but these would collide if we did.

IMO `kong.operator.*` makes sense.

Many of these are generic utilities we could share between charts, but subcharts don't work great as a packaging system, and we'd need a lot of additional CI machinery to make it or some home-grown alternative to make it viable.

charts/gateway-operator/templates/config.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: v1
+data:
+  controller_manager_config.yaml: |
+    apiVersion: controller-runtime.sigs.k8s.io/v1alpha1
+    kind: ControllerManagerConfig
```
rainest: Do we actually support these? kubebuilder apparently creates a basic one by default, but we don't load it ourselves. I'm unsure if there's any code in the controller-runtime base manager that loads it for you.

That said, it's deprecated per https://book.kubebuilder.io/component-config-tutorial/define-config and https://github.com/kubernetes-sigs/controller-runtime/blob/56159419231e985c091ef3e7a8a3dee40ddf1d73/pkg/config/config.go#L30-L39, so we should just go with envvars as the only way to manage these.

charts/gateway-operator/templates/deployment.yaml
```diff
@@ -0,0 +1,83 @@
...
+metadata:
+  labels:
+    control-plane: controller-manager
+  name: {{ template "kong.fullname" . }}-controller-manager
+  namespace:  {{ template "kong.namespace" . }}
+spec:
+  replicas: {{ .Values.replicaCount }}
+  selector:
+    matchLabels:
+      control-plane: controller-manager
```
rainest: This should include information about the release to not match Pods from another release, similar to https://github.com/Kong/charts/blob/kong-2.38.0/charts/kong/templates/_helpers.tpl#L39-L43 and https://github.com/Kong/charts/blob/kong-2.38.0/charts/kong/templates/deployment.yaml#L26-L28

charts/gateway-operator/templates/deployment.yaml
```diff
@@ -0,0 +1,83 @@
...
+  strategy:
+    type: Recreate
+  template:
+    metadata:
+      annotations:
+        kubectl.kubernetes.io/default-container: manager
+      labels:
+        control-plane: controller-manager
```
rainest: This will also need the release-scoped selector labels.

charts/gateway-operator/templates/deployment.yaml
```diff
@@ -0,0 +1,83 @@
...
+    spec:
+      containers:
+      - args:
+        - --health-probe-bind-address=:8081
+        - --metrics-bind-address=127.0.0.1:8080
```
rainest, line 32: We should avoid a mix of CLI flags and envvar args, since managing user overrides across both is more complicated.

Probably envvars, sincewe have existing machinery for merging chart-generated and user-provided envvars, but not for args.

charts/gateway-operator/templates/deployment.yaml
```diff
@@ -0,0 +1,83 @@
...
+        command:
+        - /manager
```
rainest: We can just let the [default image command](https://github.com/Kong/gateway-operator-enterprise/blob/75cfc99a43a1c05f34c5e3c46c09fb4b97517251/Dockerfile#L85) handle this instead:
```suggestion
```
It's unlikely we'd ever change it and introduce version compatibility issues, but may as well remove any chance of that.

charts/gateway-operator/templates/deployment.yaml
```diff
@@ -0,0 +1,83 @@
...
+          periodSeconds: 10
+        resources:
+          limits:
+            cpu: 500m
+            memory: 256Mi
+          requests:
+            cpu: 10m
+            memory: 128Mi
+        securityContext:
+          allowPrivilegeEscalation: false
```
rainest, line 60: This should match what we use for the proxy/controller to allow the operator to run in Restricted SCC namespaces:

```suggestion
          allowPrivilegeEscalation: false
          readOnlyRootFilesystem: true
          runAsNonRoot: true
          seccompProfile:
            type: RuntimeDefault
          capabilities:
            drop:
            - ALL
```

Hopefully I didn't break the spacing in the suggestion.

Minor note that we should not need the `runAsUser` here, since the operator uses a numeric UID in its Dockerfile.

charts/gateway-operator/templates/deployment.yaml
```diff
@@ -0,0 +1,83 @@
...
+        volumeMounts:
+        - mountPath: /tmp/k8s-webhook-server/serving-certs
+          name: webhook-certificates
```
rainest, line 70: We should use an emptyDir for all of `/tmp`, similar to https://github.com/Kong/charts/blob/kong-2.38.0/charts/kong/templates/_helpers.tpl#L586-L588 and https://github.com/Kong/charts/blob/kong-2.38.0/charts/kong/templates/_helpers.tpl#L747-L748, to satisfy the read-only root FS requirement.

charts/gateway-operator/templates/deployment.yaml
```diff
@@ -0,0 +1,83 @@
...
+          requests:
+            cpu: 5m
+            memory: 64Mi
+      securityContext:
+        runAsNonRoot: true
+      serviceAccountName: {{ template "kong.serviceAccountName" . }}
+      terminationGracePeriodSeconds: 10
+      volumes:
+      - emptyDir: {}
+        name: webhook-certificates
```
rainest, line 95: Ditto bit about having this handle all of `/tmp` from earlier: https://github.com/Kong/charts/blob/kong-2.38.0/charts/kong/templates/_helpers.tpl#L586-L588

This should include a size limit to avoid any chance of some runaway code creating a bunch of garbage and filling the kubelet emptyDir space, though a static 256M limit should be more than sufficient for the operator (the proxy can use `/tmp` for buffered uploads, so it needs more space).

charts/gateway-operator/templates/services.yaml
```diff
@@ -0,0 +1,15 @@
...
+  name: {{ template "kong.fullname" . }}
+  namespace:  {{ template "kong.namespace" . }}
+spec:
+  ports:
+  - name: https
+    port: 8443
+    protocol: TCP
+    targetPort: https
+  selector:
+    control-plane: controller-manager
```
rainest: This also needs the release-scoped selector.

charts/gateway-operator/templates/deployment.yaml
```diff
@@ -0,0 +1,83 @@
...
+      - args:
+        - --health-probe-bind-address=:8081
+        - --metrics-bind-address=127.0.0.1:8080
+        command:
+        - /manager
+        env:
+        - name: POD_NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
```
rainest, line 37: `env` will probably need to to support user variables for the initial release. It's possible that we can skip that initially for the operator since, AFAIK, there's less reason you'd need to change the defaults for its existing config options.

charts/gateway-operator/templates/deployment.yaml
```diff
@@ -0,0 +1,83 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    control-plane: controller-manager
```
rainest: This and all other resources should support user-provided labels, to allow, for example, adding custom owership metadata.

The [`kong` labels template](https://github.com/Kong/charts/blob/kong-2.38.0/charts/kong/templates/_helpers.tpl#L28-L37) also includes some stock labels related to chart/release info. Unsure if those are particularly useful in practice, but they probably won't hurt.

rainest (changes requested):
Do we have a set of follow-up issues for functionality we intend to add in separate PRs? Those should have TODO comments in the files.

The user-supplied envvars and labels can definitely be handled in a separate PR, but we should at least mark TODOs for them.

charts/gateway-operator/values.yaml
tao12345666333: Yes. It has already been addressed in this commit. https://github.com/Kong/charts/pull/1023/commits/52bbfb9d451ba80b910bb209614631b6a0947161

charts/gateway-operator/templates/rbac-resources.yaml
tao12345666333: Of course, I added comments at the beginning of the file. You can generate this file by executing these commands.

charts/gateway-operator/templates/NOTES.txt
tao12345666333: Updated.

charts/gateway-operator/templates/_helpers.tpl
tao12345666333: ref: https://helm.sh/docs/chart_template_guide/named_templates/

> Conventionally, Helm charts put these templates inside of a partials file, usually `_helpers.tpl`. 

`_helpers.tpl` is a common practice, and the majority of Helm charts use this naming convention.
WDYT?

charts/gateway-operator/templates/_helpers.tpl
tao12345666333: Will the helm chart of KGO be used as a sub-chart? Let's keep it for now and make modifications later. WDYT?

charts/gateway-operator/templates/config.yaml
tao12345666333: Good point! Thank you!
I will remove this file.

charts/gateway-operator/templates/deployment.yaml
tao12345666333: I have recorded the to-do tasks in this issue and will provide support in the subsequent PR.
https://github.com/Kong/gateway-operator/issues/1323

charts/gateway-operator/templates/deployment.yaml
tao12345666333: OK

charts/gateway-operator/templates/deployment.yaml
tao12345666333: Great! Thank you!

I would like to provide this as a follow-up PR, and I have also tracked it in https://github.com/Kong/gateway-operator/issues/1323

charts/gateway-operator/templates/deployment.yaml
tao12345666333: Yes!

I have also tracked "support user variables" in https://github.com/Kong/gateway-operator/issues/1323

charts/gateway-operator/templates/deployment.yaml
tao12345666333: I would like to provide this as a follow-up PR, and I have also tracked it in https://github.com/Kong/gateway-operator/issues/1323

charts/gateway-operator/templates/deployment.yaml
tao12345666333: I would like to provide this as a follow-up PR, and I have also tracked it in https://github.com/Kong/gateway-operator/issues/1323

charts/gateway-operator/templates/deployment.yaml
tao12345666333: OK, I would like to provide this as a follow-up PR

charts/gateway-operator/templates/deployment.yaml
rainest: To clarify, this doesn't need to be configurable, we should just use the static value above unless we find some reason it actually needs to be configurable.

The `kong` chart does let you set it, but AFAIK there's no real reason to. There are some debugging scenarios where you need to add capabilities, but for such debugging tasks I think we should just ask users to modify the Deployment temporarily.

There are some weird caveats around OpenShift if you have `runAsUser`, but this doesn't, and there was maybe a reason to let you modify it when PodSecurityPolicy was still standard. https://kubernetes.io/docs/concepts/security/pod-security-standards/ doesn't have an option for custom policies that I know of.

charts/gateway-operator/templates/_helpers.tpl
rainest: Probably not, but it's something we should do regardless.

The more likely scenario is we find some way to make shared library management easier.

rainest (approved):
For the future, please do leave TODO comments if deferring a review item to a later PR. It can be difficult to find where a change needed to go from the issue alone if we end up deferring it further.

charts/gateway-operator/templates/deployment.yaml
tao12345666333: OK, make sense

PR #1021, opened 2024-02-28
feat(services)  use servicePort=0 to disable http port on service for tls terminated at LB

mithunkrb:
With this change, servicePort=0 can be used to disable http port on service for the services exposed by LoadBalancer(for ex: proxy-service) and TLS is terminated at LB.
This is helpful in the below scenario:
- SSL/TLS terminated at Load Balancer and therefore LB sends HTTP traffic to service.
- We use overrideServiceTargetPort to map service TLS port to container HTTP port.
- However, service still keeps exposing HTTP port which is not desired however its required to expose the http port on container.
-  where its desired to only enable TLS in service however the TLS port in service is mapped to http port in container. 

  - fixes #

#### Checklist

### DIFF
--- a/charts/kong/values.yaml
+++ b/charts/kong/values.yaml
@@ -297,6 +297,9 @@ proxy:
   http:
     # Enable plaintext HTTP listen for the proxy
     enabled: true
+    # Set the servicePort: 0 to skip exposing in the service but still
+    # let the port open in container to allow https to http mapping for
+    # tls terminated at LB.
     servicePort: 80
     containerPort: 8000
     # Set a nodePort which is available if service type is NodePort
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -210,6 +210,7 @@ spec:
   ports:
   {{- if .http }}
   {{- if .http.enabled }}
+  {{- if ne ( .http.servicePort | toString ) "0" }}
   - name: kong-{{ .serviceName }}
     port: {{ .http.servicePort }}
     targetPort: {{ .http.containerPort }}
@@ -220,6 +221,7 @@ spec:
     nodePort: {{ .http.nodePort }}
   {{- end }}
     protocol: TCP
+  {{- end }}
   {{- end }}
   {{- end }}
   {{- if .tls.enabled }}
@@ -329,7 +331,9 @@ Parameters: takes a service (e.g. .Values.proxy) as its argument and returns KON
   {{- $portMaps := list -}}
 
   {{- if .http.enabled -}}
+  {{- if ne (.http.servicePort | toString ) "0" -}}
         {{- $portMaps = append $portMaps (printf "%d:%d" (int64 .http.servicePort) (int64 .http.containerPort)) -}}
+  {{- end -}}
   {{- end -}}
 
   {{- if .tls.enabled -}}


charts/kong/templates/deployment.yaml
```diff
@@ -158,21 +158,25 @@ spec:
...
         {{- if (and .Values.proxy.tls.enabled .Values.proxy.enabled)}}
+        {{- if .Values.proxy.tls.containerEnabled | default true}}
         - name: proxy-tls
           containerPort: {{ .Values.proxy.tls.containerPort }}
           {{- if .Values.proxy.tls.hostPort }}
           hostPort: {{ .Values.proxy.tls.hostPort }}
           {{- end}}
           protocol: TCP
         {{- end }}
+        {{- end }}
```
rainest: We generally want all Services and ports to support the same set of configuration, so these need to be repeated across all port definitions.

It'd make sense to add a helper for these so that we can just loop over them and generate like configuration for each (like we do for the Services), but just adding the same clauses to each block is fine if you don't want to add that to scope.

charts/kong/README.md
```diff
@@ -670,6 +670,8 @@ or `ingress` sections, as it is used only for stream listens.
 |-----------------------------------|-------------------------------------------------------------------------------------------|--------------------------|
 | SVC.enabled                       | Create Service resource for SVC (admin, proxy, manager, etc.)                             |                          |
 | SVC.http.enabled                  | Enables http on the service                                                               |                          |
+| SVC.http.serviceEnabled           | Enables http on the proxy service (to ensure backward compatibility its gets overridden by SVC.http.enabled if unset)                                                               |                          |
+| SVC.http.containerEnabled         | Expose http port on container (to ensure backward compatibility its overridden by SVC.http.enabled if unset)                                                            |                          |
```
rainest: If we don't go with the suggestion to overload `servicePort` and `containerPort` elsewhere, this should delete `(to ensure backward compatibility its overridden by SVC.http.enabled if unset)` and just list a default `true` value.

charts/kong/values.yaml
```diff
@@ -297,6 +297,8 @@ proxy:
   http:
     # Enable plaintext HTTP listen for the proxy
     enabled: true
+    serviceEnabled: true
+    containerEnabled: true
```
rainest: I'd like to avoid yet more knobs if possible, and the interaction between four different sorts of enable toggles for Services feels unintuitive if you don't look at the template innards to see what each disables.

Do you see any issue with instead using special values for existing fields? You could set `servicePort="disabled"` or `containerPort="disabled"` instead of an integer to disable these piecemeal.

Alternately, these could be one of the less-commonly used options we list in the README but not in the base values.yaml, like `hostPort`.

rainest (changes requested):
This also needs a CHANGELOG entry in addition to the other requests.

charts/kong/templates/deployment.yaml
mithunkrb: I agree, it would make more sense to add a helper and keep it consistent for all service. Will make the required change and push. Moreover, will be updating the logic considering your suggestion in this comment https://github.com/Kong/charts/pull/1021#discussion_r1509554696

charts/kong/README.md
mithunkrb: will be updating the logic considering your suggestion in this comment https://github.com/Kong/charts/pull/1021#discussion_r1509554696 accordingly will update the README.md

charts/kong/values.yaml
mithunkrb: `Do you see any issue with instead using special values for existing fields? You could set servicePort="disabled" or containerPort="disabled" instead of an integer to disable these piecemeal.`

I feel it would be better to avoid mixing int & string value for same field to avoid any conversion error/issues. Instead, I'm thinking to use port value as 0(preferably) or 65536 to indicate the disabled state. Does it sound ok? 

As per this article https://ieeexplore.ieee.org/document/8816853, I see below about port "0"
` A special case is port 0 which is reserved but was never assigned. Traffic from and to port 0 is unusual, because it should not occur in the wild. As port 0 is unassigned, there is no common service listing for connections here. Furthermore, operating systems usually interpret the request to open port 0 as the request to allocate and open any currently unused port. Thus, traffic from and to port 0 should not occur, because no application should listen there and applications cannot send from port 0.`

rainest: Yeah, special handling of `0` would be fine.

charts/kong/templates/deployment.yaml
mithunkrb: updated the code with special handling of servicePort=0 to handle the use-case.

charts/kong/README.md
mithunkrb: updated the code with special handling of servicePort=0 to handle the use-case instead of adding extra knobs

charts/kong/values.yaml
mithunkrb: updated the code with special handling of servicePort=0 to handle the use-case instead of adding extra knobs.

rainest (approved):

PR #1023, opened 2024-03-03
feat: Add conditions to automate the installation for CRDs.

tao12345666333:

Because KGO relies on GWAPI and KIC's CRDs, adding these sub-charts can control whether or not to automatically install these dependent CRDs.

In addition, there are some other reasons why using sub-charts can provide us with a simpler way to complete the updates and maintenance of these CRDs.

This PR is based on #1020 .

#### Checklist


charts/gateway-operator/charts/kic-crds/Chart.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v2
+name: kic-crds
+version: 0.0.0
```
randmonkey, line 3: How do we define the chart version of "kic-crds"? If we added new KIC CRDs or changed existing CRDs, should we bump the version and what is the rule?

tao12345666333: Since this sub-chart is directly referenced through the directory, as long as we update this sub-chart, it will be automatically updated when installed via source code.

rainest: Most installs aren't from a source checkout though, they're from the published registry bundle, and those include the specific version available at release time.

rainest (commented):
This should be a draft pending https://github.com/Kong/charts/pull/1020, correct?

I'd like to avoid subcharts if possible to avoid needing to handle multiple releases to complete a release of the chart using them. I'd lean towards installing them out of band from the upstream sources. You end up needing to do that anyway for subsequent upgrades due to Helm's lack of a CRD update mechanism, so IMO handling the initial install differently doesn't net us that much.

Bundling them all with the operator chart would be easier, but there's no way to skip only some CRDs and install the rest if, for instance, the GWAPI CRDs had already been installed separately.

You could maybe set up a modified release job that creates a virtual subchart directory during release, where instead of actually maintaining those in git, the release job checks out the repo, runs `helm create` a few times under the main chart's `charts/` directory, and does a latest version lookup and `kustomize build https://upstream-crd.example/crds > charts/whatever/crds/crds.yaml` to populate it before running chart-releaser.

tao12345666333:
> This should be a draft pending https://github.com/Kong/charts/pull/1020, correct?

Yes. We need to merge #1020 first.

I understand your concerns, Helm does not have a good management mechanism for CRDs.

Managing these dependent CRDs through sub-charts serves two purposes:

* Users can control which CRDs to install/skip by configuring them.
* Installing through sub-charts avoids the potential data overload that may occur when merging all CRDs into a single chart (exceeding the capacity of the helm release secret).

tao12345666333:
> You could maybe set up a modified release job that creates a virtual subchart directory during release, where instead of actually maintaining those in git, the release job checks out the repo, runs helm create a few times under the main chart's charts/ directory, and does a latest version lookup and kustomize build https://upstream-crd.example/crds > charts/whatever/crds/crds.yaml to populate it before running chart-releaser.

What you mean is how do we manage these dependency CRDs in the Git Repo, right?
If our basic logic remains unchanged, I prefer to keep it simpler and directly save it in the Git repo.

tao12345666333:
Sorry, I mistakenly closed this PR and used #1025 as a replacement.

PR #1027, opened 2024-03-08
feat(kgo): Allow users to modify startup parameters and env vars

tao12345666333:

#### Checklist

### DIFF
--- a/charts/gateway-operator/templates/deployment.yaml
+++ b/charts/gateway-operator/templates/deployment.yaml
@@ -30,6 +30,23 @@ spec:
       - args:
         - --health-probe-bind-address=:8081
         - --metrics-bind-address=127.0.0.1:8080
+        {{- if .Values.featureFlags }}
+        {{- if eq .Values.featureFlags.controlPlaneControllerEnabled false }}
+        - --enable-controller-controlplane=false
+        {{- end }}
+        {{- if eq .Values.featureFlags.gatewayControllerEnabled false }}
+        - --enable-controller-gateway=false
+        {{- end }}
+        {{- if eq .Values.featureFlags.dataPlaneControllerEnabled false }}
+        - --enable-controller-dataplane=false
+        {{- end }}
+        {{- if eq .Values.featureFlags.dataPlaneBlueGreenControllerEnabled false }}
+        - --enable-controller-dataplane-bluegreen=false
+        {{- end }}
+        {{- if .Values.featureFlags.aiGatewayControllerEnabled }}
+        - --enable-controller-aigateway=true
+        {{- end }}
+        {{- end }}
         env:
         - name: POD_NAMESPACE
           valueFrom:
--- a/charts/gateway-operator/templates/deployment.yaml
+++ b/charts/gateway-operator/templates/deployment.yaml
@@ -30,21 +30,21 @@ spec:
       - args:
         - --health-probe-bind-address=:8081
         - --metrics-bind-address=127.0.0.1:8080
-        {{- if .Values.featureFlags }}
-        {{- if eq .Values.featureFlags.controlPlaneControllerEnabled false }}
-        - --enable-controller-controlplane=false
+        {{- if .Values.controllers }}
+        {{- if .Values.controllers.controlplaneController }}
+        - --enable-controller-controlplane={{ .Values.controllers.controlplaneController.enabled }}
         {{- end }}
-        {{- if eq .Values.featureFlags.gatewayControllerEnabled false }}
-        - --enable-controller-gateway=false
+        {{- if .Values.controllers.gatewayController }}
+        - --enable-controller-gateway={{ .Values.controllers.gatewayController.enabled }}
         {{- end }}
-        {{- if eq .Values.featureFlags.dataPlaneControllerEnabled false }}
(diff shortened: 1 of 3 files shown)


charts/gateway-operator/values.yaml
```diff
@@ -43,3 +43,19 @@ resources:
...
+# Use this section to change the startup parameters of gateway-operator
+featureFlags:
+  # Enable the ControlPlane controller
+  controlPlaneControllerEnabled: true
+  # Enable the Gateway controller
+  gatewayControllerEnabled: true
+  # Enable the DataPlane controller
+  dataPlaneControllerEnabled: true
+  # Enable the DataPlane BlueGreen controller. Mutually exclusive with DataPlane controller
+  dataPlaneBlueGreenControllerEnabled: true
```
pmalek: Those being mutually exclusive, shouldn't be both enabled by default. I reckon we want to push for BG usage so let's enable that.

```suggestion
  # Enable the DataPlane controller. Mutually exclusive with DataPlane BlueGreen controller
  dataPlaneControllerEnabled: false
  # Enable the DataPlane BlueGreen controller. Mutually exclusive with DataPlane controller
  dataPlaneBlueGreenControllerEnabled: true
```

charts/gateway-operator/values.yaml
```diff
@@ -43,3 +43,19 @@ resources:
...
+  # Enable the AIGateway controller. (Experimental)
+  aiGatewayControllerEnabled: false
+
+# Use this section to change the environment viriables
```
pmalek:
```suggestion
# Use this section to add environment variables to operator's container
```

charts/gateway-operator/values.yaml
```diff
@@ -43,3 +43,19 @@ resources:
   requests:
     cpu: 10m
     memory: 128Mi
+
+# Use this section to change the startup parameters of gateway-operator
+featureFlags:
```
pmalek: How about we rename this section to e.g. `controllers` and then 

we do something like:

```
controllers:
  # ControlPlane controller
  controlPlane:
    enabled: true
...
```

?

This was it's a bit less verbose but still pretty clear how to enable/disable + we get more room for adding new fields later on. WDYT?

pmalek: This will also disambiguate this with feature flags/gates which we don't have now but it's rather probable we'll be adding that at some point.

charts/gateway-operator/templates/deployment.yaml
```diff
@@ -30,7 +30,25 @@ spec:
...
+        {{- if eq .Values.featureFlags.dataPlaneControllerEnabled false }}
+        - --enable-controller-dataplane=false
+        {{- end }}
+        {{- if eq .Values.featureFlags.dataPlaneBlueGreenControllerEnabled false }}
+        - --enable-controller-dataplane-bluegreen=false
+        {{- end }}
+        {{- if .Values.featureFlags.aiGatewayControllerEnabled }}
+        - --enable-controller-aigateway=true
+        {{- end }}
+        {{- end }}
```
pmalek: How about just setting this to whatever's set in the `values.yaml`?

By checking values here we keep "defaults" in 2 place: here and in `values.yaml` (not mentioning the operator's default for flags).

charts/gateway-operator/values.yaml
tao12345666333: I directly used the configuration values in KGO here. Do we also need to modify the default values in KGO?

tao12345666333: ref: https://github.com/Kong/gateway-operator/blob/678c485665fece7baaf34c2fe7edfd8f0f6e33fa/modules/cli/cli.go#L37-L38

pmalek: We might update https://github.com/Kong/gateway-operator/blob/ad98c9e851e73c37dcc23439c5fd649baf1b2479/modules/cli/cli.go#L38 as that's not really mutually exclusive but the BG controller takes precedence: whenever it's enabled we disable the standalone DataPlaneReconciler.

charts/gateway-operator/values.yaml
tao12345666333: SGTM. Let me make the change.

charts/gateway-operator/templates/deployment.yaml
tao12345666333: Because our default values are all true,
If the user does not specify, they should all be true
Only when the user specifies false, do we need to set them as false.

pmalek: Got it. But this way we need to update flags in 2 places if defaults change.

tao12345666333: Yes. 

How about this? `--enable-controller-dataplane=.Values.featureFlags.dataPlaneControllerEnabled`

charts/gateway-operator/values.yaml
tao12345666333: Got it. Thanks

charts/gateway-operator/templates/deployment.yaml
pmalek: This could work yes

pmalek (approved):

PR #1031, opened 2024-03-08
release kgo chart 0.0.1-alpha.1

tao12345666333:

The first release of the KGO chart.

This is the only commit added compared to the `prepare-kgo-chart` branch. https://github.com/Kong/charts/commit/a77739315f096aba2d501769e732575952884136 

This chart has been marked as a pre-release version. https://github.com/Kong/charts/pull/1031/files#diff-687b44a1f065c7497da04c3af622511715f75bbaaedd2ea617678b3f85ff6d88R14

#### Checklist


rainest (approved):
Ah, we didn't actually build the env/arg equivalence we have in the controller into the operator? That probably should have been something we considered necessary for a baseline release.

We don't want to have to replicate every aspect of the application config in the chart, you should be able to expose a generic `env` or `args` list that lets you set them arbitrarily (most should be unset by default--we control the upstream defaults and should choose good defaults there). Something like the `controllers` section to set a subset of args shouldn't actually exist, but we're kinda stuck with that at the moment (or writing something similar to `kong.env` for args, which is annoying due to their lack of a key/value structure).

Dedicated chart sections that set things that can otherwise be set directly via app configuration should only exist when we're composing a set of related settings together, e.g. how the `kong` chart service sections bind together configuration for the Service resource, Deployment resource, and kong.conf environment settings.

---

On a smaller note, we've done the same thing here where `env` is at the root, and we've sorta asploded bits of the Deployment configuration into disparate locations. We're unlikely to ever have this chart create multiple different Deployments like `kong` does, so we won't hit the same challenges we hit with that early design decision there, but IMO we should still try and adhere to a principle of grouping single-resource settings under a section for those resources, structured similarly to the underlying resource.

tao12345666333:
> Ah, we didn't actually build the env/arg equivalence we have in the controller into the operator?

Yes. We have an issue for tracking https://github.com/Kong/gateway-operator/issues/19

> We don't want to have to replicate every aspect of the application config in the chart, you should be able to expose a generic env or args list that lets you set them arbitrarily (most should be unset by default--we control the upstream defaults and should choose good defaults there). Something like the controllers section to set a subset of args shouldn't actually exist, but we're kinda stuck with that at the moment (or writing something similar to kong.env for args, which is annoying due to their lack of a key/value structure).

I agree with you.  We need to make some adjustments, similar to https://github.com/Kong/charts/pull/1027#discussion_r1517732656

There are still many areas that need improvement, and I will continue to follow up on them in the future.

PR #1032, opened 2024-03-12
feat(kgo): use emptyDir as /tmp

tao12345666333:

use emptyDir to satisfy the read-only root FS requirement.

#### Checklist

### DIFF
--- a/charts/gateway-operator/templates/_helpers.tpl
+++ b/charts/gateway-operator/templates/_helpers.tpl
@@ -59,3 +59,14 @@ app.kubernetes.io/instance: "{{ .Release.Name }}"
   value: {{ $val | quote }}
 {{- end -}}
 {{- end -}}
+
+{{- define "kong.volumes" -}}
+- name: {{ template "kong.fullname" . }}-certs-dir
+  emptyDir:
+    sizeLimit: {{ .Values.certsDir.sizeLimit }}
+{{- end }}
+
+{{- define "kong.volumeMounts" -}}
+- name: {{ template "kong.fullname" . }}-certs-dir
+  mountPath: /tmp/k8s-webhook-server/serving-certs
+{{- end }}
--- a/charts/gateway-operator/values.yaml
+++ b/charts/gateway-operator/values.yaml
@@ -64,3 +64,7 @@ controllers:
 
 # Use this section to add environment variables to operator's container
 env: {}
+
+# Use this section to change the certs-dir emptyDir size
+certsDir:
+  sizeLimit: 256Mi
--- a/charts/gateway-operator/templates/deployment.yaml
+++ b/charts/gateway-operator/templates/deployment.yaml
@@ -80,8 +80,7 @@ spec:
             drop:
             - ALL
         volumeMounts:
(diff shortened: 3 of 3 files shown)


charts/gateway-operator/templates/_helpers.tpl
```diff
@@ -59,3 +59,14 @@ app.kubernetes.io/instance: "{{ .Release.Name }}"
...
+
+{{- define "kong.volumes" -}}
+- name: {{ template "kong.fullname" . }}-tmp
+  emptyDir:
+    sizeLimit: {{ .Values.tmpDir.sizeLimit }}
+{{- end }}
+
+{{- define "kong.volumeMounts" -}}
+- name: {{ template "kong.fullname" . }}-tmp
+  mountPath: /tmp/k8s-webhook-server/serving-certs
```
randmonkey, line 71: I think the combination of volume name and the path/usage of the volume may be confusing:
- `{{ kong.fullname }}-tmp` means all tmp files are added into this dir
- the mountPath means that the dir is only for webhook certs
Do you have a better naming strategy?

charts/gateway-operator/values.yaml
```diff
@@ -64,3 +64,7 @@ controllers:
 
 # Use this section to add environment variables to operator's container
 env: {}
+
+# Use this section to change the tmp empty dir size
+tmpDir:
+  sizeLimit: 256Mi
```
randmonkey, line 70: The `256Mi` is just an arbitrary value, or any standards/best practices for this value?

charts/gateway-operator/templates/_helpers.tpl
tao12345666333: `certs-dir`?

randmonkey: sounds good.

charts/gateway-operator/values.yaml
tao12345666333: This directory is used for storing KGO's admission certificates.
Typically, only a tiny space is needed.  https://github.com/Kong/gateway-operator/blob/1e36e7d9381727c65a399dcbf2faf2bcd6dff79e/modules/manager/webhook.go#L130-L148


There is no definitive best practice here; I simply chose a value that would have a relatively minimal negative impact.

charts/gateway-operator/templates/_helpers.tpl
```diff
@@ -59,3 +59,14 @@ app.kubernetes.io/instance: "{{ .Release.Name }}"
   value: {{ $val | quote }}
 {{- end -}}
 {{- end -}}
+
+{{- define "kong.volumes" -}}
+- name: {{ template "kong.fullname" . }}-certs-dir
+  emptyDir:
+    sizeLimit: {{ .Values.tmpDir.sizeLimit }}
```
randmonkey: the key here is not changed while renaming.
```suggestion
    sizeLimit: {{ .Values.certsDir.sizeLimit }}
```

randmonkey (approved):

PR #1034, opened 2024-03-13
support customized service name

gotojeffray:

We'd like to customize the service name for the different kong gateway components because we're running the kong gateway in kong mesh, but the  kuma sidecar just picked up the Kong Gateway service name alphabetically. 

we just want to use the kong-proxy service name to be the 1st one in the alphabet order to keep it consistent even after we enabled kong-admin service.

e.g we will change admin to zadmin.

#### Checklist
- [ x ] PR is based off the current tip of the `main` branch.

### DIFF
--- a/charts/kong/templates/service-kong-admin.yaml
+++ b/charts/kong/templates/service-kong-admin.yaml
@@ -7,7 +7,11 @@
 {{- $_ := set $serviceConfig "namespace" (include "kong.namespace" .) -}}
 {{- $_ := set $serviceConfig "metaLabels" (include "kong.metaLabels" .) -}}
 {{- $_ := set $serviceConfig "selectorLabels" (include "kong.selectorLabels" .) -}}
+{{- if .Values.admin.serviceName }}
+{{- $_ := set $serviceConfig "serviceName" .Values.admin.serviceName -}}
+{{- else }}
 {{- $_ := set $serviceConfig "serviceName" "admin" -}}
+{{- end -}}
 {{- include "kong.service" $serviceConfig }}
 {{ if .Values.admin.ingress.enabled }}
 ---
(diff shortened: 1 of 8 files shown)


rainest (commented):
Absent multiple requests for this, it's something I'd prefer to avoid. The use case feels too narrow and uncommon to justify adding the extra configuration for it.

Can you describe the workflow where the current alpha sort is an issue in detail? It sounds like you're referring to something in https://kuma.io/docs/2.6.x/production/gui/ but it's not completely clear what, and what issue the sort order causes.

Would there be anything else that'd address the issue?

gotojeffray:
Hi Rainest,

I asked kong support engineer to create a feature request - reference case GTWY-I-851. 

Could you have a look at that please?

rainest:
The description there is the same as the original description above. Are you able to add a comment there?

gotojeffray:
sure, we have this problem when Kong gateway works with Kong mesh.

kuma-sidecar running in the kong gateway pod always picks the service name
alphabetically, but we prefer the `kong-proxy` service name to represent
the kong gateway which makes more sense in the mesh.

There is no issue if we only enable the Kong proxy, but the service name in
the mesh will be changed if we enable any service like `kong-manager` or
`kong-admin` in the helm chart.

We need to change all the service names in our MeshTrafficPermissions
because the service name has been changed to another one.

This customized service name will help us to keep the service name
consistent which means we could keep the `kong-proxy` always be the 1st one
on the service name list.

kong-proxy
kong-zadmin
kong-zmananger
kong-zportal
...

rainest:
I've checked the behavior here and reviewed it with the Mesh/Kuma team; this is a bug in the Mesh UI. Most Services are displayed as individual Services, but `gateway_delegated` (i.e. Kong proxy) Services get aggregated under a single name. `gateway_delegated` shouldn't really be identified by Service at all--they're effectively a separate type of entity tied to a Deployment instead of a Service. The Mesh UI currently uses the first alpha sort Service attached to that Deployment somewhat arbitrarily, because it doesn't have a dedicated UI component for representing `gateway_delegated` entities.

https://github.com/kumahq/kuma/issues/3832 tracks the broader work to update the Mesh UI to represent `gateway_delegated` entities separately (https://github.com/kumahq/kuma/issues/3832#issuecomment-1335439106 mentions the current name selection).

Closing this as the effect of the bug is cosmetic.

gotojeffray:
Hi Raniest, thanks for checking that with mesh team. it's a problem in Mesh UI, but also a big problem in the mesh policy like MeshTrafficPemissions. we just want a consistent mesh service name in all the mesh policies.

PR #1043, opened 2024-03-21
kgo: unpin 1.2.0 and set the default tag to 1.2

pmalek:

We've set the default KGO tag as 1.2.0 by accident (setting precedent) where all the other charts in this repo use `<major>.<minor>` tags.

This will allow users to automatically use the newest patch from 1.2.x. (currently 1.2.1)


tao12345666333 (approved):
LGTM

Thanks

PR #1052, opened 2024-04-10
fix(ingress) disable duplicate CRD installation

### LINKED ISSUE #1051: kong ingress duplicate CRD installation
When using ArgoCD and the ingress helm chart you will get the following error
`RepeatedResourceWarning
Resource apiextensions.k8s.io/CustomResourceDefinition//ingressclassparameterses.configuration.konghq.com appeared 2 times among application resources.
`

image

I think this is because currently, ingress uses the kong helm chart as a subChart twice, but don't disable the CRD installation in the gateway.

See:
https://github.com/Kong/charts/blob/e09fd7a54ba99466328ccf9a581e50a3d516f58c/charts/ingress/values.yaml#L41-L42

and

https://github.com/Kong/charts/blob/e09fd7a54ba99466328ccf9a581e50a3d516f58c/charts/kong/templates/custom-resource-definitions.yaml#L1-L4

I'm unsure if this actually works since we also have the folder called CRDs https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-1-let-helm-do-it-for-you, but I'm a bit unsure on how helm/argocd works in this case.

nissessenap:

Remove 

  - fixes #1051 

#### Checklist

### DIFF
--- a/charts/ingress/CHANGELOG.md
+++ b/charts/ingress/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- fix(ingress) remove duplicate CRD installation [#1052](https://github.com/Kong/charts/pull/1052)
+
 ## 0.12.0
 
 ### Improvements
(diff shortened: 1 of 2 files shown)


nissessenap:
I don't think this will work, due to templates for CRD don't work in helm 3 if I understand it correctly https://helm.sh/docs/topics/charts/#custom-resource-definitions-crds
We would need to do skipCRDs on the subcharts, but that isn't possible https://github.com/helm/helm/issues/10540

The issue have popped up a number of times in helm, but no luck. So I'm unsure if there is any way around this. If anyone have a better idea on how to solve it please share.

nissessenap:
This won't change anything, so I will close this

PR #1057, opened 2024-04-17
fix(deploy) Increase container security

### LINKED ISSUE #753: Enable all kube-linter rules
## Context

In https://github.com/Kong/charts/pull/751 we added kube-linter to our CI pipeline, suppressing all rules that were failing. The goal of this ticket is to enable all ignored rules OR add an explicit ignore annotation to the objects that have to break the rule due to some higher requirement.

The annotation can look as follows:
```yaml
metadata:
  annotations:
    ignore-check.kube-linter.io/privileged: "This deployment needs to run as privileged because it needs kernel access"
```

## Rules to be enabled 
- [x] "no-read-only-root-fs"
- [x] "run-as-non-root"
- [ ] "unset-cpu-requirements"
- [ ] "unset-memory-requirements"

## Acceptance criteria
- [ ] No rules are specified in the `exclude` section of the `.kube-linter.yaml` config. 
- [ ] If there was an object in the helm chart that has to break any of the rules, it's explicitly explained in form of an `ignore-check.kube-linter.io/rule-name` annotation

ubergesundheit:
Change default values to run containers with read-only file system and non-root user for both kong and postgres deployments.

  - Fixes https://github.com/Kong/charts/issues/753

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+* Run containers with read-only file system and non-root user to increase container and pod security.
+  **Note**: This change may be incompatible with user sidecar containers. In this case, change the
+  `securityContext` in your values.
+  [#1057](https://github.com/Kong/charts/pull/1057)
+
 ## 2.47.0
 
 ### Changes
@@ -346,7 +355,7 @@
 
 ## 2.26.5
 
-### Fixed 
+### Fixed
 
 * Kuma ServiceAccount Token hints and volumes are also available in migrations
   Pods.
@@ -354,19 +363,19 @@
 
 ## 2.26.4
 
-### Fixed 
+### Fixed
 
-* updated `admin_api_uri` to `admin_gui_api_url` as per [kong documentation](https://docs.konghq.com/gateway/3.4.x/reference/configuration/#admin_api_uri). 
+* updated `admin_api_uri` to `admin_gui_api_url` as per [kong documentation](https://docs.konghq.com/gateway/3.4.x/reference/configuration/#admin_api_uri).
 
 ## 2.26.3
 
-### Fixed 
+### Fixed
 
 * Enabled Service and Ingress in Kong Manager for non enterprise users.
 
 ## 2.26.2
 
-### Fixed 
+### Fixed
 
 * Add missing CRD KongConsumerGroup and extend status subresource for CRDs
 
--- a/charts/kong/values.yaml
+++ b/charts/kong/values.yaml
@@ -741,6 +741,21 @@ postgresql:
   service:
     ports:
       postgresql: "5432"
+  primary:
+    annotations:
+      "ignore-check.kube-linter.io/no-read-only-root-fs": "writable fs is required"
+    podSecurityContext:
+      runAsNonRoot: true
+      seccompProfile:
+        type: RuntimeDefault
+    containerSecurityContext:
+      runAsNonRoot: true
+      seccompProfile:
+        type: RuntimeDefault
+      allowPrivilegeEscalation: false
+      capabilities:
+        drop:
+        - ALL
 
 # -----------------------------------------------------------------------------
 # Configure cert-manager integration
@@ -985,13 +1000,16 @@ podSecurityPolicy:
 priorityClassName: ""
 
 # securityContext for Kong pods.
-securityContext: {}
+securityContext:
+  seccompProfile:
+    type: RuntimeDefault
 
 # securityContext for containers.
 containerSecurityContext:
   readOnlyRootFilesystem: true
   allowPrivilegeEscalation: false
   runAsUser: 1000
+  runAsGroup: 1000
   runAsNonRoot: true
   seccompProfile:
     type: RuntimeDefault
--- a/.kube-linter.yaml
+++ b/.kube-linter.yaml
@@ -1,8 +1,4 @@
 checks:
   exclude:
-    # TODO: exclude no rule
-    # https://github.com/Kong/charts/issues/753
-    - "no-read-only-root-fs"
-    - "run-as-non-root"
     - "unset-cpu-requirements"
     - "unset-memory-requirements"
--- a/charts/gateway-operator/CHANGELOG.md
+++ b/charts/gateway-operator/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## Unreleased
+
+## Changes
+
+- Set `readOnlyRootFilesystem: true` for kube-rbac-proxy
+  [#1057](https://github.com/Kong/charts/pull/1057)
+
 ## 0.4.9
 
 ## Changes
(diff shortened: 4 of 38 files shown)


ubergesundheit:
The reason specifying `runAsNonRoot` only in the `containerSecurityContext` is that injected kuma-init initContainers seem to need elevated privileges.

rainest:
Do you know of anything that's functionally different in terms of enforcement or compliance validation if like settings are set at the Pod versus container level?

The rough justification for having defaults only in the container section is more or less what you've mentioned: we apply those context constraints to any container the chart itself manages, and any other container in the Pod is an unknown sidecar (with Kuma having a special status as a well-known sidecar that we try to accommodate by default). Pod-level settings might thus break those unknown sidecars and AFAIK there's no difference when applying the same setting at both levels. If there's something that's Pod-level only, we should consider that in isolation.

The Postgres settings should be fine, though we've normally mostly ignored it since it's kind of a legacy wart we discourage use of and plan to ultimately remove.

ubergesundheit:
As you already wrote, the pods securityContext applies to all containers in the pod while the securityContext of a container only applies to the single container. There are some fields that can be only configured in the securityContext of the pod (`fsGroup`, `fsGroupChangePolicy`, `supplementalGroups` and `sysctls`) and some fields that are exclusive to the containers securityContext (`allowPrivilegeEscalation`, `capabilities`, `privileged`, `procMount` and `readOnlyRootFilesystem`). Fields `runAsGroup`, `runAsNonRoot`,  `runAsUser`, `seLinuxOptions`, `seccompProfile` and `windowsOptions` are supported by both.

In terms of compliance, IMO it would be better to set most of the stuff in the pods securityContext. But that breaks injected containers. Not sure how much thought was put into kube-linters rules which we're trying to make happy in this PR.

charts/kong/values.yaml
```diff
@@ -963,13 +990,18 @@ podSecurityPolicy:
 priorityClassName: ""
 
 # securityContext for Kong pods.
-securityContext: {}
+securityContext:
+  runAsUser: 1000
+  runAsGroup: 1000
```
```suggestion
```
These (and whether setting them actually breaks anything) will be very image-specific, so I don't think we should be setting them here.

We could set `runAsNonRoot: true` at the Pod level, which will cause Kubernetes to reject the Pod if any additional containers aren't set to or built with non-0 UIDs.

If the linter's too stupid to let that pass the rule instead, then yeah, the rule's a bit busted and we should just leave it off.

rainest (changes requested):
Alright, incompatibility is less of a concern: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container

> Security settings that you specify for a Container apply only to the individual Container, and they override settings made at the Pod level when there is overlap

It is still a breaking change though, since any added sidecars that are incompatible will need to specify context where they didn't previously, so we'd need to note as much in the changelog.

What issue were you seeing with Kuma, on which version, and with what install configuration? At least with https://kuma.io/docs/2.7.x/production/dp-config/cni/#installation and the KIND instructions I was able to set up Kong and our test echo server with Pod-level `runAsNonRoot: true`. That may well have tripped it since AFAIK https://kuma.io/docs/2.7.x/production/dp-config/transparent-proxying/ actually does require root, but I'd think it'd set the override on sidecars in that case.

However, I'd expect UID 1000 at the Pod level to break that also, since even if the container _can_ run as root, it wouldn't be.

ubergesundheit:
> What issue were you seeing with Kuma, on which version, and with what install configuration?

I was running tests locally with the versions specified in the test files. As said the injected Kuma initContainers were not happy with the `runAsNonRoot` setting as these were set explicitly to run with uid 0

charts/kong/values.yaml
ubergesundheit: Seems to work with only setting these in the containerSecurityContext

rainest:
Would the Pods come online if you manually added a `runAsNonRoot: false` to the kuma initContainers? I think that should work in namespaces that permit it.

Trying to do my own tests, but have been frustrated by some other issue interfering with my ability to run the transparent proxy mode locally.

ubergesundheit:
> Would the Pods come online if you manually added a runAsNonRoot: false to the kuma initContainers? I think that should work in namespaces that permit it.

As long as the pods `securityContext` does not already specify `runAsNonRoot: true`, it should work. But I did not test this.

ubergesundheit:
@rainest sorry for letting this stall. I've rebased and ran `make test.golden.update`. I've also addressed [your comment regarding securityContext](https://github.com/Kong/charts/pull/1057#discussion_r1580269862)

ubergesundheit:
@rainest is this still something worth merging?

ubergesundheit:
Hello @pmalek can you take a look at this PR? Thank you :)

charts/kong/values.yaml
```diff
@@ -934,19 +951,29 @@ podDisruptionBudget:
 podSecurityPolicy:
```
pmalek: Given that `PodSecurityPolicy` [is removed as of k8s 1.25](https://kubernetes.io/docs/concepts/security/pod-security-policy/) I'd say that I don't see the point of changing these values at all.

Do you mind sharing why would you see this being changed?

charts/kong/values.yaml
```diff
@@ -964,13 +991,16 @@ podSecurityPolicy:
...
+securityContext:
+  seccompProfile:
+    type: RuntimeDefault
 
 # securityContext for containers.
 containerSecurityContext:
   readOnlyRootFilesystem: true
   allowPrivilegeEscalation: false
   runAsUser: 1000
+  runAsGroup: 1000
```
pmalek, line 1012: This looks sensible

charts/kong/values.yaml
```diff
@@ -964,13 +991,16 @@ podSecurityPolicy:
 priorityClassName: ""
 
 # securityContext for Kong pods.
-securityContext: {}
+securityContext:
+  seccompProfile:
+    type: RuntimeDefault
```
pmalek, line 1005: This looks like a good change as well

pmalek (commented):
These changes look good. Please rebase on latest `main`. Left 1 question.

charts/kong/values.yaml
ubergesundheit: Back in April 2024, when I created the PR, I was running kong on a 1.24 cluster. I agree this isn't relevant any more, so I removed the changes

ubergesundheit:
Thanks, rebased to current main :+1:

ubergesundheit:
I've seen that kube-linter also failed for the kube-rbac-proxy in the gateway-operator chart. I rebased again and updated the containers securityContext.

The other test failures seem to be due to helm dependency failures. I think they should work after retrying

charts/kong/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+* Run containers with read-only file system and non-root user to increase container and pod security.
```
pmalek, line 7: Can you please add a note here about this change being possibly breaking for users using sidecars?

ubergesundheit: Yes. I've added 

```
 **Note**: This change may be incompatible with user sidecar containers. In this case, change the
  `securityContext` in your values.
```

pmalek (approved):

pmalek:
Thanks @ubergesundheit 🙇

PR #1058, opened 2024-04-18
kgo: add args and provide current defaults via env

pmalek:

Provide `args` section KGO chart.

This will help users to migrate from pre 1.2 versions which didn't have the ability to use the env vars (implemented in https://github.com/Kong/gateway-operator-archive/pull/1616).

The reason for this is that pre 1.2 had `ControlPlane` and `GatewayConfiguration` in `v1alpha1` and 1.2+ have those CRDs only defined in `v1beta1` which makes it impossible to easily migrate using this chart which didn't provide users a way to disable `Gateway` and `ControlPlane` controllers to prevent errors like these:

```
manager {"level":"error","ts":"2024-04-18T12:33:54Z","msg":"failed to run manager","error":"no matches for kind \"ControlPlane\" in version \"gateway-operator.konghq.com/v1beta1\"","stacktrace":"main.main\n\t/workspace/main.go:17\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:271"}
Stream closed EOF for kong-system/kgo-gateway-operator-controller-manager-7549bd865b-vvzkt (manager)
```

Using the `args` provided in this PR users will be able to disable the controllers that require manual upgrade path (`ControlPlane` and `Gateway`), migrate to a newer KGO version, install new CRDs, and if they'd like enable those disabled controllers.

### DIFF
--- a/charts/gateway-operator/ci/env-and-customenv-values.yaml
+++ b/charts/gateway-operator/ci/env-and-customenv-values.yaml
@@ -1,0 +1,6 @@
+env:
+  # gateway controller
+  enable_controller_gateway: false
+
+customEnv:
+  TZ: "Europe/Berlin"
--- a/charts/gateway-operator/Chart.yaml
+++ b/charts/gateway-operator/Chart.yaml
@@ -8,7 +8,7 @@ maintainers:
 name: gateway-operator
 sources:
   - https://github.com/Kong/charts/tree/main/charts/gateway-operator
-version: 0.1.5
+version: 0.1.6
 appVersion: "1.2"
 annotations:
   artifacthub.io/prerelease: "false"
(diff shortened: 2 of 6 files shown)


charts/gateway-operator/ci/env-and-customenv-values.yaml
czeslavo, line 1: How about adding one more with `args` supplied?

czeslavo (dismissed):

charts/gateway-operator/ci/env-and-customenv-values.yaml
pmalek: Added

pmalek: Ideally we'll cover those in more details with #1060 (thanks for creating the issue) but those should suffice for now.

PTAL

czeslavo (approved):

PR #1059, opened 2024-04-18
chore: update dep versions used in CI

pmalek:

- Update dep versions used in CI
  - kind
  - k8s
 
- Do not check out with `fetch-depth` where not necessary to make CI checkout faster.

- Use `main` target branch in `ct install` and `ct lint` calls so that we no longer rely on branch that synced in https://github.com/Kong/charts/blob/6269907dc8c8af41edbb3f43b6f446deba09a30f/.github/workflows/master-mirrors-main.yaml but rather on the actual default branch which is as of now `main`.

### DIFF
--- a/.github/workflows/release-pr.yaml
+++ b/.github/workflows/release-pr.yaml
@@ -36,7 +36,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       # Specify this here because these tests rely on ktf to run kind for cluster creation.
-      KIND_VERSION: v0.20.0
+      KIND_VERSION: v0.22.0
     strategy:
       matrix:
         chart-name:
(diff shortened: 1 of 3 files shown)


.github/workflows/release-pr.yaml
```diff
@@ -74,32 +74,32 @@ jobs:
...
     strategy:
       matrix:
         kubernetes-version:
-          - "1.23.13"
-          - "1.24.7"
-          - "1.25.9"
-          - "1.26.4"
-          - "1.27.1"
+          - "1.23.17"
+          - "1.24.17"
```
czeslavo, line 90:
```suggestion
```
I think we could drop the two oldest versions after adding the two newer ones (that would reflect the approach we follow in KIC).

pmalek: Those are still present in e.g. https://docs.konghq.com/kubernetes-ingress-controller/latest/reference/version-compatibility/#general (at least for older versions which we also tests in the integration tests suite). I'd leave them until those versions go out of support

czeslavo (approved):

PR #1062, opened 2024-04-29
feat(kgo): add ability to specify probes

pmalek:

Add ability to specify `readinessProbe` and `livenessProbe` for KGO Pods.

#### Checklist

### DIFF
--- a/charts/gateway-operator/Chart.yaml
+++ b/charts/gateway-operator/Chart.yaml
@@ -8,7 +8,7 @@ maintainers:
 name: gateway-operator
 sources:
   - https://github.com/Kong/charts/tree/main/charts/gateway-operator
-version: 0.1.6
+version: 0.1.7
 appVersion: "1.2"
 annotations:
   artifacthub.io/prerelease: "false"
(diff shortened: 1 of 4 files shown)


rainest (approved):
I guess we just don't have a changelog for this chart at all yet? We should probably add that.

FWIW this will be affected by the same issue that affects the kong chart, where you can't add checks that aren't `httpGet` due to the way Helm handles YAML merges. That probably doesn't really matter as you shouldn't need to change that outside of removing the checks entirely for special debug reasons (usually running Delve, which prevents the endpoints from responding when paused).

pmalek:
> I guess we just don't have a changelog for this chart at all yet? We should probably add that.

https://github.com/Kong/charts/pull/1063

> FWIW this will be affected by the same issue that affects the kong chart, where you can't add checks that aren't `httpGet` due to the way Helm handles YAML merges. That probably doesn't really matter as you shouldn't need to change that outside of removing the checks entirely for special debug reasons (usually running Delve, which prevents the endpoints from responding when paused).

ACK

PR #1082, opened 2024-06-10
chore: update versions in CI

pmalek:

Remove unsupported versions of k8s (as of https://docs.konghq.com/kubernetes-ingress-controller/latest/reference/version-compatibility/#general)


czeslavo (approved):
👍

PR #1094, opened 2024-06-17
Merge main into 3.x RC

rainest:
Merge main into the https://github.com/Kong/charts/issues/921 3.x branch and update various things that touch templates modified for https://github.com/Kong/charts/pull/947. Part of https://github.com/Kong/charts/issues/921.

Commits since fd06beb272d9c68bb5cf9c03dcd5af86ac14b6b7 are the non-merge fixes. I don't think there's any immediately obvious way to separate the diffs from the merge and the fixes in GitHub views. `git diff 5b969dec59ac146713f4eeed58bdfdd5964c0e37` will show the additional changes since the merge.

The admission webhook template has gotten rather complicated. We may want to simplify it by removing all the cases for KIC 2.x. I think limiting support to KIC 3.x+ is reasonable, but this would mean LTS support would have to be handled via additional chart 2.x releases.

Next steps will be to figure out how to port over the additional changes from https://github.com/Kong/charts/pull/976 and any additional stuff after.


.github/workflows/release-pr.yaml
```diff
@@ -119,34 +121,31 @@ jobs:
       - name: cleanup integration tests (cleanup)
         run: ./scripts/test-env.sh cleanup
   oldversion-integration-test:
+    timeout-minutes: 30
```
programmer04, line 124: We rather use the CI level env variable for setting it consistently in other repos

programmer04 (approved):

.github/workflows/release-pr.yaml
rainest: Minor chore for the future; as-is I'm leaving it because it's just merge content here.

PR #1105, opened 2024-07-16
Release consolidated chart as kong 3.0.0-rc1

rainest:

Releases the consolidated chart as kong 3.0.0-rc1.

Removes most existing `ingress` chart content and consolidates some `ingress` documentation into the `kong` files. Leaves a brief migration message and historical under `ingress`.

Clone of #976 to no longer ping Michał.

https://helm.sh/docs/chart_template_guide/helm_ignore_file/ provides more tooling if we need to persist `ingress` in some fashion re chartsnap and whatnot, but AFAICT we don't need it for consolidation so long as the chartsnap config is gone and `charts/ingress` just holds the old changelog/readme. It's available if some other Helm tool wants to do something with the old dir.


scripts/test-run.sh
```diff
@@ -28,21 +28,11 @@ TEST_ENV_NAME="${TEST_ENV_NAME:-kong-charts-tests}"
 KUBECTL="kubectl --context kind-${TEST_ENV_NAME}"
 KUBERNETES_VERSION="$($KUBECTL version -o json | jq -r '.serverVersion.gitVersion')"
 
-CONTROLLER_PREFIX=""
-GATEWAY_PREFIX=""
 ADDITIONAL_FLAGS=()
 
 # ------------------------------------------------------------------------------
-# Configure per-chart settings
+# Deploy Kuma configuration and test namespace
```
randmonkey: Is there anything to do with kuma?

randmonkey (commented):
Generally looks good, some minor comments.

scripts/test-run.sh
rainest: Nope, there's another block lower that does that has the same comment, this was some garbage I didn't catch when reviewing the conflicts. Good catch.

randmonkey (approved):

PR #1118, opened 2024-08-22
feat(kong): support configurable kgo deployment selector labels

szesch:

We are currently deploying KGO through kustomize and are trying to migrate to helm with zero downtime. One issue I've seen is that the deployment selector labels have changed. Since that is an immutable field it causes issues with the helm deploy of the gateway operator. This PR adds the ability to override the default selector labels.

#### Checklist

### DIFF
--- a/charts/gateway-operator/templates/deployment.yaml
+++ b/charts/gateway-operator/templates/deployment.yaml
@@ -10,7 +10,11 @@ spec:
   replicas: {{ .Values.replicaCount }}
   selector:
     matchLabels:
+      {{- if .Values.selectorLabels }}
+      {{- toYaml .Values.selectorLabels | nindent 6}}
+      {{- else }}
       {{- include "kong.selectorLabels" . | nindent 6 }}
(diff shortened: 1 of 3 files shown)


charts/gateway-operator/templates/deployment.yaml
```diff
@@ -10,7 +10,11 @@ spec:
   replicas: {{ .Values.replicaCount }}
   selector:
     matchLabels:
+      {{- if .Values.selectorLabels }}
```
pmalek, line 13: Can we add this commented to the `values.yaml` so that people know that they can use it and know what setting this field means?

pmalek (commented):
Please bump the patch version in https://github.com/Kong/charts/blob/7f129440dbfe5f4c27a6355f1463376c024a6353/charts/gateway-operator/Chart.yaml#L10-L11 and set the release version in changelog to 0.2.1.

This will cause a release straight away after merging.

pmalek (approved):

PR #1134, opened 2024-09-23
feat(kong): add metrics Service for ingress controller exposed metrics

thameezb:
Correctly adds in ports to the KIC webhook service so that the servicemonitor can correctly match 

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -6,10 +6,7 @@
 
 * Explicitly cert-manager group on Certificate. [#1128](https://github.com/Kong/charts/pull/1128)
 * Add missing load balancer service values. [#1116](https://github.com/Kong/charts/pull/1116)
-
-### Fixes
-
-* Enables KIC metrics to be scraped by adding in a service to correctly work with the defined ServiceMonitor.
+* Enables KIC metrics to be scraped by adding in a service to correctly work with the defined ServiceMonitor. [#1134](https://github.com/Kong/charts/pull/1134)
 
 ## 2.42.0
 
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -9,7 +9,7 @@
 
 ### Fixes
 
-* Fixes KIC webhook service config to correctly work with ServiceMonitor (by adding in the correct ports to match ServiceMonitor config)
+* Enables KIC metrics to be scraped by adding in a service to correctly work with the defined ServiceMonitor.
 
 ## 2.42.0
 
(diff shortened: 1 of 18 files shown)


charts/kong/templates/admission-webhook.yaml
```diff
@@ -227,6 +227,16 @@ spec:
...
+  {{- if and .Values.ingressController.enabled (semverCompare ">= 2.0.0" (include "kong.effectiveVersion" .Values.ingressController.image)) }}
+  - name: cmetrics
+    port: 10255
+    protocol: TCP
+    targetPort: cmetrics
+  - name: status
+    port: 10254
+    protocol: TCP
+    targetPort: cstatus
+  {{- end }}
```
pmalek: Instead of adding ports to webhook service, I'd suggest that we add a separate service which will be only responsible for metrics. This way it will be easier for users to create RBAC rules around and it will not require to run webhook (some users don't) for those that only want metrics.

WDYT?

pmalek (commented):
Thanks for the contribution.

I believe this is a nice addition. I'd consider adding a separate `Service` which would only expose the metrics port.

LMKWYT

charts/kong/templates/admission-webhook.yaml
thameezb: Thanks for the response. 

> This way it will be easier for users to create RBAC rules around and it will not require to run webhook (some users don't) for those that only want metrics.

For GW services these ports are already added to the service object correctly (and work as expected). However these ports are missing _just_ on the controller (webhook) service (as the service object is created in a distinct manner when compared to the GW services which use the service template). 
Therefore not opposed to creating a distinct service, but it will still be tightly coupled to the webhook service. 

With the above context, is the recommendation still to add a distinct service for this purpose?

pmalek: > but it will still be tightly coupled to the webhook service

Why do you think that's going to be the case? Metrics are a separate aspect from webhook.

I'd go the path of adding a separate `Service`, just exposing KIC metrics. Webhook should really be a concern here.

thameezb: Have added in the additional service (with the same restrictions as the webhook service)

charts/kong/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
 * Explicitly cert-manager group on Certificate. [#1128](https://github.com/Kong/charts/pull/1128)
 * Add missing load balancer service values. [#1116](https://github.com/Kong/charts/pull/1116)
 
+### Fixes
+
```
pmalek, line 11:
```suggestion
```

charts/kong/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
...
+* Enables KIC metrics to be scraped by adding in a service to correctly work with the defined ServiceMonitor.
+
```
```suggestion
* Enables KIC metrics to be scraped by adding in a service to correctly work with the defined ServiceMonitor. [#1134](https://github.com/Kong/charts/pull/1134)

```

pmalek (approved):

Issue #1060, opened 2024-04-18
Run golden tests for KGO chart


czeslavo:
Similarly to `kong` and `ingress` chart, let's run golden tests based on `charts/kgo/ci` manifests for `kgo` chart.

sandromodarelli:
closing as solved

PR #1148, opened 2024-10-16
fix(kgo): fix RBAC policy rules

pmalek:

Fix KGO's RBAC policy rules:

```
1 reflector.go:547] pkg/mod/k8s.io/client-go@v0.30.2/tools/cache/reflector.go:232: failed to list *v1alpha1.DataPlaneMetricsExtension: dataplanemetricsextensions.gateway-operator.konghq.com is forbidden: User "system:serviceaccount:kong-system:controller-manager" cannot list resource "dataplanemetricsextensions" in API group "gateway-operator.konghq.com" at the cluster scope
E1016 08:37:12.380734
```

This PR changes the generation script by using both KGO and KGOEE. The reason for this? KGO EE's manager role contains a super set of required policy rules so we can't just use KGO's manager role. Hence this change uses the base manifests + KGO EE's manager role.

There's probably a better way of doing this but this works ™️ .

NOTE: at some point we could consider using `kubebuilder` code markers instead of using already produced manifests if that would be easier/better.

The resulting RBAC manifests were generated with

```
charts/gateway-operator/scripts/update-rbac-resources.sh $KGO_REPO $KGOEE_REPO $CHARTS_REPO
```

This change used
- KGO main (9093830f84aedfe5e96aa3c812aef72d8f2c541c)
- KGO EE main (620eef100901efc029d4db7c126644bbb96b075d) + bumped KGO main dependency (same as above) (reason for this is that there's a pending PR blocked by failing tests https://github.com/Kong/gateway-operator-enterprise/pull/317)

This is related to #1060 as those tests would greatly catch errors that produce unwanted changes in manager's role policy rules (re #1138) especially now that newer versions of controller tools generate aggregated policy rules so it's harder to notice the change in permissions for a particular type.

#### Checklist


programmer04 (approved):

charts/gateway-operator/templates/rbac-resources.yaml
pmalek, line 1: This has its indentation changed because `yq --indent 0` doesn't seem to work for some reason.

PR #1176, opened 2024-11-29
feat(conf): Add support for `excludeMigrationPod` to userDefinedVolumes and userDefinedVolumeMounts

ashman1984:

At the moment Kong threats migrations PODs the same way as proxy PODs, and mounts all user defined PVs to both proxy and migration PODs. 

While this might be useful in some situations, in most cases it is unnecessary to have user defined volumes mounted to migration PODs, and in some cases even not desired by the user. 

This PR allows users to exclude user defined PODs from being mounted for migration.

This could be possible in the following ways:

1. Introduce additional property to _ userDefinedVolumeMounts_ and _ userDefinedVolumes_ to allows exclusion of user defined PODs from being mounted to migration jobs.

2. Obsolete current _userDefinedVolumeMounts_ and _ userDefinedVolumes_ properties (located under root element "deployment") in favour of dedicated mounts for proxy and migration in a similar fashion as it is done for the ingress controller, where _userDefinedVolumeMounts_ and _ userDefinedVolumes_ are defined under "ingressController" root property. 

This might be done by adding _userDefinedVolumeMounts_ and _ userDefinedVolumes_ under root section called "migrations" for migrations PODs, and for proxy under root section "proxy".

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 * Added serviceMonitor.trustCRDsExist for render based deployments
+* Add support for `excludeMigrationPod` to userDefinedVolumes and userDefinedVolumeMounts
 
 ### Changes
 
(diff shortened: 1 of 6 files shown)


pmalek (changes requested):
Thanks for submitting the change. Can you please rebase on latest tip of `main` and add a ci values file so that we can test it in CI?

This will require running `make test.golden.update` to update the golden files.

PR #1215, opened 2025-01-14
feat(kgo): added ValidatingAdmissionPolicy and ValidatingAdmissionPolicyBinding for validating DataPlane ports

pmalek:

Move `DataPlane` ports validation to `ValidatingAdmissionPolicy`.

Testing of these rules remains an aspect to tackle.

#### Checklist

### DIFF
--- a/charts/gateway-operator/CHANGELOG.md
+++ b/charts/gateway-operator/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## 0.4.3
+
+### Changes
+
+- Added `ValidatingAdmissionPolicy` and ``ValidatingAdmissionPolicyBinding` for
+  validating `DataPlane` ports.
+  [#1215](https://github.com/Kong/charts/pull/1215)
+
 ## 0.4.2
 
 ### Changes
--- a/Makefile
+++ b/Makefile
@@ -80,7 +80,9 @@ export GOLDEN_TEST_FAILURE_MSG
 _chartsnap: _chartsnap.deps
 	helm chartsnap -c ./charts/$(CHART) -f ./charts/$(CHART)/ci/ $(CHARTSNAP_ARGS) \
 		-- \
-		--api-versions gateway.networking.k8s.io/v1
+		--api-versions gateway.networking.k8s.io/v1 \
+		--api-versions admissionregistration.k8s.io/v1/ValidatingAdmissionPolicy \
+		--api-versions admissionregistration.k8s.io/v1/ValidatingAdmissionPolicyBinding
 
 .PHONY: _chartsnap.deps
 _chartsnap.deps: chartsnap
--- a/charts/gateway-operator/Chart.yaml
+++ b/charts/gateway-operator/Chart.yaml
@@ -8,7 +8,7 @@ maintainers:
 name: gateway-operator
 sources:
   - https://github.com/Kong/charts/tree/main/charts/gateway-operator
-version: 0.4.2
+version: 0.4.3
 appVersion: "1.4"
 annotations:
   artifacthub.io/prerelease: "false"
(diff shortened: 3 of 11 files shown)


pmalek:
Wait for #1216 to add chartsnap validation.

charts/gateway-operator/CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## 0.4.3
+
+### Changes
+
+- Added `ValidatingAdmissionPolicy` and ``ValidatingAdmissionPolicyBinding` for
```
czeslavo, line 7:
```suggestion
- Added `ValidatingAdmissionPolicy` and `ValidatingAdmissionPolicyBinding` for
```

czeslavo (commented):
One thing I'm worried about is that we end up with the logic responsible for validating objects (we have CEL rules defined on CRDs in KCFG and now `ValidatingAdmissionPolicy/Binding` here) scattered over different places. 

What do you think about defining these in KCFG, e.g. in a separate directory under `config/validation/gateway-operator`? That would allow testing them using the same test machinery we use for CRDs there and we would have a single place to live for the validation logic. We could "import" them in charts the same way we do with CRDs - copy them over to `charts/gateway-operator/charts/gateway-operator-validation`. One thing that is not clear to me is whether we could make the installation of a subchart depend on the capabilities like you did. _Edit: We could have a dedicated script in `charts/gateway-operator/scripts` that would do the copying and adding conditional on capabilities to every file._

pmalek:
> One thing I'm worried about is that we end up with the logic responsible for validating objects (we have CEL rules defined on CRDs in KCFG and now `ValidatingAdmissionPolicy/Binding` here) scattered over different places.

I share this concern.

> What do you think about defining these in KCFG, e.g. in a separate directory under `config/validation/gateway-operator`? That would allow testing them using the same test machinery we use for CRDs there and we would have a single place to live for the validation logic. We could "import" them in charts the same way we do with CRDs - copy them over to `charts/gateway-operator/charts/gateway-operator-validation`. One thing that is not clear to me is whether we could make the installation of a subchart depend on the capabilities like you did. _Edit: We could have a dedicated script in `charts/gateway-operator/scripts` that would do the copying and adding conditional on capabilities to every file._

That would be nice to put in kcfg which is a dedicated repo ATM for these things. I'm worried this will make the workflow here unnecessarily complicated and error prone 🤔 The subcharts approach seems to me quite problematic to update and maintain but perhaps the reason for this is not enough automation on our side.

The problem I see with using kcfg is that we might not want to create numerous releases in that repo which then would create unnecessary dependency updates in other repositories (and also be user facing)

I'm happy to discuss this further as I believe the approach you proposed has advantages.

> whether we could make the installation of a subchart depend on the capabilities

That I believe would not be possible as according to https://helm.sh/docs/topics/charts/#the-chartyaml-file the `condition` is `A yaml path that resolves to a boolean, used for enabling/disabling charts (e.g. subchart1.enabled )`

czeslavo (dismissed):

pmalek:
As agreed on slack: we keep the definition of policies for now both in charts and in KGO repositories. When we decide to move the charts internals (the releases would remain in `charts`) to KGO then we use the policies from KGO's repo.

Waiting on this one to get a final review in https://github.com/Kong/gateway-operator/pull/1007.

charts/gateway-operator/CHANGELOG.md
pmalek, line 7:
```suggestion
- Added `ValidatingAdmissionPolicy` and `ValidatingAdmissionPolicyBinding` for
```

pmalek:
@czeslavo This is now ready for review

czeslavo (dismissed):

czeslavo (dismissed):

czeslavo (approved):

PR #1252, opened 2025-02-18
fix: admissionWebhook bad position

kowaalski:
**_What this PR does / why we need it:_**

This PR fixes a misconfiguration in the Helm chart for Kong where the values for the admissionWebhook were placed at the root level instead of being nested under the ingressController block. As a result, the generated ValidatingWebhookConfiguration was not reflecting the intended settings (such as the namespaceSelector, timeoutSeconds, etc.), which could lead to unexpected behavior when validating Secrets and other resources. By moving the admissionWebhook values inside the ingressController block, the configuration becomes consistent and is rendered correctly.

Which issue this PR fixes
	•	N/A

**_Special notes:_**
	•	The change involves relocating the admissionWebhook configuration block from the root of the values file into the ingressController section.
	•	This ensures that all admission webhook settings (such as namespaceSelector, timeoutSeconds, and objectSelector) are properly applied to the generated ValidatingWebhookConfiguration resources.
	•	Please verify that the change does not impact other parts of the chart and that it still supports overriding via the ingressController.admissionWebhook values.

**_Checklist_**
- [X] PR is based off the current tip of the main branch.
- [ ]  Changes are documented under the "Unreleased" header in CHANGELOG.md
- [ ]  New or modified sections of values.yaml are documented in the README.md
- [ ]  Commits follow the [Kong commit message guidelines](https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#commit-message-format)

### DIFF
--- a/charts/kong/values.yaml
+++ b/charts/kong/values.yaml
@@ -549,6 +549,35 @@ ingressController:
     # version, set this to "3.1.0".
     effectiveSemver:
   args: []
+  admissionWebhook:
+    matchPolicy: Equivalent
+    enabled: true
+    # Limit the `secrets.plugins.validation.ingress-controller.konghq.com` webhook
+    # to only Secrets with the appropriate KIC "konghq.com/validate" label.
+    filterSecrets: false
+    failurePolicy: Ignore
+    port: 8080
+    certificate:
+      provided: false
+    namespaceSelector: {}
+    # ObjectSelector specifies which objects to match against validations.kong.konghq.com webhook webhook
+    objectSelector:
+      matchExpressions:
+      - key: owner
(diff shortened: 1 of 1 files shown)


czeslavo:
Hi @kowaalski, thanks for the contribution. Are you sure the proposed change is needed? As I can see, both before and after, the `admissionWebhook` is defined under `ingressController`, with the same indentation.

kowaalski:
> Hi @kowaalski, thanks for the contribution. Are you sure the proposed change is needed? As I can see, both before and after, the `admissionWebhook` is defined under `ingressController`, with the same indentation.

Sorry, you are correct, I have to clean my glasses.

PR #1255, opened 2025-02-26
CI: fix kcfg sync workflow

tao12345666333:
#### What this PR does / why we need it:

This pull request includes updates to the GitHub Actions workflow for syncing Kubernetes configuration CRDs. The changes focus on updating permissions, switching the shell used for a script, and modifying the process for committing changes to a new branch.

Workflow updates:

* [`.github/workflows/sync-kubernetes-configuration-crds.yaml`](diffhunk://#diff-fa583d7d8743d7a0d8820418778791764e474118a20941444c847a45a9271eb5R11-R13): Added `permissions` section to grant write access to contents.
* [`.github/workflows/sync-kubernetes-configuration-crds.yaml`](diffhunk://#diff-fa583d7d8743d7a0d8820418778791764e474118a20941444c847a45a9271eb5L21-R24): Changed the shell used for the Kustomize installation script from `sh` to `bash`.
* [`.github/workflows/sync-kubernetes-configuration-crds.yaml`](diffhunk://#diff-fa583d7d8743d7a0d8820418778791764e474118a20941444c847a45a9271eb5L37-R46): Replaced the action for creating a pull request with an action to push changes to a new branch, including updating commit message and branch creation options.

#### Which issue this PR fixes
*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*
  - fixes #

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [ ] PR is based off the current tip of the `main` branch.
- [ ] Changes are documented under the "Unreleased" header in CHANGELOG.md
- [ ] New or modified sections of values.yaml are documented in the README.md
- [ ] Commits follow the [Kong commit message guidelines](https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#commit-message-format)

PR #1278, opened 2025-03-18
ci: set .version and .appVersion in Chart.yaml when updating KGO's kubernetes-configuration-crds subchart

pmalek:
  - fixes #

#### Checklist

### DIFF
--- a/.github/workflows/sync-kubernetes-configuration-crds.yaml
+++ b/.github/workflows/sync-kubernetes-configuration-crds.yaml
@@ -29,6 +29,12 @@ jobs:
         run: |
           kustomize build /tmp/kubernetes-configuration > charts/gateway-operator/charts/kubernetes-configuration-crds/crds/kubernetes-configuration-crds.yaml
 
+      - name: Set .version and .appVersion in Chart.yaml
+        run: |
+          make ci.set-chart-version \
+            CHART_YAML=charts/gateway-operator/charts/kubernetes-configuration-crds/Chart.yaml \
+            CHART_VERSION=${{ github.event.inputs.version }}
+
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@271a8d0340265f705b14b6d32b9829c1cb33d45e
         with:
--- a/Makefile
+++ b/Makefile
@@ -81,6 +81,12 @@ test.golden.update:
 	@ $(MAKE) _chartsnap CHART=ingress CHARTSNAP_ARGS="-u"
 	@ $(MAKE) _chartsnap CHART=gateway-operator CHARTSNAP_ARGS="-u"
 
+.PHONY: ci.set-chart-version
+ci.set-chart-version:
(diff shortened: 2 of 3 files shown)


tao12345666333:
After updating the version, we need to update the dependencies

tao12345666333:
@pmalek  I remembered why I didn't add the modification for the chart version in #1255.

https://github.com/Kong/charts/pull/1255#discussion_r1977012375

If we automatically modify the chart version of `kubernetes-configuration`, it also means that we need to change the version in the`Chart.yaml` of `gateway-operator`. 
This means that a new version of the `gateway-operator` chart needs to be released. However, in some cases, it is necessary to update the binary or image version of the `gateway-operator` to work with the CRD in the `kubernetes-configuration`.

This might be a bit like the chicken and egg problem.

pmalek:
> @pmalek I remembered why I didn't add the modification for the chart version in #1255.
> 
> [#1255 (comment)](https://github.com/Kong/charts/pull/1255#discussion_r1977012375)
> 
> If we automatically modify the chart version of `kubernetes-configuration`, it also means that we need to change the version in the`Chart.yaml` of `gateway-operator`. This means that a new version of the `gateway-operator` chart needs to be released.

~Right. That is a sad limitation of us using the subcharts nested inside the chart subdirectory.~
~I wonder though if `version` in `Chart.yaml`'s `dependencies` is required 🤔~

Is that actually the case? If we change the `dependencies.version` for `kubernetes-configuration-crds` subchart we just change that, right? We don't need to change the `.version` of the chart itself. That can be done separately when we're ready to release, right?

tao12345666333:
> Is that actually the case? If we change the `dependencies.version` for `kubernetes-configuration-crds` subchart we just change that, right? We don't need to change the `.version` of the chart itself. That can be done separately when we're ready to release, right?

Right. But I'm worried that someone might use the Helm chart by cloning the git repo. This could cause some confusion.

At the same time, I also thought of another question: we haven't committed to the Chart on the main branch always being stable, right? For example, we have updated the CRDs, but have not yet updated the binaries, so the chart in the main branch cannot be used properly.

pmalek:
> > Is that actually the case? If we change the `dependencies.version` for `kubernetes-configuration-crds` subchart we just change that, right? We don't need to change the `.version` of the chart itself. That can be done separately when we're ready to release, right?
> 
> Right. But I'm worried that someone might use the Helm chart by cloning the git repo. This could cause some confusion.
> 
> At the same time, I also thought of another question: we haven't committed to the Chart on the main branch always being stable, right? For example, we have updated the CRDs, but have not yet updated the binaries, so the chart in the main branch cannot be used properly.

True, while we have not committed to keeping `main` working at all times we should strive to do so.

Even syncing the CRDs at the very last minute prior to a release of a new chart version with the corresponding image bump would break the builds (until the image is patched). Let's keep this as is then.

PR #1311, opened 2025-05-13
feat: Add ingressController.createIngressClass in values

### LINKED ISSUE #1310: Add option in values to disable creating `ingressclass`
From FTI-6672.
Customer is using a k8s user who does not have the permission to create cluster scoped resource so they failed to install KIC by helm because they cannot create the ingress class. 
We can add an option to disable creating ingress class in the charts to resolve this.

randmonkey:

Add `ingressController.createIngressClass` in the values to allow users to disable creating ingress classes. This allows users who do not have the permission to create ingress classes (For example, the user who only have permission of certain namespaces) to use helm to install KIC.

  - fixes #1310 

#### Checklist


pmalek (approved):

pmalek (commented):
Could we perhaps add a CI values file which would ensure that setting `createIngressClass` to `false` does not generate the `IngressClass` in the manifest?

pmalek (changes requested):
Requesting change to add a CI values to test `createIngressClass` set to `false`.

randmonkey:
> Requesting change to add a CI values to test `createIngressClass` set to `false`.

Done.

charts/kong/ci/__snapshots__/disable-creating-ingress-class.snap
```diff
@@ -0,0 +1,1185 @@
...
+  - list
+  - watch
+- apiGroups:
+  - networking.k8s.io
+  resources:
+  - ingressclasses
+  verbs:
+  - get
+  - list
+  - watch
```
pmalek: I believe that this might fail as well for a user who does not have permissions to work with `IngressClass`es.

In Kubernetes RBAC, to give permissions you have to have it yourself. I believe this part of the `ClusterRole` should also be removed when `createIngressClass` is `false`.

randmonkey: This was controlled by `rbac.enableClusterRoles`. As we have a specific value here, I think we should also control it by the value.

Done.

pmalek (approved):

PR #1334, opened 2025-06-24
kong: generate Gateway API RBAC policy rules by default

pmalek:

This PR makes it so that `kong` chart will generate Gateway API RBAC policy rules by default.

This can be disabled by setting `.ingressController.rbac.gatewayAPI.enabled` to `false`.

Release `kong` `2.50.0`.

#### Checklist

### DIFF
--- a/charts/kong/CHANGELOG.md
+++ b/charts/kong/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## Unreleased
+## 2.50.0
 
 ### Changes
 
@@ -9,6 +9,10 @@
   `ingressController.konnect.license.enabled` is set to `true` that enables
   the synchronization of license with Konnect.
   [#1331](https://github.com/Kong/charts/pull/1331)
+* Generate Gateway API RBAC policy rules by default.
+  Users can disable this by setting `.ingressController.rbac.gatewayAPI.enabled`
+  to `false`.
+  [#1332](https://github.com/Kong/charts/pull/1332)
 
 ## 2.49.0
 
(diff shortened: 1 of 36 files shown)


charts/kong/Chart.yaml
```diff
@@ -8,7 +8,7 @@ maintainers:
 name: kong
 sources:
   - https://github.com/Kong/charts/tree/main/charts/kong
-version: 2.49.0
+version: 2.50.0
```
randmonkey, line 11: Should we release charts 2.50 after the default KIC version bumped to 3.5?

pmalek: I'd say let's not hold this release here.

The more releases the better IMHO because we can isolate potential problems instead of aggregating e.g. bumping the image with functional changes.

WDYT?

randmonkey (approved):

charts/kong/CHANGELOG.md
```diff
@@ -9,6 +9,10 @@
   `ingressController.konnect.license.enabled` is set to `true` that enables
   the synchronization of license with Konnect.
   [#1331](https://github.com/Kong/charts/pull/1331)
+* Generate Gateway API RBAC policy rules by default.
```
tao12345666333, line 12: Should we mention that this configuration will only create gwapi's rbac regardless of whether gwapi exists in the cluster, but if gwapi exists in the cluster, this configuration has no effect and cannot prevent gwapi's rbac from being created?

pmalek: Good point, #1342. PTAL

PR #1493, opened 2026-03-05
feat(kong): add additionalProxies support for multiple proxy services

maratsal:
Support multiple proxy services when in addition to default proxy service, users want to deploy extra service with different config.

Add support for defining multiple proxy services from a single Helm values configuration via the new 'additionalProxies' map in values.yaml.

Each entry in additionalProxies creates a separate Kubernetes Service that routes to the same Kong proxy container, enabling use cases like:
- External LoadBalancer with proxy_protocol (e.g., AWS NLB)
- Internal ClusterIP service without proxy_protocol

#### Checklist


charts/kong/templates/service-kong-proxy-additional.yaml
```diff
@@ -0,0 +1,21 @@
...
+{{- range $name, $svcConfig := .Values.additionalProxies }}
+{{- if $svcConfig.enabled }}
+{{- if or (and $svcConfig.http $svcConfig.http.enabled) (and $svcConfig.tls $svcConfig.tls.enabled) }}
+{{- $serviceConfig := dict -}}
+{{- $serviceConfig := merge $serviceConfig $svcConfig -}}
+{{- $_ := set $serviceConfig "fullName" (include "kong.fullname" $) -}}
+{{- $_ := set $serviceConfig "namespace" (include "kong.namespace" $) -}}
+{{- $_ := set $serviceConfig "metaLabels" (include "kong.metaLabels" $) -}}
+{{- $_ := set $serviceConfig "selectorLabels" (include "kong.selectorLabels" $) -}}
+{{- $_ := set $serviceConfig "serviceName" (printf "proxy-%s" $name) }}
```
tao12345666333: I'm concerned about the length here; let me double-check the details.

tao12345666333: The final Service name rendered by kong.service is https://github.com/Kong/charts/blob/f0dfb5ff2e7b81940f70b2fa2a665867a0028e6a/charts/kong/templates/_helpers.tpl#L172-L176                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                              
where `.fullName` itself is already `{{ .Release.Name }}-{{ .Chart.Name }}`. Kubernetes limits Service names to 63 characters, so a long release name combined with a long $name key could exceed this limit.

The chart already applies `trunc 63 | trimSuffix "-"` in similar places, for example https://github.com/Kong/charts/blob/f0dfb5ff2e7b81940f70b2fa2a665867a0028e6a/charts/kong/templates/_helpers.tpl#L21

I think we should do the same here. Something like:

`  {{- $_ := set $serviceConfig "serviceName" (printf "proxy-%s" $name | trunc 63 | trimSuffix "-") }}`

charts/kong/templates/_helpers.tpl
```diff
@@ -1094,7 +1104,17 @@ the template that it itself is using form the above sections.
...
+  {{- $portMapStrings = append $portMapStrings (include "kong.port_maps" .Values.proxy) -}}
+  {{- range $name, $svcConfig := .Values.additionalProxies -}}
+    {{- if $svcConfig.enabled -}}
+      {{- $additionalPortMaps := (include "kong.port_maps" $svcConfig) -}}
+      {{- if (ne (len $additionalPortMaps) 0) -}}
+        {{- $portMapStrings = append $portMapStrings $additionalPortMaps -}}
+      {{- end -}}
+    {{- end -}}
+  {{- end -}}
+  {{- $_ := set $autoEnv "KONG_PORT_MAPS" ($portMapStrings | join ", ") -}}
```
tao12345666333, line 1125: `additionalProxies` streams are added to the deployment container ports, but are not included in `KONG_STREAM_LISTEN` here. This means Kong won't actually listen on those stream ports even though the container exposes them.

  Should add something similar to what was done for `KONG_PROXY_LISTEN` above:

```
  {{- range $name, $svcConfig := .Values.additionalProxies -}}
    {{- if $svcConfig.enabled -}}
      {{- $additionalStream := (include "kong.streamListen" $svcConfig) -}}
      {{- if (not (eq $additionalStream "")) -}}
        {{- $streamStrings = (append $streamStrings $additionalStream) -}}
      {{- end -}}
    {{- end -}}
  {{- end -}}

```

Issue #1079, opened 2024-06-07
[Kong Ingress Controller Chart] Deploy multiple kong instances and customize the controllerName of GatewayClass.


xiaoronglv:
Hi Kong Engineers,

Good day!

I would like to deploy two Kong Gateway Instances to an AWS Kubernetes Cluster (EKS):

1. Public Kong API Gateway
2. Internal Kong API Gateway

##  Install the first Kong Gateway Instance

Kong provides excellent [documentation](https://docs.konghq.com/kubernetes-ingress-controller/latest/install/cloud/eks/) on how to install the Kong Ingress Controller using Helm. The deployment of my public API gateway was very smooth.

Here is the values.yaml file used:

```yaml
gateway:
  proxy:
    http:
      enabled: true
    tls:
      overrideServiceTargetPort: 8000
    annotations:
      external-dns.alpha.kubernetes.io/aws-zone-type-public: true
      external-dns.alpha.kubernetes.io/hostname: "gateway-kong.example.com"
      service.beta.kubernetes.io/aws-load-balancer-backend-protocol: "http"
      service.beta.kubernetes.io/aws-load-balancer-scheme: "internet-facing" #  👈
      service.beta.kubernetes.io/aws-load-balancer-ssl-cert: "arn:aws:acm:us-west-2:****"
      service.beta.kubernetes.io/aws-load-balancer-ssl-ports: "443"
  autoscaling:
    enabled: true
 ```
 
 Here is the command to install the public API gateway:
 
 ```
 helm install kong kong/ingress -n kong --values ./values.yml --create-namespace
 ```
 
 After installation, I defined the GatewayClass with the default controllerName konghq.com/kic-gateway-controller:
 
 ```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: GatewayClass
metadata:
  name: kong
  annotations:
    konghq.com/gatewayclass-unmanaged: 'true'
spec:
  controllerName: konghq.com/kic-gateway-controller #  👈
```

Everything works fine

##  Install second Kong Gateway Instance

After installing the public kong gateway, I started to install the second kong gateway instance as the internal API gateway instance.  I need to specify a new `controllerName:  konghq.com/kic-gateway-controller-private`.  Then I could reference the controllerName when define the GatewayClass for internal purpose.

```
apiVersion: gateway.networking.k8s.io/v1
kind: GatewayClass
metadata:
  name: kong
  annotations:
    konghq.com/gatewayclass-unmanaged: 'true'

spec:
[...]

PR #1519, opened 2026-04-23
Kong chart: scope IngressClass controller name by ingress class

vkobets:
The chart previously set `IngressClass.spec.controller` to a fixed value (`ingress-controllers.konghq.com/kong`). That makes every deployment advertise the same controller identity, which breaks expectations when multiple Kong Ingress Controller instances use different `ingressController.ingressClass` values.
This PR:
- Sets `spec.controller` on the managed `IngressClass` to `ingress-controllers.konghq.com/{{ .Values.ingressController.ingressClass }}` so it matches the configured ingress class.
An Ingress resource that targets a given ingress class resolves to the correct controller instance.

- Related: #1079 (multiple instances - chart now aligns IngressClass with `ingressController.ingressClass`)

#### Checklist

### DIFF
--- a/charts/kong/templates/_helpers.tpl
+++ b/charts/kong/templates/_helpers.tpl
@@ -502,6 +502,7 @@ The name of the Service which will be used by the controller to update the Ingre
   {{- $_ := set $autoEnv "CONTROLLER_KONG_ADMIN_TLS_SKIP_VERIFY" true -}}
   {{- $_ := set $autoEnv "CONTROLLER_PUBLISH_SERVICE" ( include "kong.controller-publish-service" . ) -}}
   {{- $_ := set $autoEnv "CONTROLLER_INGRESS_CLASS" .Values.ingressController.ingressClass -}}
+  {{- $_ := set $autoEnv "CONTROLLER_GATEWAY_API_CONTROLLER_NAME" (printf "ingress-controllers.konghq.com/-%s" .Values.ingressController.ingressClass) -}}
   {{- $_ := set $autoEnv "CONTROLLER_ELECTION_ID" (printf "kong-ingress-controller-leader-%s" .Values.ingressController.ingressClass) -}}
 
   {{- if .Values.ingressController.admissionWebhook.enabled }}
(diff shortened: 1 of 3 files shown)


charts/kong/templates/ingress-class.yaml
```diff
@@ -29,5 +29,5 @@ metadata:
   labels:
   {{- include "kong.metaLabels" . | nindent 4 }}
 spec:
-  controller: ingress-controllers.konghq.com/kong
+  controller: ingress-controllers.konghq.com/{{ .Values.ingressController.ingressClass }}
```
pmalek, line 32: This change is ok 👍

charts/kong/templates/_helpers.tpl
```diff
@@ -502,6 +502,7 @@ The name of the Service which will be used by the controller to update the Ingre
   {{- $_ := set $autoEnv "CONTROLLER_KONG_ADMIN_TLS_SKIP_VERIFY" true -}}
   {{- $_ := set $autoEnv "CONTROLLER_PUBLISH_SERVICE" ( include "kong.controller-publish-service" . ) -}}
   {{- $_ := set $autoEnv "CONTROLLER_INGRESS_CLASS" .Values.ingressController.ingressClass -}}
+  {{- $_ := set $autoEnv "CONTROLLER_GATEWAY_API_CONTROLLER_NAME" (printf "ingress-controllers.konghq.com/-%s" .Values.ingressController.ingressClass) -}}
```
pmalek: I would not connect these 2 concepts. IngressClass and Gateway API controller name are 2 disjoint configuration aspects.

pmalek (changes requested):
I'm OK with change in charts/kong/templates/ingress-class.yaml but for the other one (re Gateway API controller): I'd not go this route  since those are 2 different things, IngressClass and Gateway API controller name.

charts/kong/templates/ingress-class.yaml
pmalek: After reading this second time I need to take back my comment: this is not about the ingress class here but about the controller. While we could possibly link these 2 (like you did) I'd prefer not to do this.

What I think could be a good way forward and not introducing a breaking change for users is to add a new field in `values.yaml` (possibly: `.ingressController.controllerName`, defaulting to `kong`) and use that here.

This would also prevent what copilot mentioned about controller being immutable (by setting the default to what was already hardcoded in the template).

PR #1539, opened 2026-06-30
feat(kong): configure clear-stale-pid image

tao12345666333:

For the Kong distroless image, which does not include coreutils, meaning commands like `rm` cannot be used, this PR allows users to customize the clear-stale-pid image so that even if the Kong image is distroless, the process of cleaning up PIDs can still be completed normally.

  - fixes #

Related to https://github.com/Kong/kong/issues/6696

#### Checklist

### DIFF
--- a/charts/kong/values.yaml
+++ b/charts/kong/values.yaml
@@ -24,6 +24,16 @@ deployment:
       # Enable the init container which clears the stale PIDs from Kong's prefix directory.
       clearStalePid:
         enabled: true
+        # Image used by the clear-stale-pid init container. Defaults to the Kong image
+        # when unset.
+        image: {}
+          # repository: busybox
+          # tag: "1.36"
+          # pullPolicy: IfNotPresent
+        command:
+        - "rm"
+        - "-vrf"
+        - "$(KONG_PREFIX)/pids"
   # The number of old `ReplicaSet`s to retain.
   revisionHistoryLimit: 10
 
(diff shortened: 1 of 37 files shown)


charts/kong/values.yaml
```diff
@@ -24,6 +24,16 @@ deployment:
...
+        # Image used by the clear-stale-pid init container. Defaults to the Kong image
+        # when unset.
+        image: {}
+          # repository: busybox
+          # tag: "1.36"
+          # pullPolicy: IfNotPresent
+        command:
+        - "rm"
+        - "-vrf"
+        - "$(KONG_PREFIX)/pids"
```
pmalek, line 36: - I'd prefer if we didn't change this (unless necessary)
- Isn't `$()` the syntax for commands, e.g. `$(ls)` will contain the output of `ls` command? Variables should have the `${}`, right? For this reason, I'd leave this is as and if we actually need to change this, I'd track this separately.

pmalek (commented):
This change looks good to me, just 1 minor nit.

We can consider releasing a v4 with that image changed by default to busybox.

charts/kong/values.yaml
tao12345666333: In Kubernetes, we need to use `$()` for the variable to be expanded in the command or args field.

https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#use-environment-variables-to-define-arguments

pmalek (approved):

tao12345666333:
> We can consider releasing a v4 with that image changed by default to busybox.

Good point! I'll merge this PR first, we can include this change when we plan to release v4.

